PR #3, opened 2016-08-23
tar up the right directory

petey:
hopefully root is more accurate than output_dir. If this doesn't work, we'll rework the `ember build` command.


nkatzman:
👍

PR #9, opened 2016-08-31
Add nginx configuration for ember

petey:
Ember does its own routing, so nginx needs to be configured to let it do so. This is a copy of the original config https://github.com/nginxinc/docker-nginx/blob/master/mainline/alpine/nginx.conf with our server config added.


stjohnjohnson:
👍  
xavier

PR #10, opened 2016-09-01
Bower needs a flag to install as root

petey:
Also the UI test container needs to run in node:6. Ember is transpiled with babel down to a browser-level after testing.


Filbird:
👍

stjohnjohnson:
bowser 👍

PR #11, opened 2016-09-01
Check for the string 'false' instead of boolean

petey:
We were not stopping log requests because the header value(s) of 'true' and 'false' are both truthy strings.


FenrirUnbound:
👍 

Looks reasonable to me. As a side note, are we not doing tests yet for the UI?

pt0vfsf

petey:
The logs component needs something more akin to an acceptance test, which we haven't added yet. I intend to add these soon.

PR #13, opened 2016-09-03
Mock pipeline page

petey:
Added a mock for a pipeline page, in prep for adding pipeline secrets. 
Submitting now due to the large refactor. Will flesh out secret functionality later.

**Refactoring:**
* Renamed ALL sd-* components since there didn't seem to be a very good reason to have done this in the first place, other than ember's insistence that all components must have a (-) in the name
* Renamed `build-step` to `build-step-view` and `build-step-container` to `build-step` to be more reflective of their usage.
* Added `build-layout` component to further reduce dependency route-based styles.

**Pipelines**
* Pipeline model, and banner should actually work, and load pipeline data based on url
* secrets entry section added as a mock, no functionality exists, and this eventually needs to be moved to a separate route, e.g. `pipelines/:id/environment`
* pipeline-nav is a mock and the implementation will change entirely when routes exist.

screen shot 2016-09-02 at 5 51 26 pm


stjohnjohnson:
This seems reasonable 👍 

cat fail

app/components/app-header/template.hbs
```diff
@@ -0,0 +1,16 @@
+<header class="page-header">
+  <a href="http://cd.screwdriver.cd/"><div class="logo">
```
stjohnjohnson: We should make this configurable at some point.

tkyi:
👍 

giphy-3

PR #17, opened 2016-09-09
stupid implementation of builds page for pipelines

petey:
Rather silly way of fetching builds and associating the jobs they belong to. We should redo this completely when proper apis exist.

screen shot 2016-09-08 at 9 28 31 pm

### DIFF
--- a/app/job/adapter.js
+++ b/app/job/adapter.js
@@ -1,0 +1,16 @@
+import Adapter from '../application/adapter';
+export default Adapter.extend({
+  handleResponse: (status, headers, payload, requestData) => {
+    let data = payload;
+
+    // Fix our API not returning the model name in payload
+    if (data && Array.isArray(data)) {
+      data = { jobs: data };
+    } else if (data) {
+      data = { job: data };
+    }
+
+    // Pass-through to super-class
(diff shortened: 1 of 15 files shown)


stjohnjohnson:
Looks insane 👍 

cats

app/job/adapter.js
```diff
@@ -0,0 +1,16 @@
+import Adapter from '../application/adapter';
+export default Adapter.extend({
+  handleResponse: (status, headers, payload, requestData) => {
+    let data = payload;
+
+    // Fix our API not returning the model name in payload
```
FenrirUnbound: Not a blocker for this PR. Does this mean that it's a bug with the API, or is this just to make consuming the data easier for the UI?

FenrirUnbound:
👍 

Fun 😺 Looks good to me

ibc8c89 - imgur

app/job/adapter.js
petey: Ember expects: 
```
{ model: { attribute: value } }
```

We give:
```
{ attribute: value }
```
and imply the model name with the route that was called. It isn't easy to generically parse the route that was called in order to determine an implied model name, or I'd do this in ApplicationAdapter and be done with it.

Issue #15, opened 2016-09-07
Build page should scroll with expanded log


stjohnjohnson:
As the log is streaming, the page should auto-scroll to the bottom of the current log so I can easily watch the progress.

PR #19, opened 2016-09-09
build page should refresh better

### LINKED ISSUE #14: Build page doesn't refresh when a step completes
When a step completes (say `install`), the current step doesn't update to show that it's done and that the next one is running.

Expected:
 - Open step log collapses (if exit code `0`).
 - Step bar shows final status
 - Next step is updated with in-progress
 - Next step log opens

petey:
* Resolves #14 and #15. 
* Build model reloads after 5s when queued. 
* Build model reloads when step (logs) finish loading. 
* Log panels open when step is running or failed
* log panels scroll to bottom of logs

### DIFF
--- a/app/build/route.js
+++ b/app/build/route.js
@@ -1,7 +1,19 @@
 import Ember from 'ember';
+const RELOAD_TIMER = 5000;
 
 export default Ember.Route.extend({
   model(params) {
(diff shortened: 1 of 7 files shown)


stjohnjohnson:
LGTM! 👍

PR #21, opened 2016-09-14
Add a pipeline create page

petey:
Adds a create page for pipelines:
* Requires user to be logged in to visit page. Will redirect to login, and redirect back when logged in
* Saves scmUrl in localstorage until successful save
* Redirects to pipeline page for new pipeline on successful create
* displays friendly errors at top of page when create results in error (e.g. 409)

Increased code coverage with acceptance tests.

screen shot 2016-09-14 at 2 47 10 pm

### DIFF
--- a/app/components/create-pipeline-data/component.js
+++ b/app/components/create-pipeline-data/component.js
@@ -1,0 +1,86 @@
+import Ember from 'ember';
+import git from '../../utils/git';
+
+/**
+ * Animates the input to indicate whether the value is a valid scm url or not
+ * @method indicateValidScmUrl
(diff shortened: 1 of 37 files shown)


app/components/create-pipeline-data/component.js
```diff
@@ -0,0 +1,86 @@
+import Ember from 'ember';
+import git from '../../utils/git';
+
+/**
+ * Animates the input to indiciate whether the value is a valid scm url or not
```
tkyi: indicate*

petey: Fixed

PR #22, opened 2016-09-15
add 404 and 500 pages

cynthiax:
- Initial 404/500 page
screen shot 2016-09-15 at 11 53 28 am

-- Need future edit on 503 page
-- Need style alignment


stjohnjohnson:
Oooh, where did we get those images from?

app/error/controller.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  errorHeader: 'Don\'t blink',
```
petey: Not used by the template.

app/error/controller.js
```diff
@@ -0,0 +1,20 @@
...
+  statusCode: 404,
```
petey: Should actually be contained in the model passed to the error substate: https://guides.emberjs.com/v2.8.0/routing/loading-and-error-substates/#toc_code-error-code-substates

app/error/controller.js
```diff
@@ -0,0 +1,20 @@
...
+  errorMessage: 'Unknown error',
```
petey: Also in the model

app/router.js
```diff
@@ -15,6 +15,7 @@ Router.map(function route() {
     this.route('secrets');
   });
   this.route('login');
+  this.route('error', { path: '/*path' });
```
petey: I don't know that a route is explicitly needed for an error substate.

app/error/controller.js
cynthiax: errorHeader was removed

app/error/controller.js
cynthiax: Discussed with petey, will keep it as default for now

app/error/controller.js
cynthiax: Discussed with petey, will keep it as default for now

PR #25, opened 2016-09-16
Add allowInPR field

petey:
Missing allowInPR field prevented secrets from being created properly.
Added some validation on key name, since that is the most likely thing to fail.
Added component to display error messages that occur when creating keys

### DIFF
--- a/app/components/pipeline-secret-settings/component.js
+++ b/app/components/pipeline-secret-settings/component.js
@@ -3,20 +3,39 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   newName: null,
   newValue: null,
+  newAllow: false,
   isButtonDisabled: true,
+  errorMessage: '',
   actions: {
+    /**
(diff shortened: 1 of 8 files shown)


app/components/pipeline-secret-settings/component.js
```diff
@@ -3,20 +3,39 @@ import Ember from 'ember';
...
 
       this.set('isButtonDisabled', isDisabled);
     },
+    /**
+     * Kicks off create secret flow
+     * @method addNewSecret
+     */
     addNewSecret() {
+      if (!/^[A-Z_][A-Z0-9_]*$/.test(this.get('newName'))) {
+        this.set('errorMessage', 'Secret name does not meet requirements /^[A-Z_][A-Z0-9_]*$/');
```
stjohnjohnson: :( Wish we could just use the API response.  Darn you joi.

PR #26, opened 2016-09-16
move login button over

petey:
The button has some arbitrary padding and we want it to be aligned with text on the right of the page.


minz1027:
👍

PR #31, opened 2016-09-19
Refactor secrets model to be an authenticated route

petey:
The secrets page for a pipeline should require authentication. beforeModel messes up the AuthenticatedRouteMixin so refactoring the model function helps.


minz1027:
👍

PR #33, opened 2016-09-20
Make auth a little more intelligent

petey:
Auth refresh was a little naieve about the auth token.
* Refresh now checks expiry in the auth token
* Set better auth window settings when opened


jer:
👍

PR #34, opened 2016-09-20
Adding redirect to API badge service and fix download of release

stjohnjohnson:
Fixing the download so it's not rate limited.  Also adding redirect to SD api.


petey:
👍 

ambush_preparation

PR #39, opened 2016-09-20
Cleanup pipeline header links

petey:
Makes the pipeline appId a link to pipeline page. 
Instead, the branch links to github page.

Added some text to PR column when no PRs are present.


minz1027:
👍 
cat

PR #42, opened 2016-09-21
Fix build page now that jobs api gives better data

petey:
Gets rid of hackery for the pipeline/:id/jobs api call that filters disabled jobs, and reversed job order

### DIFF
--- a/app/pipeline/builds/index/route.js
+++ b/app/pipeline/builds/index/route.js
@@ -32,12 +32,10 @@ export default Ember.Route.extend({
   },
   model() {
     return this.get('pipeline.jobs')
-      // Get rid of disabled jobs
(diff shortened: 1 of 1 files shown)


app/pipeline/builds/index/route.js
```diff
@@ -32,12 +32,10 @@ export default Ember.Route.extend({
   },
   model() {
     return this.get('pipeline.jobs')
-      // Get rid of disabled jobs
-      .then(jobs => jobs.filter(j => j.get('state') !== 'DISABLED'))
       // Split jobs from workflow
       .then(jobs => {
         // filter out the PRs, reverse to get correct order
```
d2lam: fix comment too

stjohnjohnson: :d2fixplz:

minz1027:
👍

PR #43, opened 2016-09-21
Get latest build status

minz1027:
- Get the last build of a job associated with a given `sha` (now it is getting the oldest build)

- Add style to QUEUED and RUNNING status on workflow bar for consistency

screen shot 2016-09-21 at 10 48 53 am

### DIFF
--- a/app/components/pipeline-workflow/styles.scss
+++ b/app/components/pipeline-workflow/styles.scss
@@ -55,21 +55,22 @@ a {
 
 .SUCCESS {
   background-color: $ycolor-green-6b;
-  a {
-    color: white;
-  }
   .arrow-right {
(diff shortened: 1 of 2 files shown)


app/components/pipeline-workflow/styles.scss
```diff
@@ -55,24 +55,25 @@ a {
...
   background-color: $ycolor-green-6b;
-  a {
-    color: white;
-  }
   .arrow-right {
     border-left-color: $ycolor-green-6b;
   }
 }
 
 .FAILURE {
```
petey: need an ABORTED status that acts like FAILURE.

minz1027: Fixed

PR #51, opened 2016-09-24
Acceptance Tests

petey:
In Ember, acceptance tests are one step above integration tests exercising the entirety of the app from route to component. The goal of this PR is to get code coverage > 95%.

Adding tests for:
* [x] Build Page
* [x] Pipeline page
* [x] Pipeline secrets page
* [x] Search Page
* [x] app-header component

Lines 98.09% | Branches 91.41% | Functions 100% | Statements 98.11%

### DIFF
--- a/.ember-cli
+++ b/.ember-cli
@@ -5,6 +5,6 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false,
+  "disableAnalytics": true,
   "usePods": true
 }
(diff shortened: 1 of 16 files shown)


.ember-cli
```diff
@@ -5,6 +5,6 @@
 
     Setting `disableAnalytics` to true will prevent any data from being sent.
   */
-  "disableAnalytics": false,
+  "disableAnalytics": true,
```
petey: I can't think of a good reason to tell google what I'm doing in ember-cli. ember-cli spits out errors when not connected to the internet when this is enabled.

stjohnjohnson: Oh, I read this backwards, I thought it was `analytics: true` Ugh.  Too many negatives.

PR #52, opened 2016-09-26
add a loading panel

petey:
This will automatically fill an {{outlet}} with a loading indication when a model is slow to resolve.

loading

### DIFF
--- a/app/components/loading-view/component.js
+++ b/app/components/loading-view/component.js
@@ -1,0 +1,80 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  /**
+   * Quotes from: https://www.npmjs.com/package/funnies which has an MIT License
+   * @property {Array} funnies
+   */
+  funnies: [
+    'Reticulating splines...',
+    'Generating witty dialog...',
+    'Swapping time and space...',
+    'Spinning violently around the y-axis...',
+    'Tokenizing real life...',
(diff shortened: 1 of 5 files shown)


app/components/loading-view/component.js
```diff
@@ -0,0 +1,72 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  funnies: [
```
stjohnjohnson: Please reference where this came from (and include a comment about this using the MIT license here)

app/components/loading-view/template.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="loading">
+  <h2>{{fa-icon "fa-spinner fa-pulse fa-fw"}} Loading...</h2>
```
tkyi: 👍 I prefer `fa-pulse` over `fa-spin`

app/components/loading-view/component.js
```diff
@@ -0,0 +1,80 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  /**
+   * Quotes from: https://www.npmjs.com/package/funnies which has an MIT License
+   * @property {Array} funnies
+   */
+  funnies: [
+    'Reticulating splines...',
```
d2lam: Sims FTW!

PR #59, opened 2016-09-30
reload secrets after delete

minz1027:
- Reload the secrets after delete it

screen shot 2016-09-30 at 3 17 15 pm

### DIFF
--- a/app/application/adapter.js
+++ b/app/application/adapter.js
@@ -75,7 +75,7 @@ export default DS.RESTAdapter.extend({
    */
   handleResponse(status, headers, payload, requestData) {
     // handle generically when there is an error key in the payload
-    if (payload.error) {
+    if (payload && payload.error) {
       return this._super(status, headers, { errors: payload }, requestData);
     }
 
--- a/tests/unit/application/adapter-test.js
+++ b/tests/unit/application/adapter-test.js
@@ -120,3 +120,15 @@ test('it wraps errors', function (assert) {
 
   assert.ok(payload instanceof DS.AdapterError);
 });
+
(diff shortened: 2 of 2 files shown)


d2lam:
👍  Here is your favorite:

app/application/adapter.js
```diff
@@ -75,7 +75,7 @@ export default DS.RESTAdapter.extend({
    */
   handleResponse(status, headers, payload, requestData) {
     // handle generically when there is an error key in the payload
-    if (payload.error) {
+    if (payload && payload.error) {
```
petey: We should add a test for the adapter receiving no payload.

minz1027: fixed :)

tests/unit/application/adapter-test.js
```diff
@@ -120,3 +120,15 @@ test('it wraps errors', function (assert) {
...
+test('it takes care of empty payload', function (assert) {
+  let adapter = this.subject();
+
+  const requestData = {
+    url: 'http://localhost:8080/v4/pipelines/1234'
+  };
+
+  const payload = adapter.handleResponse(204, {}, null, requestData);
+
+  assert.ok(payload);
```
petey: Whenever I see `assert.ok` on a value, I always wonder if it is actually a boolean being tested.  In this case it is not a boolean. I suppose we really don't care much what the modified payload looks like in this case, but `assert.deepEqual(payload, {});` would be a more accurate test.

minz1027: Thx Petey! Good to know. Fixed already.

PR #61, opened 2016-10-03
Deploy to beta first, then production

stjohnjohnson:
This ensures we test beta environment first and don't accidentally break production

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -1,5 +1,6 @@
 workflow:
   - publish
+  - beta
   - prod
 
 shared:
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -34,6 +35,24 @@ jobs:
...
+      - wait-docker: DOCKER_TAG=`git describe --abbrev=0 --tags` ./ci/docker-wait.sh
+      - deploy-k8s: K8S_TAG=`git describe --abbrev=0 --tags` ./ci/k8s-deploy.sh
+      - test: echo Put acceptance tests here
+    environment:
+      DOCKER_REPO: screwdrivercd/ui
+      K8S_CONTAINER: screwdriver-ui
+      K8S_IMAGE: screwdrivercd/ui
+      K8S_HOST: api.k8s.screwdriver.cd
+      K8S_DEPLOYMENT: sdui-beta
+      SD_UI: cd.screwdriver.cd
```
minz1027: why we need this `SD_UI`?

stjohnjohnson: So the functional tests (when added) can test against the right UI

petey: shouldn't this be beta.cd.screwdriver.cd?

petey:
👍

PR #79, opened 2016-10-24
Update to ember 2.9

petey:
Replaces:
* #77 
* #75 
* #74


tests/index.html
```diff
@@ -21,7 +21,7 @@
     {{content-for "body"}}
     {{content-for "test-body"}}
 
-    <script src="{{rootURL}}testem.js" integrity=""></script>
+    <script src="/testem.js" integrity=""></script>
```
d2lam: Not sure why only this got changed to `/`?

petey: That's what the template in ember was changed to. Didn't see any reason why I needed to do something different. It resolves to the same value.

d2lam: Okay. Just wondering 👍

PR #80, opened 2016-10-25
Update copy on login page

petey:
Not sure if it is any more clear than before, but this is no longer specifically referencing github


stjohnjohnson:
I like it.  We should also document this on our OAuth product description (detailing why we need the permission we asked for).

minz1027:
Looks good! 👍

PR #81, opened 2016-10-26
Change the workflow UI and search page

petey:
- Added an updated UI for the workflow component.
  - simplified workflow component by extracting user-is-authenticated logic to route template
  - made build-bubbles, and the build start button separate components
  - Hovering over workflow bubble also highlights build bubble for a given commit

screen shot 2016-10-26 at 5 04 30 pm
- Enabled the search bar in header. Search page will filter and sort results based on search input.

screen shot 2016-10-26 at 4 34 48 pm
- Added a screwdriver-colors file to give canonical names to colors that we use very often. Started using these color names, instead of fuji-color names in touched components.


app/components/build-bubble/component.js
```diff
@@ -0,0 +1,53 @@
...
+
+      switch (build.get('status')) {
+      case 'SUCCESS':
+        return 'check';
+      case 'FAILURE':
+        return 'times';
+      case 'ABORTED':
+        return 'stop';
+      case 'RUNNING':
+        return 'fa-spinner fa-spin';
```
tkyi: Is there a reason why we use `fa-spin` sometimes and `fa-pulse` for other things? Should we try to be more consistent with spinner styles?

app/styles/app.scss
```diff
@@ -15,6 +15,7 @@
 
 @import "font-awesome";
 @import "fuji-colors";
+@import "screwdriver-colors";
```
tkyi: Cool! 👍

app/components/build-bubble/component.js
petey: The only place we have `fa-pulse` is the spinner on the loading page. I've updated the pr to change that to `fa-spin`

tkyi: Aw okay. I liked `fa-pulse` 😢

PR #83, opened 2016-11-01
Update options page

petey:
Now with more toggles!

screen shot 2016-10-31 at 5 27 20 pm

### DIFF
--- a/app/components/pipeline-options/component.js
+++ b/app/components/pipeline-options/component.js
@@ -5,11 +5,8 @@ export default Ember.Component.extend({
   showRemoveButtons: false,
   isRemoving: false,
   actions: {
-    enableJob(jobId) {
-      this.get('setJobStatus')(jobId, 'ENABLED');
-    },
-    disableJob(jobId) {
-      this.get('setJobStatus')(jobId, 'DISABLED');
+    toggleJob(jobId, event) {
+      this.get('setJobStatus')(jobId, event.newValue ? 'DISABLED' : 'ENABLED');
     },
     showRemoveButtons() {
       this.set('showDangerButton', false);
--- a/app/components/pipeline-options/template.hbs
+++ b/app/components/pipeline-options/template.hbs
@@ -6,20 +6,12 @@
         {{#each pipeline.jobs as |job|}}
         <li>
           <div class="pure-g">
-            <div class="pure-u-1 pure-u-md-2-3">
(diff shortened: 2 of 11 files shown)


app/components/pipeline-options/component.js
```diff
@@ -5,11 +5,8 @@ export default Ember.Component.extend({
   showRemoveButtons: false,
   isRemoving: false,
   actions: {
-    enableJob(jobId) {
-      this.get('setJobStatus')(jobId, 'ENABLED');
-    },
-    disableJob(jobId) {
-      this.get('setJobStatus')(jobId, 'DISABLED');
+    toggleJob(jobId, event) {
+      this.get('setJobStatus')(jobId, event.newValue ? 'DISABLED' : 'ENABLED');
```
FenrirUnbound: Does `event.newValue` only update when it's transitioning from _Enabled_ to _Disabled_?

petey: `event.newValue` will be the value that is changed to when the toggle is clicked. Those values are set in the labels: https://github.com/screwdriver-cd/ui/blob/798890ed499e9de173de708a3d70c188a6acac4f/app/components/pipeline-options/template.hbs#L18

FenrirUnbound: Yeah, I was just confused by the syntax. It looks a lot like if `event.newValue` is populated, it should always be *Disabled*

petey: `event.newValue` is always a boolean, and `event.newValue === true` is unnecessary.

petey: The real confusion is that this is tied to `isDisabled`, so `true` is disabled.

app/components/pipeline-options/template.hbs
```diff
@@ -6,20 +6,16 @@
...
         <li>
           <div class="pure-g">
-            <div class="pure-u-1 pure-u-md-2-3">
-              <h4>{{if job.isDisabled "Enable" "Disable"}} {{job.name}}</h4>
+            <div class="pure-u-4-5">
+              <h4>{{job.name}}</h4>
               {{#if job.isDisabled}}
               <p>Get your {{job.name}} job running again.</p>
               {{else}}
               <p>Stop the line by disabling the {{job.name}} job.</p>
```
minz1027: Can we make this sentence more clear? Like `Toggle to disable the xx job`. Now it is a bit confusing >_<

petey: done

PR #85, opened 2016-11-03
Add unknown user image

petey:
Default image for users when they are not known.


petey:
We shouldn't point to some unknown domain for avatar and url for committers when unknown. This provides an image we host for unknown users.
https://github.com/screwdriver-cd/scm-github/pull/38

d2lam:
@petey  Haha yeah I was just putting ???? as a joke :p

PR #102, opened 2016-12-03
refactor: error substate and component to view page errors

petey:
Before this PR, when the page encountered an error, especially with API calls loading a model, the page would just stop loading data, and wouldn't say why (except in console.error). This PR adds an error substate that should be displayed when errors are caught by Ember during transitions.

* refactored page-not-found logic and template into a component
* pipeline page will now transition to page-not-found on error fetching pipeline, and will display an error message as to why
* Error substate is displayed for "caught" errors, and should display a 500 error image with a reason.

screen shot 2016-12-02 at 6 29 29 pm


app/pipeline/route.js
```diff
@@ -5,6 +5,14 @@ export default Ember.Route.extend({
...
   model(params) {
+    this.set('pipelineId', params.pipeline_id);
+
     return this.get('store').findRecord('pipeline', params.pipeline_id);
+  },
+
+  actions: {
+    error(reason) {
+      this.transitionTo('page-not-found', { path: `pipelines/${this.get('pipelineId')}`, reason });
+    }
```
minz1027: You can also get 403 when trying to access secrets of someone else's repo. But it will be shown as 404 also... Can we have a generic error template and read the statusCode dynamically? 
e.g. click secrets for this repo
https://cd.screwdriver.cd/pipelines/88042a14324158afd35badf37fa2ba52eb2b9fd9

petey: The reason object has a status code embedded in it, but our error component uses images, for some reason, with only a 404 and 500 image existing. I'd like to revisit the actual content of the error-view component in a later PR. This PR is just refactoring it to the error-view component.

minz1027: Cool with another PR 👍

PR #109, opened 2016-12-10
feat: reload content on the pipeline page periodically

petey:
* Adds ModelReloader mixin that will schedule reload on a model periodically.
* Build-bubble, event-list, event-view all use ModelReloader to schedule various model reloads
* event-list only shows 5 (configurable) events by default. Adds a link to show all.

https://github.com/screwdriver-cd/screwdriver/issues/391

### DIFF
--- a/app/components/build-bubble/component.js
+++ b/app/components/build-bubble/component.js
@@ -1,11 +1,24 @@
 import Ember from 'ember';
+import ModelReloaderMixin from 'screwdriver-ui/mixins/model-reloader';
+import ENV from 'screwdriver-ui/config/environment';
 
-export default Ember.Component.extend({
+export default Ember.Component.extend(ModelReloaderMixin, {
   tagName: 'span',
   classNames: ['build-bubble'],
   classNameBindings: ['build.id', 'small'],
+  modelToReload: 'build',
+  reloadTimeout: ENV.APP.BUILD_RELOAD_TIMER,
+  shouldReload(build) {
+    if (build) {
+      const s = build.get('status');
 
-  icon: Ember.computed('jobIsDisabled', 'build', {
+      return s === 'RUNNING' || s === 'QUEUED';
+    }
+
+    return false;
+  },
+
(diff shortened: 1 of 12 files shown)


app/event/model.js
```diff
@@ -19,7 +19,7 @@ export default DS.Model.extend({
     }
   }),
   creator: DS.attr(),
-  duration: Ember.computed('builds', {
+  duration: Ember.computed('builds.[]', 'isComplete', {
```
minz1027: @petey what is the difference between `builds` and `builds.[]`?

app/components/build-bubble/component.js
```diff
@@ -49,5 +62,11 @@ export default Ember.Component.extend({
   },
   mouseLeave() {
     Ember.$('.build-bubble').removeClass('highlight');
+  },
+  willRender() {
+    this.startReloading();
+  },
+  willDestroy() {
+    this.stopReloading();
```
d2lam: Any reason why we don't put `willRender()` and `willDestroy()` inside the ModelReloaderMixin? That way we don't need to repeat the code here and `pipeline-event-view`
@minz1027 and I tried it out locally and it worked.

app/event/model.js
petey: `builds` is the reference to the ArrayProxy object, `builds.[]` is a reference to the contents of the array. http://emberjs.com/api/classes/Ember.Array.html#property__each

By watching the array contents, we recalculate when new values are added to the array, as opposed to never calculating because the array's reference doesn't change.

app/components/build-bubble/component.js
petey: We could do that refactor, but I didn't want to add it initially as I wasn't sure that everything that used the mixin would actually have those events.

petey:
I was concentrating mainly on event-based content reloading. I was planning on tackling job-based stuff, e.g. workflow and PRs later. The build bubbles in workflow do update themselves automatically, though.

app/event/model.js
minz1027: Cool!

PR #111, opened 2016-12-12
fix: Search button actually searches

minz1027:
- Make the search-icon clickable
- Now both `Enter` and clicking `search-icon` will take you to the search page 
- Resolve https://github.com/screwdriver-cd/screwdriver/issues/390

### DIFF
--- a/app/components/app-header/template.hbs
+++ b/app/components/app-header/template.hbs
@@ -7,10 +7,10 @@
 
       {{#if showSearch}}
       <div class="search">
-        <form class="search-form" action="/search" method="GET">
-          {{input class="search-input" name="query" placeholder="Enter your search terms" focus-in="onFocus" focus-out="onBlur"}}
+        <form class="search-form" {{action "triggerSearch" on="submit"}}>
+          {{input class="search-input" name="query" placeholder="Search Screwdriver pipelines " focus-in="onFocus" focus-out="onBlur"}}
         </form>
-        <div class="search-icon">
+        <div class="search-icon" {{action "triggerSearch"}}>
           {{fa-icon "fa-search"}}
         </div>
       </div>
--- a/tests/unit/application/controller-test.js
+++ b/tests/unit/application/controller-test.js
@@ -29,3 +29,14 @@ test('it calls session.authenticate', function (assert) {
   controller.send('invalidateSession');
   assert.equal(called, 1);
 });
+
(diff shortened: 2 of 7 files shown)


tests/unit/application/controller-test.js
```diff
@@ -29,3 +29,14 @@ test('it calls session.authenticate', function (assert) {
   controller.send('invalidateSession');
   assert.equal(called, 1);
 });
+
+test('it calls search in controller', function (assert) {
+  let controller = this.subject();
```
FenrirUnbound: I expected a `const` here since it's not being reassigned, but it's a `let`. I noticed the other tests also declare as `let controller`. Why isn't this declared a `const`?

FenrirUnbound:
When you click on the search icon, it transitions pretty well. But when you press `Enter` on the keyboard, it does a full page reload. Is that expected?

tests/unit/application/controller-test.js
d2lam: I guess that's because it needs to submit a form vs transition to the page directly

d2lam:
I guess that's because it needs to submit a form vs transition to the page directly

tests/unit/application/controller-test.js
minz1027: It is auto-generated. It can be const actually.

app/components/app-header/template.hbs
```diff
@@ -7,10 +7,10 @@
 
       {{#if showSearch}}
       <div class="search">
-        <form class="search-form" action="/search" method="GET">
-          {{input class="search-input" name="query" placeholder="Enter your search terms" focus-in="onFocus" focus-out="onBlur"}}
+        <form class="search-form" onsubmit={{action "triggerSearch"}}>
```
petey: If you do the following, does it prevent the full-page refresh?

```
{{action "triggerSearch" on="submit"}}
```

d2lam: Oh yeah that worked! Let me updated

petey: The way I suggested will allow ember to prevent the default behavior of the browser when the submit event fires.

tests/unit/application/controller-test.js
FenrirUnbound: Ah ok. Thanks! This was just a learning experience for me 😺

PR #113, opened 2016-12-13
feat: add number of search results and highlight query term

minz1027:
- Show number of search results
- Highlight query term

screen shot 2016-12-15 at 9 58 30 am

Related to https://github.com/screwdriver-cd/screwdriver/issues/394

### DIFF
--- a/app/components/search-list/styles.scss
+++ b/app/components/search-list/styles.scss
@@ -1,47 +1,55 @@
 & {
   width: 640px;
   margin: 2em auto;
+
+  a {
+    display: block;
+    text-decoration: none;
+    color: $sd-link;
+  }
+
+  .appId {
+    text-align: left;
+    width: 60%;
+  }
+
+  .branch {
+    width: 35%;
+    text-align: right;
+  }
+
+  tbody tr:hover {
+      background-color: $sd-separator;
+      .appId {
+        text-decoration: underline;
+      }
+  }
+
+  .num-results {
+    padding: 10px;
+    font-size: 14px;
+    color: $sd-text-med;
+  }
+
+  table {
+    width:100%;
+  }
+
+  th {
+    font-size: 14px;
+    text-transform: uppercase;
+    font-weight: 300;
+    color: $sd-text-med;
+  }
+
+  th, td {
+    padding: 10px;
+    border-bottom: 1px solid $sd-text-light;
+  }
 }
 
 @media (max-width: 47.999em) {
   & {
     width: 100%;
   }
-}
-
-ul {
-  list-style: none;
-  margin: 0;
-  padding: 0;
-}
-
-a {
-  display: block;
-  text-decoration: none;
-  color: $sd-link;
-  padding: 10px;
-  border-bottom: 1px solid $sd-separator;
-
-  .appId {
-    display: inline-block;
-    width: 400px;
-  }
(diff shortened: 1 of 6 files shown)


tkyi:
- Would be nice if there was some search term history information (if something was misspelled)
- Also some more metadata for search results (pipeline admin?, date created?)

app/components/search-list/template.hbs
```diff
@@ -1,11 +1,15 @@
+<div class="num-results">
+{{if isEmpty "No Results" (concat "Found " filteredPipelines.length " results") }}
```
minz1027: Using `result(s)`! yay!

app/components/search-list/styles.scss
```diff
@@ -21,18 +21,24 @@ a {
   color: $sd-link;
   padding: 10px;
   border-bottom: 1px solid $sd-separator;
+}
 
-  .appId {
-    display: inline-block;
-    width: 400px;
-  }
+.appId {
```
tkyi: Should this still be nested under `a`? Or at least be `a .appId {`?

app/components/search-list/styles.scss
```diff
@@ -21,18 +21,24 @@ a {
...
+  width: 400px;
+}
 
-  .branch {
-    display: inline-block;
-    width: 215px;
-    text-align: right;
-    color: $sd-text;
-  }
+.branch {
```
tkyi: same here

app/components/search-list/styles.scss
petey: No, this class applies to a span.

app/components/search-list/styles.scss
petey: This also applies to a span.

app/components/search-list/template.hbs
```diff
@@ -1,11 +1,21 @@
+<div class="num-results">
+  {{if isEmpty "No Results" (concat "Found " filteredPipelines.length " result(s)") }}
+</div>
+<div class="header">
```
petey: This is a good place to use a table. Would greatly simplify your css, too.

minz1027: @petey should we use table for just the header or both header and the records?

app/components/search-list/template.hbs
```diff
@@ -1,11 +1,21 @@
...
+  <span class="heading admin">Admin</span>
+</div>
 <ul>
   {{#each filteredPipelines as |pipeline|}}
   <li>
+    {{#highlight-terms query}}
     {{#link-to "pipeline" pipeline.id}}
       <span class="appId">{{pipeline.appId}}</span>
       <span class="branch">{{fa-icon "fa-code-fork"}}{{pipeline.branch}}</span>
+      <span class="admin">{{pipeline.admin}}</span>
```
stjohnjohnson: Please don't list admins, it's really not useful information on the search page.

app/components/search-list/template.hbs
minz1027: fixed

app/components/search-list/template.hbs
minz1027: fixed

d2lam: 😿

PR #114, opened 2016-12-14
fix: html entities in logs caused view issues

petey:
* Moves ansi colorizing to a htmlbars helper
* Use handlebars to sanitize raw logs before applying ansi colors
* Move html markup for log lines into template
* Reload logs on next tick if MAX_LOG_LINES is returned instead of waiting for BUILD_RELOAD_TIMER

Addresses https://github.com/screwdriver-cd/screwdriver/issues/395

### DIFF
--- a/app/helpers/ansi-colorize.js
+++ b/app/helpers/ansi-colorize.js
@@ -1,0 +1,15 @@
+import Ember from 'ember';
+
+/**
+ * Transform ansi color codes to html tags
+ * @method ansiColorize
+ * @param  {Array}     params   Values passed to helper
+ * @return {String}             Html string
+ */
+export function ansiColorize([message]) {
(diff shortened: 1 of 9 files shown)


app/helpers/ansi-colorize.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+/**
+ * Transform ansi color codes to html tags
+ * @method ansiColorize
+ * @param  {Array}     params   Values passed to helper
+ * @return {String}             Html string
+ */
+export function ansiColorize(params) {
+  const m = Ember.Handlebars.Utils.escapeExpression(params[0]);
```
d2lam: Hmmm why is params an array? Isn't it `log.m`? Is that a string or an array?

PR #115, opened 2016-12-15
feat(396): Add prev and next links to build banner

d2lam:
Add `Previous Job` and `Next Job` links to build banner.
Pass in builds of the current event and get the previous and next build IDs. If there is no previous job, it won't display. Same as next job.  

screen shot 2016-12-15 at 12 08 18 pm

screen shot 2016-12-15 at 12 08 26 pm

Related to https://github.com/screwdriver-cd/screwdriver/issues/396

### DIFF
--- a/app/components/build-banner/template.hbs
+++ b/app/components/build-banner/template.hbs
@@ -1,16 +1,24 @@
-<div class="line1">
-  <ul>
-    <li></li>
-    <li><h1>{{jobName}}</h1></li>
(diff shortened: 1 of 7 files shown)


app/components/build-banner/template.hbs
```diff
@@ -1,3 +1,9 @@
+<div class="line0">
```
d2lam: @petey: fixed

PR #140, opened 2017-02-06
fix: Remove white-space style in build logs

okuryu:
This fixes screwdriver-cd/screwdriver#449. This issue seems to occur on Safari and Firefox, not occur on Chrome. I verified it resolved by removing `white-space` style for build logs components. Line breaks disappear from some build logs by this change, but I think it's better for now.

### DIFF
--- a/app/components/build-log/styles.scss
+++ b/app/components/build-log/styles.scss
@@ -1,7 +1,6 @@
 .logs {
   font-size: 1em;
   font-family: monospace;
-  white-space: pre-wrap;
   background-color: #333;
   color: #eee;
   height: 0;


d2lam:
Wow! awesome! 👍

petey:
Hmm. I'm not sure that losing line breaks is worth this change. It has the potential to make logs really difficult to parse.

Before:
screen shot 2017-02-06 at 1 12 07 pm

After:
screen shot 2017-02-06 at 1 13 23 pm

okuryu:
yup, but it's same behavior with Travis CI. should replace `\n` with `<br>`?

petey:
That might work, but it may be tricky to do that and continue to keep the logs safe. This needs to be done without the potential for exposing XSS, and it also needs to be smart enough not to replace and escaped `\n` character, e.g. `\\n`.

stjohnjohnson:
@catto is there anything else we can do instead of changing the `white-space` attribute?

catto:
I think this issue is occurred by spaces around log elements. I'll check it and submit pull request.
We don't need to remove the `white-space` attribute if it works.

tkyi:
Closing in favor of #145 .
Thanks Catto!

PR #145, opened 2017-02-12
fix: Remove spaces around log elements

catto:
I removed spaces around log elements which has the `white-space: pre-wrap` attribute. The spaces include CR or LF, so I think it causes the issue (screwdriver-cd/screwdriver#449).
This PR is another option of #140.

Chrome
image

FireFox (after)
image
FireFox (before)
image

Opera
image


okuryu:
👍 it's probably better.

petey:
Nice! Thanks @catto!

catto:
🐈

PR #146, opened 2017-03-09
feat(468): add artifact tree to build page

minz1027:
Add artifact tree to build page.
- There is a component `artifact-tree`
- Inside `artifact-tree`, it will call service `build-artifact` to fetch manifest data to build the tree
- Use [ember-cli-jstree](https://github.com/ritesh83/ember-cli-jstree) to build tree
- Added `SDSTORE_HOSTNAME` and  `SDSTORE_NAMESPACE` in config

screen shot 2017-03-08 at 6 42 05 pm

related to https://github.com/screwdriver-cd/screwdriver/issues/468


bower.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "ansi_up": "^1.3.0",
     "humanize-duration": "^3.9.1",
-    "jquery-highlight": "~3.2.2"
+    "jquery-highlight": "~3.2.2",
+    "jstree": "^3.3.3"
```
minz1027: ember-cli-jstree need this as dependency.

tkyi:
So each build page will have its own artifacts section?
What happens if you click on an artifact?

d2lam:
@tkyi  Yeah, artifacts are per build. And each one is a link, it should take you to the artifact link.

app/build-artifact/service.js
```diff
@@ -0,0 +1,98 @@
...
+  let parts;
+
+  if (!tree.children) {
+    parts = tree;
+  } else {
+    parts = tree.children;
+  }
+
+  parts.forEach((part) => {
+    if (part.children.length === 0) {
```
FenrirUnbound: What happens with empty directories?

app/build-artifact/service.js
```diff
@@ -0,0 +1,98 @@
...
+      part.type = 'file';
+      delete part.children;
+    } else {
+      delete part.a_attr;
```
FenrirUnbound: Why is this significant?

d2lam: We only what it to be the link to the store if it's a file. If it's a directory then remove the link and call changeFiles recursively.

FenrirUnbound: Could that be documented in one of the function docs? It's not clear whether that's an intended feature

app/build-artifact/service.js
minz1027: Is it a common case? Someone just create a folder under ARTIFACTS and do no write anything in it? If that's the case, there is no way for the UI to tell it's a file or directory. We don't provide any extra info in the manifest, `/emptydir` can be a dir and can be a file. File is the best guess here.

FenrirUnbound: It's not rare, but it's an uncommon scenario in V3.

What happens to the link if it's an empty directory?

minz1027: You get 404.

FenrirUnbound: That behavior should be documented or captured in an issue

PR #147, opened 2017-03-11
fix(478): redirect to a pipeline page if a build doesn't belong to the pipeline

catto:
I fixed a bug described in https://github.com/screwdriver-cd/screwdriver/issues/478.
There are 2 ways to avoid this bug.
1. show 404 page
2. redirect to somewhere
  2.1 `/pipelines/:pipeline_id`
  2.2 `/pipelines/:correct_pipeline_id/build/:build_id`

I think this will happen much when the end of the URL is trimmed such as copy/paste failure, so I chose the method 2.1.

### DIFF
--- a/tests/unit/pipeline/builds/build/route-test.js
+++ b/tests/unit/pipeline/builds/build/route-test.js
@@ -34,5 +34,5 @@ sinonTest('it redirects if build not found', function (assert) {
   route.afterModel(model);
 
   assert.ok(stub.calledOnce, 'transitionTo was called once');
-  assert.ok(stub.calledWithExactly('pipeline', pipelineId), 'transtion to pipeline');
+  assert.ok(stub.calledWithExactly('pipeline', pipelineId), 'transition to pipeline');
 });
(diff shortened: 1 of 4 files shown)


tests/unit/pipeline/builds/build/route-test.js
```diff
@@ -15,3 +16,23 @@ test('it exists', function (assert) {
...
+    },
+    job: {
+      get: type => (type === 'id' ? jobId : null)
+    }
+  };
+
+  route.afterModel(model);
+
+  assert.ok(stub.calledOnce, 'transitionTo was called once');
+  assert.ok(stub.calledWithExactly('pipeline', pipelineId), 'transtion to pipeline');
```
tkyi: small typo: transition*

catto: Fixed!

minz1027:
@catto is there any reason that we want to introduce the new test framework `ember-sinon-qunit`?

catto:
@minz1027 `ember-qunit` doesn't support sinon's stub directly. I think we can read the tests easily if we use `ember-sinon-qunit` instead of using `sinon` and `ember-qunit` separately.
`ember-sinon-qunit` has the compatibility with `ember-qunit`, so I can replace them. Do I need to do so? I think it's not a scope of this PR, maybe later.

PR #148, opened 2017-03-11
feat(425): Redirect to previous page after login

catto:
I added a feature described in screwdriver-cd/screwdriver#425. 
I haven't added tests yet, but could you review the changes if you have a chance? @petey

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -13,6 +13,18 @@ App = Ember.Application.extend({
   Resolver
 });
 
+// Detect and store previous path for redirection after login
+Ember.Route.reopen({
+  afterModel: function(resolvedModel, transition) {
+    transition.then(() => {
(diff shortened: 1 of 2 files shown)


minz1027:
This implementation will cover most of the cases, except this:
1. User clicks any route requires authentication, using `'ember-simple-auth/mixins/authenticated-route-mixin'`, e.g. `create`, `secrets`
2. Click other routes that do not require that, e.g. `search`, `builds`
3. Click login 
4. Users will not be redirected to the previous route, instead, users will be redirected to the one using `authenticated-route-mixin`

But such case is kinda rare. I am okay with how it works now.

petey:
Seems ok

catto:
thanks!

PR #152, opened 2017-04-05
feat: Add validator UI

petey:
Provides a way to copy in a configuration and see if anything is wrong.

### DIFF
--- a/app/validator/controller.js
+++ b/app/validator/controller.js
@@ -1,0 +1,33 @@
+import Ember from 'ember';
+
+/**
+ * Fetches validator results in debounce
+ * @method getResults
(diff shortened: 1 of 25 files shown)


app/validator/controller.js
```diff
@@ -0,0 +1,33 @@
...
+  onYamlChange: Ember.observer('yaml', function onYamlChange() {
+    const yaml = this.get('yaml').trim();
+
+    if (!yaml) {
+      this.set('results', '');
+
+      return;
+    }
+
+    Ember.run.debounce(this, getResults, 250);
```
tkyi: So fancy 🎩

PR #156, opened 2017-05-03
fix: Add instructions to the create and secrets pages

ian-fox:
Added instructions to the "create a pipeline" and secrets pages as outlined in https://github.com/screwdriver-cd/screwdriver/issues/489. Also added a title to the secrets page.

screen shot 2017-05-03 at 1 12 27 pm
screen shot 2017-05-03 at 1 23 02 pm


app/components/pipeline-secret-settings/template.hbs
```diff
@@ -1,4 +1,13 @@
 {{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
+<h1>Secrets</h1>
+<p>
+  User secrets are stored in an encrypted database, and their values are only released to those builds that are authorized by the user.<br>
+  Secret keys can only consist of uppercase alphabetic characters and numbers.<br>
```
r3rastogi: I think it'll be better to have all three of these sentences as one paragraph, as opposed to three separate lines.

tkyi: I kind of like how the sentences are broken up. I think it's generally a bad user experience to have text spanning the entire page, because they'll have to move their eyes all the way across. But maybe @petey would know better?

r3rastogi: The first line is already long enough that on many screens, a user will be scanning all the way to the very end anyways. I'd agree with the fragmentation if we were writing bullets, but I disagree with it if we're keeping a full-sentence structure. At the very least, I would combine the first two lines, but preferably all three.

tkyi: Hm I feel like the first line is more general information about Screwdriver Secrets while the other two are just tips/instructions for the Secrets form.

tkyi:
Might be good to change the placeholders in the Secrets form to have better examples (maybe an all caps key example and a fake secret?).
https://github.com/screwdriver-cd/ui/blob/d34cc492d5018a7279edd8d21eaf187fa51fe89f/app/components/pipeline-secret-settings/template.hbs#L17-L18

app/components/pipeline-secret-settings/template.hbs
```diff
@@ -1,4 +1,13 @@
...
+  The "Allow in PR" option will expose the key in Pull Request builds.
```
petey: This isn't entirely accurate. The "allow in PR" checkbox only flags a secret as being allowed to be exposed in a pull-request, but it does not explicitly expose it. The user still needs to add the yaml configuration [to `main`] to expose them just like any other key.

app/components/pipeline-secret-settings/template.hbs
r3rastogi: Ehh I'd still rather have more of the vertical space dedicated to actual content rather than description

app/components/pipeline-secret-settings/template.hbs
```diff
@@ -1,4 +1,13 @@
...
+</p>
+<p>
+  For usage, see the <a href="http://docs.screwdriver.cd/user-guide/configuration/secrets/">user guide</a>.
```
petey: I kinda feel like we should just have a link next to the heading using http://fontawesome.io/icon/question-circle/ to link to the docs, instead of this text.

app/components/pipeline-secret-settings/template.hbs
petey: I'm really torn on whether these instructions should be on this page at all, at least not being right in the users face. The documentation clearly points out how this is supposed to work.

We could make the experience better by adding the rules as `title` attributes on the individual inputs, so they would appear as tooltips. Since we are modifying those elements anyways, we should probably make the placeholder text valid as well.

PR #161, opened 2017-05-05
Added lines to load the new lua module in nginx.conf

jer:
- Borrowed from here https://github.com/openresty/lua-nginx-module#installation
- I don't have any idea if this will work in the intended way, but I do
  know that nginx starts with this config in place.


FenrirUnbound:
For others following this: I'm assuming the `load_module` definitions came from here https://github.com/openresty/lua-nginx-module#building-as-a-dynamic-module

PR #164, opened 2017-06-01
fix: Add new header and navbar

tkyi:
### Notes:
- Rewrite HTML to get rid of purecss.io, since we don't think it's worth the effort anymore
- Use ember-bootstrap module
- Remove text as a part of the new spec
- Add formatting when search bar is active
- Make menu more dynamic/reactive
- Fix icon spacing; also have them float right

Current look:
image

When Search Bar is active:
image

When window is minimized:
image


app/components/app-header/styles.scss
```diff
@@ -1,178 +1,112 @@
...
+      background: -moz-linear-gradient(bottom left, #3570F4, #3697F2); /* For Firefox 3.6 to 15 */
+      background: linear-gradient(to bottom left, #3570F4, #3697F2); /* Standard syntax */
+      border-style: none;
+
+      .fa {
+        font-size: 15px;
+        font-weight: 300;
+      }
+
+      &:hover {
```
tkyi: Do we need this `&` here?

package.json
```diff
@@ -32,15 +32,17 @@
     "Darren Matsumoto <aeneascorrupt@gmail.com>",
     "Jeremiah Wuenschel <jeremiah.wuenschel@gmail.com>",
     "Min Zhang <minzhang@andrew.cmu.edu>",
-    "Noah Katzman <nbkatzman@gmail.com>",
```
tkyi: Are we removing Noah from all our repos?

app/components/app-header/template.hbs
```diff
@@ -1,68 +1,63 @@
...
+      {{/nav.item}}
+      {{#nav.item}}
+        <a href="https://github.com/screwdriver-cd" title="Screwdriver Github repo" class="icon github">
+          {{inline-svg "github" class="img"}}
+          <span class='icontitle'>Github</span>
+        </a>
+      {{/nav.item}}
+      {{#nav.item}}
+        {{#if session.isAuthenticated}}
+          <a {{action 'invalidateSession'}} title="Log out of Screwdriver" class="icon auth">
```
tkyi: Nice 👍

tests/integration/components/app-header/component-test.js
```diff
@@ -5,7 +5,8 @@ moduleForComponent('app-header', 'Integration | Component | app header', {
   integration: true
 });
 
-test('it renders', function (assert) {
+// this test should pass when search bar feature flag is turned off
```
ian-fox: I'm not super experienced with feature flags, how long do they stay around after they've been switched on? Is there any time when we'd turn it off?

petey: They would stick around for as long as we decide to keep the flag there. We could potentially want to turn this off if we determine that search causes unacceptable performance problems. At least until we would fix those issues.

package.json
petey: I don't really see why.

app/components/app-header/styles.scss
petey: In SASS, `&` always refers to the parent element: http://sass-lang.com/documentation/file.SASS_REFERENCE.html#parent-selector

So at the root of the file, `&` refers to the root element of the component. In this case, it refers to `.navbar-form button`

tests/integration/components/app-header/component-test.js
ian-fox: Makes sense.

package.json
tkyi: Can we leave him in this one?

r3rastogi: Nah, we don't have to. I was just adding me, and he hasn't contributed to this repo. But I can modify this to put him back in if you want

r3rastogi: @tkyi sure. I'll update

dvdizon:
👍

PR #166, opened 2017-06-07
feat: remove purecss

petey:
Context:
========
We added bootstrap in the previous app-header pr. There's no good reason to have both purecss and bootstrap, and the ember-bootstrap module does more.

Objective:
----------
Remove everything to do with purecss from the code.

screen shot 2017-06-07 at 11 10 30 am


app/components/build-bubble/styles.scss
```diff
@@ -48,11 +48,7 @@
       border: 2px solid $sd-text-light;
       width: 16px;
       height: 16px;
-      font-size: 12px;
-
-      i {
-        line-height: 16px;
-      }
+      font-size: 9px;
```
ian-fox: Does bootstrap scale text down, or is this another change that happens to come at the same time as removing purecss?

petey: I'm not sure what caused it, but the build bubble icons were no longer centered, and the quick fix was to reduce the font size.

app/components/pipeline-nav/styles.scss
```diff
@@ -1,27 +1,30 @@
 & {
-  padding-top: 5px;
-  background: url('/assets/main_bkgrd_pipeline_bottom.png') transparent no-repeat;
-  background-size: 100%;
-  height: 39px;
-}
+  padding-left: 25px;
+  background-color: $sd-dark-blue;
```
ian-fox: With bootstrap is the component too small to make use of that image?

petey: We don't want to use those graphics anymore.

ian-fox: Ah, that also makes sense. That was the only other thing that stood out to me, looks good!

r3rastogi: Should we also remove those graphics in this PR then?

r3rastogi: ^ nvm! Didn't scroll down far enough to see that you already did :)

PR #165, opened 2017-06-06
feat: add alpha banner with feature flag

petey:
Context:
========
New users to the software should not expect this to be feature complete.

Objective:
----------
Display a banner (using a feature flag) to tell people about the alpha status.

Requires users set an environment variable `FEAT_DISPLAY_ALPHA` as 'true' when using the docker image for the UI.

screen shot 2017-06-06 at 4 17 51 pm

PR #167, opened 2017-06-07
Revert "feat: add alpha banner with feature flag"

petey:
Reverts screwdriver-cd/ui#165

For an unknown reason, `FEAT_DISPLAY_ALPHA` is being set to `nil` in the nginx config regardless of the value of the environment variable.

### DIFF
--- a/app/components/app-header/template.hbs
+++ b/app/components/app-header/template.hbs
@@ -68,4 +68,3 @@
     {{/navbar.nav}}
   {{/navbar.content}}
 {{/bs-navbar}}
-{{#bs-alert visible=showAlphaWarning type="warning"}}This is alpha software. Please use at your own risk.{{/bs-alert}}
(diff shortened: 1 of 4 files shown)


app/components/app-header/template.hbs
```diff
@@ -67,5 +67,4 @@
       {{/nav.item}}
     {{/navbar.nav}}
   {{/navbar.content}}
-{{/bs-navbar}}
-{{#bs-alert visible=showAlphaWarning type="warning"}}This is alpha software. Please use at your own risk.{{/bs-alert}}
+{{/bs-navbar}}
```
ian-fox: Does eslint complain about the lack of newline?

ian-fox: Oh wait this is a handlebars file nevermind

PR #168, opened 2017-06-08
fix(584): Make search bar smaller and make pipeline page tabs easier to read

tkyi:
Addressing a few of the issues in https://github.com/screwdriver-cd/screwdriver/issues/584

 - [x] Start button is offset out of the page -> Start button has been moved to the left side of the screen
 - [x] Pipeline search bar takes up too much space -> changed to 326px
 - [x] Color of tabs on Pipeline page hard to read -> made them brighter
 - [x] Header switches into hamburger under 1230px (which is pretty big) -> switches at 962px now
 - [x] Header looks crazy at exactly 1230px -> no more crazy

Also in this PR:
- [x] Made Search icon centered in button
- [x] Hovering over a Pipeline page tab will make the background change color so it's more clear that it's clickable and is supposed to act like a tab

Before:
screen shot 2017-06-07 at 5 04 03 pm

After:
screen shot 2017-06-07 at 5 01 10 pm

New Start button position:
screen shot 2017-06-07 at 6 24 59 pm

### DIFF
--- a/app/components/app-header/styles.scss
+++ b/app/components/app-header/styles.scss
@@ -71,6 +71,7 @@
       .fa {
         font-size: 15px;
         font-weight: 300;
+        padding-left: 5px;
       }
 
       &:hover {
@@ -101,12 +102,12 @@
   }
 }
 
-@media screen and (min-width: 1231px) {
+@media screen and (min-width: 962px) {
   .navbar-default .navbar-nav > li > a.icon .icontitle {
     display: none;
   }
 
   .navbar-form input {
-    width: 522px;
+    width: 326px;
   }
 }
--- a/app/components/pipeline-nav/styles.scss
+++ b/app/components/pipeline-nav/styles.scss
@@ -6,10 +6,10 @@
     a {
       border-radius: 2px 2px 0 0;
       background-color: transparent;
-      color: $sd-text-med;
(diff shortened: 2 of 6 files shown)


tkyi:
I know the search bar isn't to spec, but we could change it back after the demo?

petey:
I talked to jay and he said we should reduce the search bar size. 👍

tkyi:
Is the size I chose okay? Or should we go **smaller**?

petey:
It's 326px in the dropdown....

Also, you can change: https://github.com/screwdriver-cd/ui/blob/master/app/styles/app.scss#L39 to reduce the page size before the hamburger shows up.
Also change: https://github.com/screwdriver-cd/ui/blob/master/app/components/app-header/styles.scss#L104 to match.

tkyi:
I saw the hamburger thingy but I wasn't sure what number makes sense. Also why do the Icon titles suddenly show up?

petey:
I added bootstrap tooltips.

r3rastogi:
Jay said 400px for the search bar, but I'm not sure if he just made up a number or that's the new spec for us

tkyi:
He probably just made that up. We can update the size with a subsequent PR some other time.

PR #170, opened 2017-06-09
fix(589): fix calls to reload() on Array

petey:
Addresses: https://github.com/screwdriver-cd/screwdriver/issues/589

### DIFF
--- a/app/pipeline/builds/index/template.hbs
+++ b/app/pipeline/builds/index/template.hbs
@@ -20,7 +20,7 @@
 
 <div class="row">
   <div class="col-xs-12 col-md-9">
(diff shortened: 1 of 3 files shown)


app/pipeline/builds/index/template.hbs
```diff
@@ -36,4 +36,4 @@
   {{/modal-dialog}}
 {{/if}}
 
-{{outlet}}
+{{outlet}}
```
ian-fox: Missing a newline here (if that's something we worry about), otherwise everything looks good to me :+1:

petey: Should not matter in templates.

PR #171, opened 2017-06-16
fix: add a dropdown for communications links

petey:
Context:
========
There's lots of icons in the header making it crowded. There are more icons desired.

Objective:
----------
Put all the "communication" related icons under the "other" circle as designated by our mocks.

Dropdown menu:
screen shot 2017-06-16 at 3 00 33 pm

With reduced screen width:
screen shot 2017-06-16 at 3 00 49 pm


tkyi:
Not going to put in the new icons into this PR (for Slack, Create, Blog/Tumblr)?

petey:
I needed to massage the icons we got a bit as they had embedded fill styles.

tkyi:
The icons look like they're not centered vertically(?) on the header bar.
Otherwise, LGTM

petey:
I don't know that they are centered vertically, but they are aligned:

screen shot 2017-06-16 at 4 33 05 pm

app/components/app-header/template.hbs
```diff
@@ -17,35 +17,43 @@
...
-        </a>
+        {{#nav.dropdown as |dd|}}
+          {{#dd.toggle}}{{inline-svg "more-circle" class="img"}}
+            <span class='icontitle'>Communication</span>
+            <span class="caret"></span>
+          {{/dd.toggle}}
+          {{#dd.menu as |ddm|}}
+            {{#ddm.item}}
+              <a href="http://blog.screwdriver.cd" class="icon blog">
+                {{inline-svg "tumblr" class="img"}} Screwdriver Blog
```
d2lam: I don't think we need to put `Screwdriver` here, to be consistent with other options.

app/components/app-header/template.hbs
```diff
@@ -17,35 +17,43 @@
...
-          {{inline-svg "blog" class="img"}}
-          <span class='icontitle'>Screwdriver Blog</span>
-        </a>
-      {{/nav.item}}
-      {{#nav.item}}
-        <a href="http://slack.screwdriver.cd" class="icon community">
-          {{inline-svg "community" class="img"}}
-          <span class='icontitle'>Slack Channel</span>
-        </a>
+        {{#nav.dropdown as |dd|}}
```
ian-fox: Some of the duplicate styling could be avoided by replacing `{{nav.item}}` with `{{#nav.dropdown tagName="li" as |dd|}}`, though that wouldn't have the properties/methods of a [`nav.item`](http://www.ember-bootstrap.com/api/classes/Components.NavItem.html) like "active".

Those properties I think are preferable to avoiding some duplicate styling. If there were a way to have the component be both a `nav.item` and a `nav.dropdown` that'd probably be ideal, but ember doesn't seem to do multiple inheritance. 

All in all I think your version is better :+1:

tkyi:
@petey Oh I see. Was that in the spec? Just looks like there's more space under the icons than above. We can fix it in another PR though as long as that's not a design no-no or something.

petey:
Our spec shows them bottom aligned with the logo text and the height of the header, but not the size of the images. We can tweak the pixels later.

tkyi:
Okay 👍

petey:
Now aligned better:
screen shot 2017-06-19 at 11 36 47 am

PR #173, opened 2017-06-19
fix: pin major and minor version for ember-data to fix tests for now

minz1027:
Unit tests are failing due to the new ember-data version `v2.14.x` published during the weekend.
https://github.com/emberjs/data/releases

Failed build: https://cd.screwdriver.cd/pipelines/7/builds/5805

Unblock  the pipeline by pinning major and minor for now. Will create a sticky and figure out how to actually resolve it.


petey:
Good catch.

PR #175, opened 2017-06-29
fix(608): Visual difference for screwdriver-provided steps

joelseq:
Make screwdriver provided steps a grey-ish color to easily distinguish between user-defined steps. Related to [screwdriver-cd/screwdriver#608](https://github.com/screwdriver-cd/screwdriver/issues/608)

Screenshot of what it looks like:
image

### DIFF
--- a/app/components/build-step-view/component.js
+++ b/app/components/build-step-view/component.js
@@ -4,6 +4,11 @@ export default Ember.Component.extend({
   classNames: ['step-view'],
   classNameBindings: ['status'],
   isOpen: false,
+  isSdStep: Ember.computed('stepName', {
+    get() {
+      return this.get('stepName').match(/^sd-/);
+    }
+  }),
 
   /**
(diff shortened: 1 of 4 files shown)


tkyi:
Seems like a good first stab. Not sure if it gets confusing since all of the times are also in the same gray color as the grayed out SD steps. Could lead users to ignore everything except step names?

tkyi:
@jpcollins what do you think?

jpcollins:
I think it's improved.  Let's roll with it for now.

app/components/build-step-view/component.js
```diff
@@ -4,6 +4,11 @@ export default Ember.Component.extend({
   classNames: ['step-view'],
   classNameBindings: ['status'],
   isOpen: false,
+  isSdStep: Ember.computed('stepName', {
```
minz1027: Should at least test this function

PR #176, opened 2017-06-30
feat(tokens): Add tokens UI

ian-fox:
* login/out item in the menu bar is now a dropdown when logged in which points to a "user-settings" page
* added user-settings page which displays a user's tokens and allows for creation, updating, refreshing, and revoking
* made the title of the "secrets" page smaller to match the tokens page (following feedback that it was too large)

Related to https://github.com/screwdriver-cd/screwdriver/issues/532
Resolves https://github.com/screwdriver-cd/screwdriver/issues/615

screen shot 2017-06-30 at 1 04 19 pm
screen shot 2017-06-30 at 1 04 35 pm
screen shot 2017-06-30 at 1 48 32 pm
screen shot 2017-06-30 at 1 48 41 pm

### DIFF
--- a/app/components/pipeline-secret-settings/template.hbs
+++ b/app/components/pipeline-secret-settings/template.hbs
@@ -1,10 +1,10 @@
 {{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
-<h1>
+<h3>
   Secrets
   <a href="http://docs.screwdriver.cd/user-guide/configuration/secrets/">
     <i class="fa fa-question-circle" title="More information"></i>
   </a>
-</h1>
+</h3>
 <p>
   User secrets must also be added to the Screwdriver YAML.
 </p>
--- a/app/components/token-list/component.js
+++ b/app/components/token-list/component.js
@@ -1,0 +1,118 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  tokenSorting: ['name', 'description', 'lastUsed'],
+  sortedTokens: Ember.computed.sort('tokens', 'tokenSorting'),
+
+  // Error for failed to create/update/remove/refresh token
+  error: null,
+  errorMessage: Ember.computed('error', function errorMessage() {
+    const error = this.get('error');
+
+    return error ? error.errors[0].detail : '';
+  }),
+
+  // Adding a new token
+  isButtonDisabled: Ember.computed('newName', 'isSaving', function isButtonDisabled() {
+    return !this.get('newName') || this.get('isSaving');
+  }),
+  isSaving: false,
+  newDescription: null,
+  newName: null,
+
+  // Confirmation dialog
+  isShowingModal: false,
+  modalAction: null,
+  modalButtonText: Ember.computed('modalAction', function modalButtonText() {
+    return Ember.String.capitalize(this.get('modalAction'));
+  }),
+  modalTarget: null,
+  modalText: null,
+
+  // Don't show the "new token" and "error" dialogs at the same time
+  errorObserver: Ember.observer('error', function errorObserver() {
+    if (this.get('error')) {
+      this.set('newToken', null);
+      this.set('isSaving', null);
+    }
+  }),
+  newTokenObserver: Ember.observer('newToken', function newTokenObserver() {
+    if (this.get('newToken')) {
+      this.set('error', null);
+      this.set('isSaving', null);
+    }
+  }),
+
+  actions: {
+    /**
+     * Kicks off create token flow
+     * @method addNewToken
+     */
+    addNewToken() {
+      const newName = this.get('newName');
+
+      this.set('isSaving', true);
+
+      return this.get('onCreateToken')(newName, this.get('newDescription'))
+        .then(() => {
+          this.set('newName', null);
+          this.set('newDescription', null);
+        }).catch((error) => {
+          this.set('error', error);
+        });
+    },
+    /**
+     * Set the error to be displayed from child components
+     * @param {Object} error
+     */
+    setError(error) {
+      this.set('error', error);
+    },
+    /**
+     * Show or hide the saving modal from child components
+     * @param {Boolean} isSaving
+     */
+    setIsSaving(isSaving) {
+      this.set('isSaving', isSaving);
+    },
+    /**
+     * Confirm an action
+     * @method confirmAction
+     * @param {String} action   One of "refresh" or "revoke"
+     * @param {Number} id
+     */
+    confirmAction(action, id) {
+      this.set('modalTarget', this.get('tokens').find(token => token.get('id') === id));
+      this.set('modalAction', action);
+
+      if (action === 'delete') {
+        this.set('modalText', `The "${this.get('modalTarget.name')}" token will be deleted.`);
+      } else {
+        this.set('modalText',
+          `The current "${this.get('modalTarget.name')}" token will be invalidated.`);
+      }
+
+      this.set('isShowingModal', true);
+    },
+    /**
+     * Close the modal, calling a callback if necessary
+     * @method closeModal
+     * @param {Boolean} confirm
+     */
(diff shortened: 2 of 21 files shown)


app/components/app-header/template.hbs
```diff
@@ -73,7 +84,7 @@
 <div class="tooltips">
   {{#bs-tooltip placement="bottom" triggerElement=".icon.create" renderInPlace=true}}Create a new Pipeline{{/bs-tooltip}}
   {{#bs-tooltip placement="bottom" triggerElement=".icon.docs" renderInPlace=true}}Screwdriver Documentation{{/bs-tooltip}}
-  {{#bs-tooltip placement="bottom" triggerElement=".icon.auth" renderInPlace=true}}
-    {{#if session.isAuthenticated}}Log out of Screwdriver{{else}}Login to Screwdriver{{/if}}
-  {{/bs-tooltip}}
-</div>
\ No newline at end of file
+  {{#if not session.isAuthenticated}}
```
petey: I'm not really sure that `not` is available without installing `ember-truth-helpers`.  Even so, this should probably be `{{#if (not session.isAuthenticated)}}`.

screwdriver.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
       - setup-ci: git clone https://gist.github.com/3d2388b2a7ba658cdcdaffa8cd874e50.git ci
       - install: npm install
       - bower: npm install bower && ./node_modules/.bin/bower install --allow-root
-      - build: ./node_modules/.bin/ember build --environment production
+      - build: ./node_modules/.bin/ember build
```
petey: I'm pretty sure we want to package the production version of the code.

app/components/token-list/component.js
```diff
@@ -0,0 +1,100 @@
...
+
+export default Ember.Component.extend({
+  tokenSorting: ['name', 'description', 'lastUsed'],
+  sortedTokens: Ember.computed.sort('tokens', 'tokenSorting'),
+
+  // Error for failed to create/update/remove/refresh token
+  errorMessage: Ember.computed('error', function errorMessage() {
+    const error = this.get('error');
+
+    if (!error) {
```
petey:
```
return error ? error.errors[0].detail || '';
```

app/components/token-list/component.js
```diff
@@ -0,0 +1,100 @@
...
+  actions: {
+    /**
+     * Kicks off create token flow
+     * @method addNewToken
+     */
+    addNewToken() {
+      const newName = this.get('newName');
+      const createPromise = this.get('onCreateToken')(newName, this.get('newDescription'));
+
+      createPromise.then(() => {
```
petey: You may want to return this promise.

app/components/token-list/component.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (!error) {
+      return '';
+    }
+
+    return error.errors[0].detail;
+  }),
+
+  // Adding a new token
+  isButtonDisabled: true,
```
petey: `Ember.computed.empty('newToken')`

app/components/token-list/component.js
```diff
@@ -0,0 +1,100 @@
...
+        this.set('newName', null);
+        this.set('newDescription', null);
+        this.set('isButtonDisabled', true);
+      });
+    },
+    /**
+     * Sets disabled state of "add" button
+     * @method enableButton
+     */
+    enableButton() {
```
petey: Not necessary. You can use Ember.computed for this.

app/components/token-list/style.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  .new td {
+    padding: 5px;
+    background-color: $sd-text-light;
+  }
+
+  tfoot button {
+    border-width: 0;
+    width: 100px;
+    background-color: #0078ff;
```
petey: All colors should be defined in screwdriver-colors.scss

ember-cli-build.js
```diff
@@ -7,6 +7,9 @@ module.exports = function (defaults) {
     'ember-font-awesome': {
       useScss: true // for ember-cli-sass
     },
+    minifyJS: {
+      enabled: false
```
petey: nope.

screwdriver.yaml
ian-fox: Oh whoops this was for debugging, sorry! Good catch!

ember-cli-build.js
ian-fox: Sorry, more debugging

ian-fox:
Tagging as WIP because I found a bug while fixing the stuff @petey commented on

tests/integration/components/token-view/component-test.js
```diff
@@ -0,0 +1,90 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import Ember from 'ember';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('token-view',
+'Integration | Component | token view', {
+  integration: true
+});
```
petey: The indentation here looks really weird. How about:
```
moduleForComponent(
  'token-view',
  'Integration | Component | token view', 
  { integration: true }
);
```

ian-fox: Looked through some other tests, turns out just about all of them have everything besides the `integration: true` on one line (with the notable exceptions of one where the line was too long,  the one I copied the style of, and another in the same style)

Fixed the two inconsistent ones (and mine) to match the rest of the tests, and the line-length one to be as suggested.

app/components/pipeline-secret-settings/template.hbs
```diff
@@ -1,10 +1,10 @@
 {{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
-<h1>
+<h3>
```
ian-fox: Changing this to be consistent with the user-settings page following feedback that the title was too big.

app/components/token-view/component.js
```diff
@@ -0,0 +1,56 @@
...
+
+export default Ember.Component.extend({
+  tagName: 'tr',
+  buttonAction: 'Delete',
+  newName: null,
+  newDescription: null,
+  lastUsedPretty: Ember.computed('lastUsed', function lastUsedPretty() {
+    const lastUsed = this.get('token.lastUsed');
+
+    return lastUsed ? fromNow(lastUsed)._getter() : 'Never used';
```
ian-fox: Calling an underscored method seems wrong, but I can't find another way to get the value out of this. Without calling `_getter` it just displays as `object Object`.

petey: Use the template helper to handle it, then.
```
{{#if lastused}}{{moment-from-now lastused}}{{else}}Never used{{/if}}
```

petey: https://www.npmjs.com/package/ember-moment

app/components/token-list/component.js
```diff
@@ -0,0 +1,118 @@
...
+    addNewToken() {
+      const newName = this.get('newName');
+
+      this.set('isSaving', true);
+
+      return this.get('onCreateToken')(newName, this.get('newDescription'))
+        .then(() => {
+          this.set('newName', null);
+          this.set('newDescription', null);
+        }, (error) => {
```
tkyi: Should this be `}).catch((error) => {` ?

ian-fox: Either one works, you can either use `.then(successHandler).catch(errorHandler)` or `.then(successHandler, errorHandler)`

tkyi: Oh okay. Just wondering since I didn't see it being used this way anywhere else.

ian-fox: The difference between those being (I believe) in what happens if the successHandler throws an error, but that shouldn't happen here since the only thing it's doing is setting the values of those two variables.

PR #177, opened 2017-06-30
feat(562): Allow users to update checkout URL for pipeline

tkyi:
## Context

Users would like to be able to update their pipeline repo and branch from the Screwdriver UI.

## Objective

This PR will add a pipeline section to Options tab and allow users to update checkout URL. There will also be validation to pipeline checkout URL update. There is also a one-liner added to README instructions for running a single test with ember.

**Invalid checkoutURL:**
screen shot 2017-06-30 at 12 51 21 pm

**Valid checkoutURL:**
screen shot 2017-06-30 at 12 51 09 pm

**Note:** The entire Pipeline options page definitely needs a redesign. Leaving those changes for another PR/time.

## Relevant Links

- Original issue: https://github.com/screwdriver-cd/screwdriver/issues/562
- API PR: https://github.com/screwdriver-cd/screwdriver/pull/627

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ Make use of the many generators for code, try `ember help generate` for more det
 * `ember test`
 * `ember test --server`
 
+To run a single ember test:
+* `ember t -s -m '<TEST_NAME>'`   // e.g. ember t -s -m 'Integration | Component | pipeline options'
+
 ### Building
 
 * `ember build` (development)
--- a/app/components/pipeline-options/component.js
+++ b/app/components/pipeline-options/component.js
@@ -1,13 +1,50 @@
 import Ember from 'ember';
+import { parse, getCheckoutUrl } from '../../utils/git';
 
 export default Ember.Component.extend({
-  showDangerButton: true,
-  showRemoveButtons: false,
+  // Updating a pipeline
+  init() {
+    this._super(...arguments);
+    this.set('scmUrl', getCheckoutUrl({
+      appId: this.get('pipeline.appId'),
+      scmUri: this.get('pipeline.scmUri')
+    }));
+  },
+  errorMessage: '',
+  scmUrl: '',
+  isValid: Ember.computed('scmUrl', {
+    get() {
+      const val = this.get('scmUrl');
+
+      return val.length !== 0 && parse(val).valid;
+    }
+  }),
+  isInvalid: Ember.computed.not('isValid'),
+  isDisabled: Ember.computed.or('isSaving', 'isInvalid'),
+  // Removing a pipeline
   isRemoving: false,
   isShowingModal: false,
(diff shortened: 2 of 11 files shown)


app/components/pipeline-options/styles.scss
```diff
@@ -120,3 +120,88 @@
...
+  &.bad-text-input {
+    border-color: $sd-failure;
+  }
+
+  &.good-text-input {
+    border-color: $sd-success;
+  }
+}
+
+.blue-button {
```
ian-fox: Is there some way we can get this to use the styling from the place this got copy/pasted from? Then we don't have to worry about having the same code in two places.

tkyi: I agree. Not sure how to do it

README.md
```diff
@@ -35,6 +35,9 @@ Make use of the many generators for code, try `ember help generate` for more det
 * `ember test`
 * `ember test --server`
 
+To run a single ember test:
+* `ember t -s -m '<TEST_NAME>'`   // e.g. ember t -s -m 'Integration | Component | pipeline options'
```
minz1027: Nice!

app/components/pipeline-options/component.js
```diff
@@ -1,13 +1,51 @@
...
+  scmUrl: '',
+  isValid: Ember.computed('scmUrl', {
+    get() {
+      const val = this.get('scmUrl');
+
+      return val.length !== 0 && parse(val).valid;
+    }
+  }),
+  isInvalid: Ember.computed.not('isValid'),
+  isSaving: false,
```
minz1027: Why we need this here? We have another one in the controller and it's passed in to the template.

app/pipeline/options/controller.js
```diff
@@ -12,6 +14,22 @@ export default Ember.Controller.extend({
...
+    },
+    updatePipeline(scmUrl) {
+      let pipeline = this.model;
+
+      pipeline.set('checkoutUrl', scmUrl);
+
+      this.set('isSaving', true);
+
+      pipeline.save()
+        .then(() => this.set('errorMessage', ''))
```
minz1027: Why we need this if the `errorMessage` is empty?

ian-fox: If we hit save and there's an error, it will set `errorMessage` to some string describing the error.

If we hit save again and it works though, we don't want that message to still be displayed. `this` in this context is the controller, not the `pipeline` object.

README.md
tkyi: @ian-fox told me how to do it 😸

app/components/pipeline-options/component.js
tkyi: Good point! 👍 Deleted

PR #185, opened 2017-07-14
fix(tokens): display new token value in a modal

ian-fox:
Following feedback from @petey have moved new token values into a modal!

Also fixed a bug where trying to create a token with an empty description would fail.

screen shot 2017-07-14 at 2 09 52 pm


tkyi:
Aw I liked the little green check mark. 😢 
Also the button is large

PR #186, opened 2017-07-14
fix(tokens): Can create tokens with empty description

ian-fox:
## Context
There was a bug where "refreshed" was not being displayed on the alert when a token had its value refreshed, and one where creating a token with an empty description failed.

There is also a bug where if you refresh a token and then try to modify the same token without refreshing the page the button won't change from "delete" to "update." This seems to be because the `Ember.computed` block changing the button text stops caring about changes to the values after the line that sets it to `Delete` following an update.

This is fixed by modifying the setter for `buttonAction`.

Related to https://github.com/screwdriver-cd/screwdriver/issues/532

### DIFF
--- a/app/components/secret-view/component.js
+++ b/app/components/secret-view/component.js
@@ -2,22 +2,20 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   tagName: 'tr',
-  buttonAction: 'Delete',
+  buttonAction: Ember.computed('newValue', 'secret.allowInPR', 'originalAllowInPR', {
+    get() {
+      return (this.get('newValue')
+        || this.get('originalAllowInPR') !== this.get('secret.allowInPR')) ?
+        'Update' : 'Delete';
+    }
+  }),
   newValue: null,
   originalAllowInPR: null,
   init() {
     this._super(...arguments);
-    this.set('originalAllowInPR', this.get('secret').get('allowInPR'));
+    this.set('originalAllowInPR', this.get('secret.allowInPR'));
   },
   actions: {
-    secretDidChange() {
-      if (this.get('newValue')
-        || this.get('originalAllowInPR') !== this.$('.allow input').prop('checked')) {
-        this.set('buttonAction', 'Update');
-      } else {
-        this.set('buttonAction', 'Delete');
-      }
-    },
     modifySecret() {
       const secret = this.get('secret');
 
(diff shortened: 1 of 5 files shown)


app/token/serializer.js
```diff
@@ -16,6 +16,10 @@ export default DS.RESTSerializer.extend({
 
     const h = Ember.merge(hash, dirty);
 
+    if (h.description === null) {
```
minz1027: If you add it here, if the user wants to update a current token's `description` to empty then it will never happend. I feel you should just add logic to modifyToken to make description empty string instead. Just like what you did for `create`.

ian-fox:
Actually it occurs to me while looking at secrets now that this is the reason secrets use a method called on keyup to change the button text, instead of a computed property with a setter.

I'll defer to @petey on which one seems cleaner.

For those curious: I made an [ember twiddle](https://ember-twiddle.com/?openFiles=controllers.application.js%2C) to demonstrate the difference.

app/components/token-view/component.js
```diff
@@ -2,12 +2,17 @@ import Ember from 'ember';
...
 
-    return (this.get('newName') !== token.get('name')
-         || this.get('newDescription') !== token.get('description')) ?
-      'Update' : 'Delete';
+      return (this.get('newName') !== token.get('name')
+           || this.get('newDescription') !== token.get('description')) ?
+        'Update' : 'Delete';
+    },
+    set(key, value) {
+      return value;
```
petey: This seems really weird since you aren't actually doing anything with the `set` value, or setting any other internal properties. The correct thing to do would probably be to execute `this.setProperties({ newName: null, newDescription: null })` somewhere after those values are no longer required, and let the computed method recalculate as it is supposed to.

ian-fox: The difficulty is that `newName` and `newDescription` are actually the values that are displayed in the textboxes (perhaps I should change the names, or add a comment explaining that), so setting them to null means it looks as though the token has been erased.

I did realize that there was a much easier way though, which is that I should have had `token.name` and `token.description` in the dependent properties of the computed property. That solves the whole problem very nicely.

app/components/secret-view/component.js
```diff
@@ -27,12 +25,9 @@ export default Ember.Component.extend({
         if (this.get('newValue')) {
           secret.set('value', this.get('newValue'));
         }
-        if (this.get('originalAllowInPR') !== this.$('.allow input').prop('checked')) {
-          secret.set('allowInPR', this.$('.allow input').prop('checked'));
-        }
         secret.save();
         this.set('newValue', null);
-        this.set('buttonAction', 'Delete');
+        this.set('originalAllowInPR', secret.get('allowInPR'));
```
ian-fox: This was causing a bug where after you changed the `allowInPR` property and saved it, further changes would not make the button change to "update".

app/components/secret-view/component.js
```diff
@@ -27,12 +25,9 @@ export default Ember.Component.extend({
         if (this.get('newValue')) {
           secret.set('value', this.get('newValue'));
         }
-        if (this.get('originalAllowInPR') !== this.$('.allow input').prop('checked')) {
-          secret.set('allowInPR', this.$('.allow input').prop('checked'));
```
ian-fox: This isn't needed because Ember is already updating that value based on the checkbox.

PR #178, opened 2017-07-07
fix: update to ember 2.14.0

petey:
Context:
========
Ember 2.14 was recently released (7/6)

* Smaller overall file size
* Improved page performance

https://emberjs.com/blog/2017/07/06/ember-2-14-released.html

Objective:
----------
Make sure that everything works well with ember@2.14.* and ember-data@2.14.*

Misc:
-----
Fix errors like the following:

```
Assertion Failed: You modified "model.pipeline.events" twice on <screwdriver-ui@model:pipeline::ember602:300021> in a single render. 
It was rendered in "component:pipeline-events-list" and modified in "component:pipeline-events-list". 
This was unreliable and slow in Ember 1.x and is no longer supported.
```

PR #187, opened 2017-07-17
fix: Fix delete pipeline flow

tkyi:
The delete pipeline flow was broken with the change to the model. (https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/options/route.js#L10)

screen shot 2017-07-17 at 1 59 07 pm

This PR fixes the issue.

Related to: https://github.com/screwdriver-cd/ui/pull/178, https://github.com/screwdriver-cd/ui/pull/181

### DIFF
--- a/app/pipeline/options/controller.js
+++ b/app/pipeline/options/controller.js
@@ -11,7 +11,7 @@ export default Ember.Controller.extend({
       job.save();
     },
     removePipeline() {
-      this.model.destroyRecord().then(() => {
+      this.model.pipeline.destroyRecord().then(() => {
         this.transitionToRoute('home');
       });
     },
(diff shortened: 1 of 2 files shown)


app/pipeline/options/controller.js
```diff
@@ -11,7 +11,9 @@ export default Ember.Controller.extend({
       job.save();
     },
     removePipeline() {
-      this.model.destroyRecord().then(() => {
+      let pipeline = this.model.pipeline;
+
+      pipeline.destroyRecord().then(() => {
```
petey: I would have just put `this.model.pipeline` here instead of declaring a variable that isn't used again.

tkyi: Okay

PR #188, opened 2017-07-20
feat(builds): only show pipeline options when logged in

ian-fox:
## Background
Some users have pointed out that it's hard to tell when they're logged in, and additionally it was mentioned in the security training the other day that it's generally better not to show users options for things they aren't authorized to do.

## Objective
This PR makes it such that if you're logged out you don't get to see the menu that would give you the "options" and "secrets" tabs when viewing a pipeline.

## Screenshots
Logged in:
screen shot 2017-07-20 at 1 51 12 pm

Logged out:
screen shot 2017-07-20 at 1 51 25 pm

### DIFF
--- a/app/pipeline/controller.js
+++ b/app/pipeline/controller.js
@@ -1,0 +1,5 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  session: Ember.inject.service('session')
+});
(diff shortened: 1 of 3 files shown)


minz1027:
Ah, that's just hiding the nav bar from the pipeline page right? what if the user just hit the route directly? `https://cd.screwdriver.cd/pipelines/205/options`

ian-fox:
If they do that (and they aren't logged in) it redirects them to the login page, just like clicking on it now would do.

PR #189, opened 2017-07-20
Bootstrap alerts

ian-fox:
## Context
Ember bootstrap provides a compoent to render alerts, which look nice and have a "close" button.

## Objective:
Update our `info-message` class to use bootstrap alerts.

## Pictures
Before:
screen shot 2017-07-20 at 3 02 58 pm
screen shot 2017-07-20 at 3 06 22 pm

After:

image
image

## Links
Related to https://github.com/screwdriver-cd/screwdriver/issues/615


app/components/pipeline-create-form/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
+{{info-message message=errorMessage type="warning"}}
```
minz1027: The yellow color looks too soft to me as `error`?

ian-fox: I'll defer to @petey on this because I think both look reasonable

tkyi:
I kind of liked the icons. Would it be okay have them in the messages too?

ian-fox:
image
image

Sort of like this?

app/components/pipeline-create-form/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
+{{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
```
joelseq: so does passing in `type="warning"` here override the `type: 'info'` set in `component.js`?

app/components/token-list/template.hbs
```diff
@@ -1,11 +1,14 @@
 <h3>Access Tokens</h3>
-{{info-message message=errorMessage type="error" icon="exclamation-triangle"}}
+{{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
 {{#if newToken}}
-  <div class="new-token">
-    <p>{{fa-icon "check"}} Token {{newToken.action}}. You can only see this value once, so remember to copy it!</p>
-    <span class="new-name">{{newToken.name}}:</span>
-    <span class="new-value">{{newToken.value}}</span>
-  </div>
+  {{#bs-alert type="success" onDismissed=(action "clearNewToken")}}
```
joelseq: Can you not use `info-message` for this alert as well? I feel like if you add a`{{yield}}` in the `info-message` component you could maybe use it here?

app/components/pipeline-create-form/template.hbs
ian-fox: Yes, `info` is just the default.

app/components/info-message/styles.scss
```diff
@@ -1,41 +1,25 @@
...
-  padding: 10px;
-
   i.fa {
-    font-size: 26px;
-  }
-
-  span {
-    display: inline-block;
-    vertical-align: middle;
+    font-size: 18px;
```
jithin1987: Rather than using a hardcoded pixel value you can use responsive font-size values. This can be roughly expressed as `1.1rem`

app/components/token-list/style.scss
```diff
@@ -5,6 +5,11 @@
     margin-top: 10px;
   }
 
+  i.fa {
+    vertical-align: top;
+    font-size: 18px;
```
jithin1987: we should try avoiding hardcoded pixel values for font-sizes, this can be roughly `1.1rem`

ian-fox: Should we fix the rest of the app too? Most of the files are using pixel values right now. It should be pretty simple to do, just `pixel value / 14` right?

jithin1987: `/14` is an oversimplification, it depends on what user has set as default font-size in browser (most cases default is `16px`) If it's a simple enough change go for it, else I am fine with gradually updating when we touch each component later on.

ian-fox: Ahh, it's browser defined? It looks like one of our CSS files is setting it to be 14 when I inspect in chrome.

app/components/token-list/template.hbs
ian-fox: For that I think we'd need to put a `yield` in the component rather than just passing in the message param. If we're doing that, I'm not sure if there's a good way to tell it when to trigger when to show itself again after somebody clicks hide.

@jithin1987 can you think of a good way to do that? I feel like we'd have to have an observer in every context where we're using the component, which kind of removes the point of wrapping the `bs-alert` in a component to handle hiding/showing.

joelseq: @ian-fox yeah in that case I'd say just go with this 👍

jithin1987: @ian-fox can we pass `onDismissed=(action "clearNewToken")` to `info-message` and inside `clearMessage` (https://github.com/screwdriver-cd/ui/blob/e42661eb508cc94f968c9336e1f520b5735718d5/app/components/info-message/component.js#L7)  do `typeof this.get('onDismissed') === 'function' && this.get('onDismissed')()` if it's a function ?

ian-fox: That would work for clearing it, my concern is with how to have it watch to see when to show up again (unless we want to move the `{{if message}}` bit outside of the template.

PR #190, opened 2017-07-21
fix: Show Next job link once current job finishes

jithin1987:
Currently when an ongoing job finishes, it does not display link to the next job which should be part of the same workflow. This PR fixes it by refetching builds tied to the same event.

### DIFF
--- a/tests/integration/components/build-banner/component-test.js
+++ b/tests/integration/components/build-banner/component-test.js
@@ -59,7 +59,7 @@ test('it renders', function (assert) {
   this.set('jobMock', jobMock);
   this.render(hbs`{{build-banner
     buildContainer="node:6"
-    buildStatus="SUCCESS"
+    buildStatus="RUNNING"
     duration="5 seconds"
     event=eventMock
     isAuthenticated=false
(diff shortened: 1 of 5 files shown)


tests/integration/components/build-banner/component-test.js
```diff
@@ -83,7 +83,7 @@ test('it renders', function (assert) {
 });
 
 test('it does not render a restart button for main job when authenticated', function (assert) {
-  assert.expect(2);
+  assert.expect(1);
```
ian-fox: Should we test that it renders the restart button when the build is done?

tkyi: huh? 🤔

jithin1987: build status in this case is `ABORTED` which does not trigger `reloadBuild` and `reloadBuild` is set to `willRender` which contains an `assert`

r3rastogi: Can you add comments for all these numbers (wherever there's an `assert.expect(#)`) please? They should have probably been there before, but it would help when making future code changes

jithin1987: This number directly translates to number of assert's in the test case. I am removing the unnecessary asserts so that the number matches exactly with number of assert's in the test case, without any confusion.

PR #195, opened 2017-08-03
feat(523): Collections-flyout component

joelseq:
## Context

For the user dashboard, the user needs to be able to create / view collections.

## Objective

This PR adds a component that will act as a flyout component that can be added to the routes that require it. Right now, the component is not collapsible but in the future it will be.

Screenshots:

Flyout:
Flyout

Modal:
Create Modal

## References

### DIFF
--- a/app/components/collections-flyout/component.js
+++ b/app/components/collections-flyout/component.js
@@ -1,0 +1,35 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  showModal: false,
+  errorMessage: null,
+  actions: {
+    /**
+     * Action to create a new collection
+     */
+    addNewCollection() {
+      const name = this.get('name');
+      const description = this.get('description');
+
+      return this.get('onCreateCollection')(name, description)
+        .then(() => {
+          this.set('showModal', false);
+        })
+        .catch((error) => {
+          this.set('errorMessage', error.errors[0].detail);
+        });
+    },
+    /**
+     * Action to open / close the create collection modal
+     * @param {boolean} open - whether modal should be open
+     */
+    setModal(open) {
+      if (!open) {
+        this.set('name', null);
+        this.set('description', null);
+        this.set('errorMessage', null);
+      }
+      this.set('showModal', open);
+    }
+  }
+});
--- a/app/components/collections-flyout/styles.scss
+++ b/app/components/collections-flyout/styles.scss
@@ -1,0 +1,31 @@
+& {
+  border-right: 1px solid $sd-text-light;
+  border-bottom: 1px solid $sd-text-light;
+
+  a {
+    cursor: pointer;
+  }
+
+  .flyout {
+    padding: 20px;
+    background-color: $sd-flyout-bg;
+    color: $sd-text;
+  }
+
+  .collection-wrapper {
+    padding: 10px 15px;
+    border-bottom: 1px solid $sd-white;
+    position: relative;
+
+    &:hover {
+      background-color: $sd-flyout-hover;
+      .name {
+        text-decoration: underline;
+      }
+    }
+  }
+
+  .no-collections-text {
+    padding: 15px;
+  }
+}
--- a/app/components/collections-flyout/template.hbs
+++ b/app/components/collections-flyout/template.hbs
@@ -1,0 +1,38 @@
+<div class="flyout row">
+  <h3 class="header-text">
+    <a class="new" {{action "setModal" true}}>{{fa-icon "plus-circle" title="Create a collection" size="md"}}</a> Collections
+  </h3>
+    {{#each collections as |collection|}}
+      <div class="collection-wrapper row">
+        {{#link-to "collection.show" collection.id}}{{collection.name}}{{/link-to}}
+      </div>
+    {{else}}
+      <p class="no-collections-text">No collections to display.</p>
+    {{/each}}
+    {{#if showModal}}
+      {{#bs-modal onHide=(action "setModal" false) as |modal|}}
+        {{#modal.header}}
+          <h4 class="modal-title">Create a new Collection</h4>
+        {{/modal.header}}
+        {{#modal.body}}
+          {{#if errorMessage}}
+            {{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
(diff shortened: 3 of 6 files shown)


r3rastogi:
When you open a pull request for the UI, please include screenshots of what the new changes will render as

app/components/collections-flyout/component.js
```diff
@@ -0,0 +1,39 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  // Boolean whether to display create collection Modal
+  showModal: false,
+  // Error for failing to create a collection
+  error: null,
+  errorMessage: Ember.computed('error', function errorMessage() {
```
ian-fox: I found that `Ember.computed` was slightly too lazy; it was causing problems with being able to have the error message come up again after it was dismissed. 

I'd recommend assigning `errorMessage` directly whenever you were setting `error` before.

https://github.com/screwdriver-cd/ui/pull/193/files

joelseq: Sounds good, thanks!

app/components/collections-flyout/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+  border-right: 1px solid $sd-text-light;
+  border-bottom: 1px solid $sd-text-light;
+
+  a {
+    cursor: pointer;
+  }
+
+  .flyout {
+    padding: 20px;
+    background-color: #f0f4ff;
```
jithin1987: define colors in `app/styles/screwdriver-colors.scss`

app/components/collections-flyout/component.js
```diff
@@ -0,0 +1,35 @@
...
+    addNewCollection() {
+      const name = this.get('name');
+      const description = this.get('description');
+
+      return this.get('onCreateCollection')(name, description)
+        .then(() => {
+          this.set('showModal', false);
+        })
+        .catch((error) => {
+          this.set('errorMessage', error.errors[0].detail);
```
jithin1987: is it a standard way in ember to access errors like `error.errors[0]` ? Shouldn't we check if  `error.errors` is a array of non zero length ?

app/components/collections-flyout/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+    color: #2b2b2b;
+  }
+
+  .collection-wrapper {
+    padding: 10px 15px;
+    border-bottom: 1px solid #ffffff;
+    position: relative;
+
+    &:hover {
+      background-color: #f7faff;
```
jithin1987: move this also to `colors.scss`,

app/components/collections-flyout/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  .flyout {
+    padding: 20px;
+    background-color: $sd-flyout-bg;
+    color: #2b2b2b;
+  }
+
+  .collection-wrapper {
+    padding: 10px 15px;
+    border-bottom: 1px solid #ffffff;
```
tkyi: `$sd-white`

app/components/collections-flyout/styles.scss
tkyi: Can you either use a color from `app/styles/screwdriver-colors.scss` or add this color there?

app/components/collections-flyout/component.js
joelseq: I'm not sure but I believe if there is an error there must at least be 1 element in the errors array. I was referencing Ian's comment and his way of setting it: https://github.com/screwdriver-cd/ui/pull/193/files#diff-a04c46df51fe00937202423e899eec5cR56

petey: @jithin1987 If there is, I haven't yet discovered it. You could probably cast error as an ember object
`Ember.Object(error).get('errors.0.detail')`, but I'm not sure if this would work either. You can dive deeper into the [error objects](https://emberjs.com/api/ember-data/2.14/classes/DS.AdapterError?anchor=#ember436444&show=inherited,protected,private,deprecated&show=inherited,protected,private,deprecated)

app/components/collections-flyout/template.hbs
```diff
@@ -0,0 +1,38 @@
+<div class="flyout row">
+  <h3 class="header-text">
+    <a class="new" {{action "setModal" true}}>{{fa-icon "plus-circle" title="Create a collection" size="md"}}</a> Collections
+  </h3>
+    {{#each collections as |collection|}}
+      <div class="collection-wrapper row">
+        {{#link-to "collection.show" collection.id}}{{collection.name}}{{/link-to}}
```
petey: I'm honestly surprised this doesn't fail due to the `collection.show` route not existing.

app/components/collections-flyout/template.hbs
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    {{else}}
+      <p class="no-collections-text">No collections to display.</p>
+    {{/each}}
+  {{#if showModal}}
```
petey: I'm not really sure why this is not indented and the block above is.

app/components/collections-flyout/template.hbs
joelseq: Yeah if I try to add it to a template for a route then it causes an error but I'm guessing Ember's unit testing accounts for it

app/components/collections-flyout/template.hbs
joelseq: Sorry, I intended to indent it. Fixed!

app/components/collections-flyout/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+  border-bottom: 1px solid $sd-text-light;
+
+  a {
+    cursor: pointer;
+  }
+
+  .flyout {
+    padding: 20px;
+    background-color: $sd-flyout-bg;
+    color: #2b2b2b;
```
ian-fox: Possible to use something from `colors.scss` or else add to that file? Pretty similar to `$sd-text` which is `#26282a`.

joelseq: Good catch, I didn't notice I still had an undefined variable hex in the css. I'll change it right now

PR #197, opened 2017-08-08
refactor(collections): Move store interactions to component

joelseq:
## Context

Currently, all the interactions with the store that the collections-flyout component needs to do are done through the controllers of the routes that require the collections-flyout component. This is not very DRY.

## Objective

This PR moves all the store interactions into the collections-flyout component itself so that it can be a self-sustained component. So now, any route that needs the collections-flyout component can simply just add it to the template like so `{{collections-flyout}}`.

## Reference

### DIFF
--- a/app/components/collections-flyout/component.js
+++ b/app/components/collections-flyout/component.js
@@ -17,19 +17,21 @@ export default Ember.Component.extend({
       const name = this.get('name');
       const description = this.get('description');
 
-      const newCollection = this.get('store').createRecord('collection', {
-        name,
-        description
-      });
-
-      return newCollection.save()
-        .then(() => {
-          this.set('showModal', false);
-        })
-        .catch((error) => {
-          newCollection.destroyRecord();
-          this.set('errorMessage', error.errors[0].detail);
+      Ember.run.schedule('actions', () => {
+        const newCollection = this.get('store').createRecord('collection', {
+          name,
+          description
         });
+
+        return newCollection.save()
+          .then(() => {
+            this.set('showModal', false);
+          })
+          .catch((error) => {
+            newCollection.destroyRecord();
+            this.set('errorMessage', error.errors[0].detail);
+          });
+      });
     },
     /**
      * Action to open / close the create collection modal
--- a/app/components/collections-flyout/component.js
+++ b/app/components/collections-flyout/component.js
@@ -1,8 +1,14 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  showModal: false,
+  collections: Ember.computed('store', {
+    get() {
+      return this.get('store').findAll('collection');
+    }
+  }),
   errorMessage: null,
+  showModal: false,
+  store: Ember.inject.service(),
   actions: {
     /**
      * Action to create a new collection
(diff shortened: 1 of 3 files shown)


tests/integration/components/collections-flyout/component-test.js
```diff
@@ -85,3 +96,58 @@ test('it opens collection create modal', function (assert) {
...
+      assert.deepEqual(data, {
+        name: 'Test',
+        description: 'Test description'
+      });
+
+      return collectionModel;
+    },
+    findAll() {
+      return new Ember.RSVP.Promise((resolve) => {
+        resolve([{
```
jithin1987: can you extract the object in array to a `const mockCollection`, then it can be reused in line 9 as well.

tests/integration/components/collections-flyout/component-test.js
```diff
@@ -85,3 +96,58 @@ test('it opens collection create modal', function (assert) {
...
+  this.render(hbs`{{collections-flyout
+    collections=collections
+    showModal=showModal
+    name=name
+    description=description
+  }}`);
+
+  Ember.run(() => $('.new').click());
+
+  return wait().then(() => {
```
jithin1987: wait() should not be required.

tests/integration/components/collections-flyout/component-test.js
```diff
@@ -85,3 +96,58 @@ test('it opens collection create modal', function (assert) {
...
+    this.set('name', 'Test');
+    this.set('description', 'Test description');
+
+    assert.ok(this.get('showModal'));
+    Ember.run(() => $('.create').click());
+
+    return wait().then(() => {
```
jithin1987: wait() should not be required.

ian-fox:
Should we also be removing the logic from the controllers where it won't be needed any more?

tests/integration/components/collections-flyout/component-test.js
```diff
@@ -85,3 +96,49 @@ test('it opens collection create modal', function (assert) {
...
+  this.inject.service('store');
+
+  this.render(hbs`{{collections-flyout
+    collections=collections
+    showModal=showModal
+    name=name
+    description=description
+  }}`);
+
+  Ember.run(() => $('.new').click());
```
petey: My preference would be to explicitly schedule the use of the run loop in the component, instead of requiring the test to do this work: https://guides.emberjs.com/v2.14.0/applications/run-loop/

app/components/collections-flyout/component.js
```diff
@@ -11,11 +17,17 @@ export default Ember.Component.extend({
       const name = this.get('name');
       const description = this.get('description');
 
-      return this.get('onCreateCollection')(name, description)
+      const newCollection = this.get('store').createRecord('collection', {
```
petey: `Ember.run.schedule('actions', () => {...});`

tests/integration/components/collections-flyout/component-test.js
joelseq: Technically we only need the `Ember.run` for the click on the `create` button. But I'll do the refactor on the component's side

jithin1987: @petey @joelseq As per docs we won't have to use `Ember.run` in regular use cases.
https://www.emberjs.com/api/ember/2.14.1/namespaces/Ember.run

I saw Ember.run being used for clicking buttons in Ember testing documentation https://guides.emberjs.com/v2.14.0/testing/testing-components/

petey: Per the run loop documentation:
> Although autoruns are convenient, they are suboptimal. The current JS frame is allowed to end before the run loop is flushed, which sometimes means the browser will take the opportunity to do other things, like garbage collection. GC running in between data changing and DOM rerendering can cause visual lag and should be minimized.
> Relying on autoruns is not a rigorous or efficient way to use the run loop. Wrapping event handlers manually are preferred.

In this component, clicking specific buttons cause asynchronous activity through the `store` service. By manually scheduling this to happen during a specific phase of the run loop you can avoid blocking on other phases of the run loop.

PR #196, opened 2017-08-04
feat(523): Add collection model and serializer

joelseq:
## Objective

This PR adds a model and serializer for collections for use with `ember-data`.

## References

Blocked by PR: #195 
Issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

PR #198, opened 2017-08-10
feat(collections): Add dashboard routes

joelseq:
## Objective

This PR adds dashboards along with all the routing logic and components needed for it so that users can view their collections.

These are the possible scenarios along with the resulting routes:

1) User is not logged in and visits `/`
2) User is logged in but has no collections and visits `/`
3) User is logged in but has no collections and visits `/dashboards`

    Resulting route: `/`
home route

4) User is not logged in and visits `/dashboards`

    Resulting route: `/login`
login route

5) User is logged in and visits `/`
6) User is logged in and visits `/dashboards`

    Resulting route: `/dashboards/{primaryCollectionId}` where `primaryCollectionId` is the id of the first collection created by the user (for now).
dashboards route

One tradeoff of this and possible source of confusion for the user might be that once they are logged in and have a collection, they can no longer see the getting started page or be on the `/` route as it will always go to their first collection. This means that clicking the logo on the top left will land them on their primary collection's page as well.

## References

Related PRs: #197 #196 #195


tkyi:
`/dashboards`?

app/components/collection-view/styles.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  tbody tr:hover {
+    background-color: $sd-separator;
+    .name {
+      text-decoration: underline;
+    }
+  }
+
+  th {
+    font-size: 14px;
```
jithin1987: Why do we need to define font-size ? If you are trying to reduce size use 0.9em or similar.

app/components/collection-view/template.hbs
```diff
@@ -0,0 +1,28 @@
...
+    <h2>{{collection.name}}</h2>
+    {{#if collection.description}}
+      <p>{{collection.description}}</p>
+    {{/if}}
+  </div>
+  <div class="collection-table row">
+    <table class="col-md-10">
+      <thead>
+        <tr>
+          <th class="appId">Name</th>
```
jithin1987: use `app-id` (kebab-casing) for class names.

app/components/collection-view/template.hbs
```diff
@@ -0,0 +1,28 @@
...
+          <th class="branch">Branch</th>
+        </tr>
+      </thead>
+      <tbody>
+        {{#each collection.pipelines as |pipeline|}}
+          <tr>
+            <td class="appId">
```
jithin1987: use `app-id`

app/dashboard/index/route.js
```diff
@@ -0,0 +1,27 @@
+import Ember from 'ember';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, {
+  activate() {
+    return this.get('store').findAll('collection')
```
jithin1987: this is the same logic in home route, is this required here ? If so any way to avoid duplication ?

app/home/route.js
```diff
@@ -1,4 +1,25 @@
 import Ember from 'ember';
+import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
 
-export default Ember.Route.extend({
+export default Ember.Route.extend(UnauthenticatedRouteMixin, {
+  beforeModel() {
+    if (this.get('session').get('isAuthenticated')) {
```
petey: This extends unauthenticated route, so this should always be false. You can specify a key `routeIfAlreadyAuthenticated` to tell ember where to go if the user is already logged in.

http://ember-simple-auth.com/api/classes/UnauthenticatedRouteMixin.html

joelseq: The reason I am using this here is because I am overriding the beforeModel inherited from `UnauthenticatedRouteMixin`. I cannot use `routeIfAlreadyAuthenticated` because in the `/dashboards` route I am routing someone to the `home` route if they don't have a collection. So if I use `routeIfAlreadyAuthenticated` here to route someone logged in to the `dashboards` route it will cause an infinite loop since it will always route someone that is logged in to the `dashboards` route regardless of if they have a collection or not.

app/dashboard/index/route.js
joelseq: The reason I have this here is just in case someone goes to the `/dashboards` route directly, it will not just show an empty page and instead either direct them to `/login`, `/home`, or `/dashboards/{primaryCollectionId}`. The only way to avoid the duplication that I can think of is `home` always directing a logged in user to `dashboards` and then `dashboards` deciding where to further route the user. But this might be inefficient as in most cases it will be 2 redirections before the user reaches the route that they are supposed to be on.

joelseq:
@tkyi, spoke to @petey about that and we decided on `/dashboards` since it will be the primary route for logged in users with collections

PR #200, opened 2017-08-13
feat(365): Add login buttons for each scm

catto:
## Context

Multiple SCM feature API has been implemented to API(#365).
There is no multiple scm support for UI yet, so we can't test it.

## Objective

I added the support for multiple scms feature to the UI so that we can test the feature on cd.screwdriver.cd.
New login page uses `/v4/auth/contexts` API to show buttons for each scm. It shows the same button as current one if the API returns an error, for backward compatibility.

When the contexts API returns data as below,
```
[
  {
    "context": "github:github.com",
    "displayName": "github"
  },
  {
    "context": "github:mygithubinstance.mycompany.com",
    "displayName": "mygithub_instance"
  }
]
```
UI shows buttons as below.
image

## References

Multiple SCMs feature: screwdriver-cd/screwdriver#365

TODO:
- [x] API change for UI: screwdriver-cd/screwdriver#692

### DIFF
--- a/app/authenticators/screwdriver-api.js
+++ b/app/authenticators/screwdriver-api.js
@@ -3,7 +3,7 @@ import Base from 'ember-simple-auth/authenticators/base';
 // eslint-disable-next-line camelcase
 import { jwt_decode } from 'ember-cli-jwt-decode';
 import ENV from 'screwdriver-ui/config/environment';
-const loginUrl = `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/auth/login/web`;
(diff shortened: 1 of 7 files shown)


app/authenticators/screwdriver-api.js
```diff
@@ -58,10 +58,17 @@ export default Base.extend({
    * @method authenticate
```
tkyi: Can you add JSdocs for `scmContext`

d2lam:
@catto FYI, the API changed was merged & deployed

catto:
Thank you for the JSDoc fix!

PR #201, opened 2017-08-14
feat(collections): Highlight current dashboard

jithin1987:
screen shot 2017-08-15 at 5 03 45 pm


config/environment.js
```diff
@@ -35,6 +35,11 @@ module.exports = (environment) => {
   };
 
   if (environment === 'development') {
+
+    let localhost = process.env.LOCALHOST || 'localhost';
+    ENV.APP.SDAPI_HOSTNAME = `http://${localhost}:9001`;
+    ENV.APP.SDSTORE_HOSTNAME = `http://${localhost}:9002`;
```
joelseq: I don't think you should include this change in this PR. It might be better to have this in a separate PR

app/components/collections-flyout/component.js
```diff
@@ -1,10 +1,25 @@
...
-    get() {
-      return this.get('store').findAll('collection');
-    }
+  init() {
+    this._super(...arguments);
+    this.initialDataFetch = true;
+  },
+  collections: Ember.computed('store', 'selectedCollectionId', function () {
+    let currentCollectionId = this.get('selectedCollectionId');
+    let fetchConf = { reload: this.get('initialDataFetch') };
```
ian-fox: This is probably something you've checked, but is this setting by value or reference?

jithin1987: @ian-fox could you clarify this? I did't get the context.

ian-fox: We're setting the value of `fetchConf.reload` to be `initialDataFetch`, and then modifying the value of `initialDataFetch`. If it's getting/setting by reference, then line 16 would always see the value as `false`, I think.

jithin1987: It won't.  Value of `this.get('initialDataFetch')` is not an object, it's a primitive `boolean`.

However even if value of `this.get('initialDataFetch')` was an object, I find that modifying Object property via `this.set()` is not mutating the data in an object from `this.get()`. Why ? I am not sure, I'm checking Ember docs/source.

config/environment.js
jithin1987: will do. We do need a better way than devs, manually editing `config/environment.js ` and reverting it each time during a git push.

app/components/collections-flyout/component.js
```diff
@@ -1,10 +1,25 @@
...
+
+    this.set('initialDataFetch', false);
+
+    return DS.PromiseArray.create({
+      promise: this.get('store').findAll('collection', fetchConf)
+        .then(collections => collections.map(collection => ({
+          id: collection.get('id'),
+          name: collection.get('name'),
+          current: collection.id === currentCollectionId ? 'row--active' : ''
```
joelseq: Would it be better to just set a boolean here for whether or not it is the current collection and setting the class in the template instead?

app/components/collections-flyout/template.hbs
```diff
@@ -3,7 +3,7 @@
     <a class="new" {{action "setModal" true}}>{{fa-icon "plus-circle" title="Create a collection" size="md"}}</a> Collections
   </h3>
     {{#each collections as |collection|}}
-      <div class="collection-wrapper row">
+      <div class="collection-wrapper row {{collection.current}}">
```
joelseq: Related to my comment above, if you set a boolean of whether or not the collection is the current one, you can just add an if helper here to conditionally add the class. You can look [here](https://github.com/screwdriver-cd/ui/blob/master/app/components/build-step-view/template.hbs#L6) for an example

app/components/collections-flyout/component.js
ian-fox: Ahh, of course. 👍

r3rastogi:
Could you include a screenshot of what the highlighting looks like, please?

app/components/search-list/component.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
+  session: Ember.inject.service(),
```
joelseq: Since you're already injecting the session service in the `search` controller would it be better to just pass in `isAuthenticated` as an attribute rather than injecting the session inside this component?

tests/integration/components/collections-flyout/component-test.js
```diff
@@ -142,3 +150,26 @@ test('it creates a collection', function (assert) {
 
   assert.notOk(this.get('showModal'));
 });
+
+test('it renders an active a collection', function (assert) {
```
joelseq: I think you might have a typo here. Should it be 'it renders an active collection'?

app/components/collections-flyout/template.hbs
```diff
@@ -1,9 +1,13 @@
 <div class="flyout row">
   <h3 class="header-text">
+    {{#if session.isAuthenticated}}
     <a class="new" {{action "setModal" true}}>{{fa-icon "plus-circle" title="Create a collection" size="md"}}</a> Collections
+    {{else}}
+    Collections
```
ian-fox: Do we want to show the flyout at all if they aren't logged in?

jithin1987: We still show the flyout with a single item which is the current Dashboard under navigation.

app/components/search-list/component.js
jithin1987: I don't have a strong preference on this. Session is a service, so it's purpose is to live throughout the lifecycle of the app, we still are not creating new local state, but dependent on global state of the service.

app/components/collections-flyout/template.hbs
ian-fox: The one that says "Introducing Screwdriver?"

jithin1987: This will be the logged out view 

screen shot 2017-08-16 at 2 23 22 pm

PR #199, opened 2017-08-11
feat(collections): Add pipelines to a collection from search route

joelseq:
## Context

For the dashboards feature, users need to be able to add pipelines to their collections

## Objective

This PR adds functionality to allow users to add pipelines to their collections from the search results page.

**Screenshots**

Search page:
search page

Dropdown:
dropdown

Successful add:
image

## References

Issue: [screwdriver-cd/screwdriver#523](https://github.com/screwdriver-cd/screwdriver/issues/523)

PR #202, opened 2017-08-15
test(collections): Add more tests for collections / dashboards

joelseq:
Forgot to add some acceptance tests for dashboards previously so adding it in this PR along with some more integration tests for collections-flyout component.

## References

Related PRs: #195, #197, #198, #199

### DIFF
--- a/app/components/collections-flyout/template.hbs
+++ b/app/components/collections-flyout/template.hbs
@@ -22,7 +22,7 @@
           {{#if errorMessage}}
             {{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
           {{/if}}
-          {{#bs-form onSubmit=(action "addNewCollection") as |form|}}
+          {{#bs-form class="collection-form" onSubmit=(action "addNewCollection") as |form|}}
             {{#form.group class="name"}}
               <label class="control-label">Name</label>
               <input value={{name}} class="form-control" oninput={{action (mut name) value="target.value"}} type="text">
@@ -32,8 +32,8 @@
               <input value={{description}} class="form-control" oninput={{action (mut description) value="target.value"}} type="text">
             {{/form.group}}
             {{#form.group class="footer"}}
-            {{#bs-button class="cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
-            {{#bs-button buttonType="submit" class="create" type="success"}}Create{{/bs-button}}
+            {{#bs-button class="collection-form__cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
+            {{#bs-button buttonType="submit" class="collection-form__create" type="success"}}Create{{/bs-button}}
             {{/form.group}}
           {{/bs-form}}
         {{/modal.body}}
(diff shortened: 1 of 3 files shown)


app/components/collections-flyout/template.hbs
```diff
@@ -28,8 +28,8 @@
               <input value={{description}} class="form-control" oninput={{action (mut description) value="target.value"}} type="text">
             {{/form.group}}
             {{#form.group class="footer"}}
-            {{#bs-button class="cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
-            {{#bs-button buttonType="submit" class="create" type="success"}}Create{{/bs-button}}
+            {{#bs-button class="collection-cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
+            {{#bs-button buttonType="submit" class="collection-create" type="success"}}Create{{/bs-button}}
```
joelseq: I had to change the class name here from `create` to `collection-create` because the create pipeline button also has a class `create` so it was causing issues with my acceptance tests.

app/components/collections-flyout/template.hbs
```diff
@@ -28,8 +28,8 @@
               <input value={{description}} class="form-control" oninput={{action (mut description) value="target.value"}} type="text">
             {{/form.group}}
             {{#form.group class="footer"}}
-            {{#bs-button class="cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
-            {{#bs-button buttonType="submit" class="create" type="success"}}Create{{/bs-button}}
+            {{#bs-button class="collection-cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
```
jithin1987: I would rather have `collection-form` added to the `bs-form` on top and call these `collection-form__create` & `collection-form__cancel` This is the BEM class name notation http://getbem.com/naming/

joelseq: Yeah I could do that. I just figured since we weren't using the BEM naming convention anywhere else that it wouldn't be needed in this case either.

jithin1987: That's something we can gradually introduce. It won't be inconsistent with existing class names.

app/components/collections-flyout/template.hbs
```diff
@@ -28,8 +28,8 @@
               <input value={{description}} class="form-control" oninput={{action (mut description) value="target.value"}} type="text">
             {{/form.group}}
             {{#form.group class="footer"}}
-            {{#bs-button class="cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
-            {{#bs-button buttonType="submit" class="create" type="success"}}Create{{/bs-button}}
+            {{#bs-button class="collection-form__cancel" onClick=(action modal.close) type="danger"}}Cancel{{/bs-button}}
```
r3rastogi: why are we using a double-underscore (`_` vs `__`)

joelseq: This is the BEM naming convention, I was following along with @jithin1987's [review comment](https://github.com/screwdriver-cd/ui/pull/202#discussion_r133350562)

jithin1987: This is BEM notation https://css-tricks.com/bem-101/

r3rastogi: mkay thanks

PR #203, opened 2017-08-17
feat(collections): Add remove pipeline functionality

joelseq:
## Context

Users need to be able to remove pipelines from a collection (if they own it).

## Objective

This PR adds functionality to allow users to remove pipelines that they don't want to have in a collection

Example:

Screen

## Reference

### DIFF
--- a/app/components/collection-view/component.js
+++ b/app/components/collection-view/component.js
@@ -1,4 +1,24 @@
 import Ember from 'ember';
 
 export default Ember.Component.extend({
+  session: Ember.inject.service(),
+  removePipelineError: null,
+  actions: {
+    /**
+     * Action to remove a pipeline from a collection
+     *
+     * @param {Number} pipelineId - id of pipeline to remove
+     * @param {Number} collectionId - id of collection to remove from
+     * @returns {Promise}
+     */
+    pipelineRemove(pipelineId, collectionId) {
+      return this.get('onPipelineRemove')(+pipelineId, collectionId)
+        .then(() => {
+          this.set('removePipelineError', null);
+        })
+        .catch((error) => {
+          this.set('removePipelineError', error.errors[0].detail);
+        });
+    }
+  }
 });
--- a/app/components/collection-view/template.hbs
+++ b/app/components/collection-view/template.hbs
@@ -1,4 +1,7 @@
 <div>
+  {{#if removePipelineError}}
+    {{info-message message=removePipelineError type="warning" icon="exclamation-triangle"}}
+  {{/if}}
   <div class="header">
     <h2>{{collection.name}}</h2>
     {{#if collection.description}}
@@ -11,15 +14,25 @@
         <tr>
           <th class="app-id">Name</th>
           <th class="branch">Branch</th>
+          {{#if session.isAuthenticated}}
+            <th></th>
+          {{/if}}
         </tr>
       </thead>
       <tbody>
         {{#each collection.pipelines as |pipeline|}}
-          <tr>
+          <tr class="collection-pipeline">
             <td class="app-id">
               {{#highlight-terms query}}{{#link-to "pipeline" pipeline.id}}{{pipeline.scmRepo.name}}{{/link-to}}{{/highlight-terms}}
             </td>
             <td class="branch">{{fa-icon "fa-code-fork"}}{{pipeline.scmRepo.branch}}</td>
+            {{#if session.isAuthenticated}}
+              <td onclick={{action "pipelineRemove" pipeline.id collection.id}} class="collection-pipeline__remove">
+                <span>
+                  &times;
+                </span>
+              </td>
+            {{/if}}
           </tr>
         {{/each}}
       </tbody>
--- a/app/dashboard/show/controller.js
+++ b/app/dashboard/show/controller.js
@@ -1,0 +1,16 @@
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  actions: {
+    removePipeline(pipelineId, collectionId) {
+      return this.store.findRecord('collection', collectionId)
+        .then((collection) => {
+          const pipelineIds = collection.get('pipelineIds') || [];
+
+          collection.set('pipelineIds', pipelineIds.filter(id => id !== pipelineId));
+
+          return collection.save();
+        });
+    }
+  }
+});
--- a/app/components/collection-view/styles.scss
+++ b/app/components/collection-view/styles.scss
@@ -42,6 +42,17 @@
   .collection-table {
     padding: 25px;
   }
+
+  .collection-pipeline__remove {
+    opacity: 0;
+    cursor: pointer;
+    font-size: 1.5em;
+  }
+
+  // When collection-pipeline is hovered on, show remove button
+  .collection-pipeline:hover .collection-pipeline__remove {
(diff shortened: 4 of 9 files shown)


app/components/collection-view/component.js
```diff
@@ -1,4 +1,24 @@
...
+     * @param {Number} collectionId - id of collection to remove from
+     * @returns {Promise}
+     */
+    pipelineRemove(pipelineId, collectionId) {
+      return this.get('onPipelineRemove')(+pipelineId, collectionId)
+        .then(() => {
+          this.set('removePipelineError', null);
+        })
+        .catch(() => {
+          this.set('removePipelineError', 'User does not have permission');
```
jithin1987: is this the only reason for error ? We need to differentiate between permission vs other API error

app/components/collection-view/template.hbs
```diff
@@ -11,15 +14,25 @@
...
       </thead>
       <tbody>
         {{#each collection.pipelines as |pipeline|}}
-          <tr>
+          <tr class="collection-pipeline">
             <td class="app-id">
               {{#highlight-terms query}}{{#link-to "pipeline" pipeline.id}}{{pipeline.scmRepo.name}}{{/link-to}}{{/highlight-terms}}
             </td>
             <td class="branch">{{fa-icon "fa-code-fork"}}{{pipeline.scmRepo.branch}}</td>
+            {{#if session.isAuthenticated}}
```
jithin1987: weren't we planning to show this only if logged in user is the owner of collection?

app/components/collection-view/component.js
joelseq: I'm not sure but I had added this because I think this is the only known cause for it to error. If something goes wrong on the API side of things I'm not sure if this action will even make it as it is dependent on the store having the pipeline and collection listed above.

joelseq: I'll change it to handle cases for API errors too

app/components/collection-view/component.js
```diff
@@ -1,4 +1,28 @@
...
+     * @returns {Promise}
+     */
+    pipelineRemove(pipelineId, collectionId) {
+      return this.get('onPipelineRemove')(+pipelineId, collectionId)
+        .then(() => {
+          this.set('removePipelineError', null);
+        })
+        .catch((error) => {
+          if (typeof error === 'string') {
+            this.set('removePipelineError', 'User does not have permission');
```
jithin1987: don't hardcode it here, you are already throwing doing `Ember.RSVP.reject` from parent controller, read that error message here.

joelseq: Oops sorry I just recently changed it after your comment above, forgot to change that to use the string instead. Will change it right now 👍

app/components/collection-view/template.hbs
joelseq: Yes ideally that's what we would do. But I think for now since everything else has a similar interaction wherein everything is visible to logged in users this check is sufficient.

app/components/collection-view/component.js
```diff
@@ -1,4 +1,28 @@
...
+     * @param {Number} collectionId - id of collection to remove from
+     * @returns {Promise}
+     */
+    pipelineRemove(pipelineId, collectionId) {
+      return this.get('onPipelineRemove')(+pipelineId, collectionId)
+        .then(() => {
+          this.set('removePipelineError', null);
+        })
+        .catch((error) => {
+          if (typeof error === 'string') {
```
jithin1987: can you double check if this line is getting called ? I would be surprised if promise rejection from `onPipelineRemove` gets converted to string here.

joelseq: I have a [test case](https://github.com/screwdriver-cd/ui/pull/203/files#diff-dc139d5d9170f745c607201ae8075798R120) that checks for that

jithin1987: well Ember continues to surprise me.

app/components/collection-view/template.hbs
jithin1987: Do we have anything so far in the UI which does not belong to the logged in user ?

app/components/collection-view/component.js
jithin1987: ooh actually it will since promise rejection value is a string, I was looking to see of we can reject this in a way which is consistent with standard error format of  `error.errors[0].detail`

app/dashboard/show/controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default Ember.Controller.extend({
+  actions: {
+    removePipeline(pipelineId, collectionId) {
+      return this.store.findRecord('collection', collectionId)
+        .then((collection) => {
+          const pipelineIds = collection.get('pipelineIds');
+
+          if (!pipelineIds) {
+            return Ember.RSVP.reject('User does not have permission');
```
jithin1987: how does this translate to user permission error ?

tkyi: I was confused about that too

app/components/collection-view/template.hbs
joelseq: I believe right now in the UI anyone can view secrets of a pipeline even if they don't own the pipeline but if they try to add a secret or delete one that exists then it will just show them an unauthorized error.

app/components/collection-view/component.js
minz1027: Don't like we have 2 types error format either

app/dashboard/show/controller.js
joelseq: So the way that the Ember store works is that initially it will do a `GET` to `/collections` to get a list of all the user's collections. This will populate the store with an array of the user's collections in this format:

```js
{
  name: "Some name",
  description: "Some description",
  pipelineIds: []
}
```

Then when the user tries to go to a dashboard to see a collection, this will send a `GET` request to `/collections/1` for example and this request will send a response in the format

```js
{
  name: "Some name",
  description: "Some description",
  pipelines: [
       // Populated pipeline objects
  ]
}
```

So if the user does not own a collection and the store has that collection, then it will not have the `pipelineIds` field populated which is currently the only way to check if someone owns a collection or not since we don't have the `id` of the user. The check here is to make sure that it doesn't go on to the `pipelines.filter` line which causes an error because `pipelines` is undefined.

One way to work around this would be if the `GET` request for a particular collection also includes the `pipelineIds` field in the response, in which case it would exist in the store, and then the API would be the one that throws the error instead of me doing a check here. Another way would be if I set `pipelineIds` to an empty array if it doesn't already exist and then try to perform the `save` operation and that would cause the API to throw an error as well. What do you all think? @minz1027 @jithin1987 @tkyi

app/components/collection-view/component.js
joelseq: This would be possible I think. My thought process to making it a string is to be able to distinguish between when the error is coming from the API vs the error that I'm throwing from my validation in the controller for dashboard/show.

app/dashboard/show/controller.js
minz1027: Joel and I discussed this today. I feel having the extra field `pipelineIds` in `GET collections/id` is more clean than what it is now. 

And if we have this, Joel doesn't need to throw the error as a string here. The UI can use the error message from API directly.

PR #204, opened 2017-08-18
feat(collections): Delete collections from flyout

joelseq:
## Objective

Add functionality to delete collections from flyout

Example:

Delete

## Reference

### DIFF
--- a/app/components/collection-view/template.hbs
+++ b/app/components/collection-view/template.hbs
@@ -3,9 +3,9 @@
     {{info-message message=removePipelineError type="warning" icon="exclamation-triangle"}}
   {{/if}}
   <div class="header">
-    <h2>{{collection.name}}</h2>
+    <h2 class="header__name">{{collection.name}}</h2>
     {{#if collection.description}}
-      <p>{{collection.description}}</p>
+      <p class="header__description">{{collection.description}}</p>
     {{/if}}
   </div>
   <div class="collection-table row">
--- a/app/components/collections-flyout/component.js
+++ b/app/components/collections-flyout/component.js
@@ -8,18 +8,25 @@ const formatCollection = (collection, currentCollectionId) => ({
 });
 
 export default Ember.Component.extend({
-  collections: Ember.computed('storeCollections.[]', 'selectedCollectionId', function () {
-    let currentCollectionId = this.get('selectedCollectionId');
+  collections: Ember.computed('storeCollections.[]', 'selectedCollectionId', {
+    get() {
+      let currentCollectionId = this.get('selectedCollectionId');
 
-    return DS.PromiseArray.create({
-      promise: this.get('storeCollections').then(collections =>
-        collections.map(collection => formatCollection(collection, currentCollectionId)))
-    });
+      return DS.PromiseArray.create({
+        promise: this.get('storeCollections').then(collections =>
+          collections.map(collection => formatCollection(collection, currentCollectionId)))
+      });
+    }
   }),
-  storeCollections: Ember.computed('store', function () {
-    return this.get('store').findAll('collection');
+  storeCollections: Ember.computed('store', {
+    get() {
+      return this.get('store').findAll('collection');
+    }
   }),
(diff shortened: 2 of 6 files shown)


app/components/collections-flyout/component.js
```diff
@@ -8,18 +8,25 @@ const formatCollection = (collection, currentCollectionId) => ({
 });
 
 export default Ember.Component.extend({
-  collections: Ember.computed('storeCollections.[]', 'selectedCollectionId', function () {
-    let currentCollectionId = this.get('selectedCollectionId');
+  collections: Ember.computed('storeCollections.[]', 'selectedCollectionId', {
+    get() {
```
r3rastogi: Why did we add this get()? What does that do?

app/components/collections-flyout/template.hbs
```diff
@@ -1,19 +1,49 @@
 <div class="flyout row">
-  <h3 class="header-text">
+  <div class="header row">
     {{#if session.isAuthenticated}}
-    <a class="new" {{action "setModal" true}}>{{fa-icon "plus-circle" title="Create a collection" size="md"}}</a> Collections
+    <h2 class="header__text col-xs-9">
```
tkyi: Should stuff after an `if` be indented?

tests/integration/components/collections-flyout/component-test.js
```diff
@@ -240,3 +240,73 @@ test('it cancels creation of a collection', function (assert) {
...
+
+test('it deletes a collection', function (assert) {
+  assert.expect(7);
+
+  injectSessionStub(this);
+
+  const $ = this.$;
+  const collectionModelMock = {
+    destroyRecord() {
+      // Dummy assert to make sure this function gets called
```
r3rastogi: We can't do an `assert.isCalled()` here?

app/components/collection-view/template.hbs
```diff
@@ -1,8 +1,8 @@
 <div>
   <div class="header">
-    <h2>{{collection.name}}</h2>
+    <h2 class="header__name">{{collection.name}}</h2>
```
tkyi: Do we need to add a check to make sure `{{#if collection}}` before these blocks, or is that covered somewhere else?

joelseq: If the collection doesn't exist it doesn't reach this part anyway

app/components/collections-flyout/component.js
joelseq: This is the format we use for all the other computed properties [example](https://github.com/screwdriver-cd/ui/blob/master/app/components/build-step-view/component.js) and there were linter warnings because of the unnamed function anyway

app/components/collections-flyout/template.hbs
joelseq: I personally prefer to ident it after an `if` but it was previously not indented so I left it that way and didn't indent my own things nested inside an `if` in this file for consistency. Do you think I should just go ahead and indent it?

tests/integration/components/collections-flyout/component-test.js
joelseq: I wasn't sure if there was an `isCalled()` on the assert as I couldn't figure out all the functions available for asserts. I'll try it out and if it works I'll make the change 👍

app/components/collections-flyout/component.js
r3rastogi: Ahhh ok. Thanks, that makes sense

tests/integration/components/collections-flyout/component-test.js
joelseq: Just tried and there isn't an `isCalled`. I checked the [docs](https://api.qunitjs.com/assert/equal) and I cannot seem to find a better assertion for this use case

app/components/collections-flyout/template.hbs
tkyi: I'm not really sure what the convention for our team is

PR #205, opened 2017-08-19
fix(collections): After delete collection navigate to home page if on…

jithin1987:
… collection detail page

Fix console errors by usine peekRecord instead of findRecord

----

Test cases coming

### DIFF
--- a/app/components/collections-flyout/component.js
+++ b/app/components/collections-flyout/component.js
@@ -65,13 +65,17 @@ export default Ember.Component.extend({
      * @param {collection} collection - the collection to delete
      */
     deleteCollection(collection) {
-      return this.get('store').findRecord('collection', collection.id)
-        .then(c =>
-          c.destroyRecord()
-            .then(() => {
-              this.set('collectionToDelete', null);
-            })
-        );
+      const c = this.get('store')
+        .peekRecord('collection', collection.id);
+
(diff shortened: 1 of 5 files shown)


tests/unit/dashboard/show/controller-test.js
```diff
@@ -61,3 +62,13 @@ test('it calls removePipeline', function (assert) {
   // Remove pipeline with id 3 from collection with id 1
   controller.send('removePipeline', 3, 1);
 });
+
+sinonTest('it calls onDeleteCollection', function (assert) {
```
joelseq: This test is really cool! I didn't know about ember-qunit-sinon before 👍

app/components/collections-flyout/component.js
```diff
@@ -65,13 +65,17 @@ export default Ember.Component.extend({
...
      */
     deleteCollection(collection) {
-      return this.get('store').findRecord('collection', collection.id)
-        .then(c =>
-          c.destroyRecord()
-            .then(() => {
-              this.set('collectionToDelete', null);
-            })
-        );
+      let c = this.get('store')
```
tkyi: Should it be `const c`?

jithin1987: @tkyi updated

PR #208, opened 2017-08-25
fix: navigating to newly created collection should work

jithin1987:
## Context

Currently if you create a new collection and then try to navigate to it, navigation won't work. This PR fixes it.

@joelseq @r3rastogi

### DIFF
--- a/tests/integration/components/collections-flyout/component-test.js
+++ b/tests/integration/components/collections-flyout/component-test.js
@@ -156,22 +156,16 @@ test('it renders an active collection', function (assert) {
   assert.expect(4);
   const $ = this.$;
 
-  const storeStub = Ember.Object.extend({
-    findAll() {
-      return new Ember.RSVP.Promise(resolve => resolve([mockCollection]));
-    }
-  });
-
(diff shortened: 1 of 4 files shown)


tests/integration/components/collections-flyout/component-test.js
```diff
@@ -156,22 +156,16 @@ test('it renders an active collection', function (assert) {
...
   this.set('selectedCollectionId', 1);
 
-  this.render(hbs`{{collections-flyout selectedCollectionId=selectedCollectionId}}`);
+  this.render(hbs`{{collections-flyout collections=collections
+    selectedCollectionId=selectedCollectionId}}`);
 
   assert.equal($('.header__text').text().trim(), 'Collections');
   assert.notOk($('.header__text a i').length);
-  assert.equal($($('.collection-wrapper a').get(0)).text().trim(), 'name');
+  assert.equal($($('.collection-wrapper a').get(0)).text().trim(), 'Test');
```
joelseq: How come this worked previously with 'name' and now it's changed to 'Test'?

PR #213, opened 2017-09-07
feat(522): Template discovery views

petey:
Initial stab at https://github.com/screwdriver-cd/screwdriver/issues/522

Still need to add a link in the header to the templates page, and cleanup CSS a bit.

Lists all available templates (de-duped by name).
screen shot 2017-09-06 at 5 58 34 pm

Detail view displays info about a given template and all the versions.
screen shot 2017-09-06 at 5 58 44 pm


app/components/template-header/template.hbs
```diff
@@ -0,0 +1,12 @@
+<h1>{{template.name}} ({{template.version}})</h1>
+<p>{{template.description}}</p>
+<ul>
+  {{! TODO: Link to pipeline and/or source code}}
```
r3rastogi: Ohh yes, linking to source code would be good!

app/components/template-header/template.hbs
```diff
@@ -0,0 +1,12 @@
...
+  <li>Released by: <a href="mailto:{{template.maintainer}}">{{template.maintainer}}</a></li>
```
r3rastogi: Should we change this to "Maintained by: ", so that we can differentiate between who created the template vs who owns it (in the case that someone creates one but then hands off ownership to someone else)?

petey: I was going by the available mocks. We do not currently have fields that differentiate "creator" and "owner". We could change this when data becomes available.

r3rastogi: Cool, makes sense

tests/integration/components/template-header/component-test.js
```diff
@@ -0,0 +1,34 @@
...
+  const $ = this.$;
+
+  this.set('mock', TEMPLATE);
+  this.render(hbs`{{template-header template=mock}}`);
+
+  assert.equal($('h1').text().trim(), 'foo/bar (2.0.0)');
+  assert.equal($('p').text().trim(), 'A test example');
+  assert.equal($('ul li:first-child').text().trim(), 'Released by: bruce@wayne.com');
+  assert.equal($('ul li:first-child a').attr('href'), 'mailto:bruce@wayne.com');
+  assert.equal($('ul li:last-child').text().trim(), 'Tags: cararmored');
```
r3rastogi: I might be misunderstanding something here, but shouldn't this be two separate tags?

tests/unit/templates/route-test.js
```diff
@@ -0,0 +1,12 @@
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('route:templates', 'Unit | Route | templates', {
+  // Specify the other units that are required for this test.
+  // needs: ['controller:foo']
+});
+
+test('it exists', function (assert) {
```
r3rastogi: This is still just the default test, right? Are we going to add others?

tkyi:
Looks good for a first stab. There are some additional features (that I've added to the issue description) that I'd like to see, such as additional metadata like "last updated" or "number of downloads" or the template README.

I'm also wondering if the breadcrumb at the top of the template detail page makes sense. We don't use the breadcrumbs anywhere else in our product. Moreover, would there ever be an case where a simple "back to template Search page" button or something more simple wouldn't suffice? Just not sure when we would ever have more levels deeper than one (eg `Templates > template_name`).

tests/unit/templates/route-test.js
petey: I don't really know what to add to this test. The route does absolutely nothing.

petey:
@tkyi I agree about the breadcrumb. It's pretty wonky.

We currently don't return any data in the api calls that look like timestamps or number of downloads. We can add those features after the api changes.

tests/integration/components/template-header/component-test.js
petey: I'm testing the `text()` of the parent node, so it doesn't display the HTML tags, only the actual text in that element. So it *is* two tags, it just doesn't look that way when looking at the text.

app/components/template-header/template.hbs
tkyi: Maybe even adding a badge for the pipeline would be cool 👍

app/templates/index/route.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default Ember.Route.extend({
+  template: Ember.inject.service(),
+  model() {
+    return this.get('template').getAllTemplates().then((templates) => {
+      const result = [];
+      const names = {};
+
+      // Reduce versions down to one entry
+      templates.forEach((t) => {
```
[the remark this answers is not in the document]
petey: The api returns multiple versions of the same template.

app/components/app-header/template.hbs
```diff
@@ -22,10 +22,29 @@
         {{/link-to}}
       {{/nav.item}}
       {{#nav.item}}
-        <a href="http://docs.screwdriver.cd" class="icon docs">
-          {{inline-svg "docs" class="img"}}
+        {{#nav.dropdown as |dd|}}
+          {{#dd.toggle}}{{inline-svg "docs" class="img"}}
           <span class='icontitle'>Documentation</span>
-        </a>
+          <span class="caret"></span>
```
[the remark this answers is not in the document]
petey: Ideally, yes, but this is what ember-bootstrap is doing. The caret is styled by that.
http://www.ember-bootstrap.com/api/classes/Components.Dropdown.html

app/templates/index/route.js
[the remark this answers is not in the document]
jithin1987: Can we pick the latest version here instead of 1st one with same name ? Or is it absolutely safe to assume that API is sorting results based on version ?

app/components/app-header/template.hbs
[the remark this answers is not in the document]
jithin1987: sounds good 👍

app/templates/index/route.js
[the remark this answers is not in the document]
petey: I think the list of all templates comes sorted by their numeric id, so most recent first.

PR #215, opened 2017-09-11
fix: use targets and real browsers

petey:
# Context:
phantomjs is no longer maintained, and has not had an update since 1/2016. phantomjs does not support any ES2105 syntax, and tests will break when the transpiled code is created for current browsers. Newer browser releases now support a headless mode, which means we can test with the actual browsers we support.

## Objective
* Utilize [ember-cli targets](http://rwjblue.com/2017/04/21/ember-cli-targets/) to produce sane transpiled code for evergreen browsers. 
* Test with `chrome` in CI mode [`ember test`] instead of `phantomjs`
* Test with `chrome` and `firefox` in dev mode [`ember test --server`]

## Caveats
* The `node:6` container does not support user namespaces, so Chrome must run with the `--no-sandbox` flag. This is normally not recommended.
* Headless Chrome was introduced in Chrome 59. This should install at least Chrome 61.
* Headless Firefox works on Firefox 55 on linux, and Firefox 56 on macOS. Firefox 56 is still the beta version and can be fetched from [here](https://www.mozilla.org/en-US/firefox/channel/desktop/#beta). 
* When testing in dev mode [`ember test --server`], in order to get a "headed" browser, you must visit a link in a browser that is provided by the test framework on startup. A "headed" browser will not be started automatically anymore.

### DIFF
--- a/config/targets.js
+++ b/config/targets.js
@@ -2,8 +2,8 @@
 
 module.exports = {
   browsers: [
-    // 'last 1 Chrome versions'
-    // 'last 1 Firefox versions'
-    // 'last 1 Safari versions'
+    'last 1 Chrome versions',
+    'last 1 Firefox versions',
+    'last 1 Safari versions'
   ]
 };
--- a/bin/setup_chrome.sh
+++ b/bin/setup_chrome.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
+sh -c 'echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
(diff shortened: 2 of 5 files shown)


config/targets.js
```diff
@@ -2,8 +2,8 @@
 
 module.exports = {
   browsers: [
-    // 'last 1 Chrome versions'
-    // 'last 1 Firefox versions'
-    // 'last 1 Safari versions'
+    'last 1 Chrome versions',
+    'last 1 Firefox versions',
+    'last 1 Safari versions'
```
minz1027: We are not specifying safari in `testem.js`. If we only have it here will it actually do anything?

petey: We could add it to the dev mode tests, but not ci mode. Safari doesn't have native headless mode, yet, and can't be installed in a Debian container in any case.

PR #223, opened 2017-10-14
fix: update to ember2.16.2

petey:
Context:
========
Continue to stay up to date with the latest version of emberjs

Objective:
----------
Upgrade to ember core code to 2.16.2. Ember 2.16 brings in use of the JavaScript Modules API. This requires a change to pretty much every file to import from a specific module as opposed to the Ember global object.
Brought in eslint-plugin-ember to begin checking for ember-specific issues. Will turn on more checks in later PR(s). There are a lot of lint errors that will be introduced by this plugin.

[Ember 2.16 Anouncement](https://emberjs.com/blog/2017/10/11/ember-2-16-released.html)
[eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember)

There are only 2 more minor releases planned for Ember.js before they release Ember 3.x. For major releases, Ember typically only removes deprecated code from the framework. This tends to result in a very smooth transition between major versions. By staying current, we reduce the chances that we are tied to deprecated code.
[Ember 3.x](https://emberjs.com/blog/2017/10/03/the-road-to-ember-3-0.html)


app/event/model.js
```diff
@@ -1,31 +1,32 @@
...
     }
   }),
-  durationText: Ember.computed('duration', {
+  durationText: computed('duration', {
     get() {
       return humanizeDuration(this.get('duration'), { round: true, largest: 1 });
     }
   }),
-  isComplete: Ember.computed('builds.[]', 'builds.lastObject.status', {
+  isComplete: computed('builds.{[],lastObject.status}', {
```
tkyi: 😮

petey: Yeah, Ember is crazy!

petey: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/use-brace-expansion.md

PR #225, opened 2017-10-19
feat(723): Create event to start builds

tkyi:
## Context

With our new workflow revamp, all new builds should be created using the new POST event endpoint. 

## Objective

This PR calls the new create event endpoint when starting builds. 
- When a user creates a new build from the main pipeline, they will be redirected to the same pipeline page instead of the new build page.
- When a user restarts a PR build, they will still be redirected to the PR build page.

_Note: This PR does not add new functionality, it only duplicates the current functionality using a new endpoint._

## Relevant links

Related to https://github.com/screwdriver-cd/screwdriver/issues/723
Blocked by https://github.com/screwdriver-cd/screwdriver/pull/762


app/pipeline/builds/index/controller.js
```diff
@@ -8,27 +8,18 @@ export default Controller.extend({
...
+        pipelineId,
+        startFrom: '~commit'
+      });
 
-      return newBuild.save().then((b) => {
+      return newEvent.save().then((e) => {
         this.set('isShowingModal', false);
 
-        return this.transitionToRoute('pipeline.builds.build', b.get('id'));
+        return this.transitionToRoute('pipeline', e.get('pipelineId'));
```
minz1027: Can just use `newEvent`. Just to be consistent

PR #226, opened 2017-10-23
Workflow graphed with vis

petey:
Context:
========
With the updates to the workflow backend, we need a better way to display a workflow than a linear set of builds

Objective:
-----------
Display a directed graph of the workflow for pipelines and events.

Misc:
------
This PR is meant as a stop-gap to display a graph of the workflow, but the intention is to replace this graph with something much better ASAP. 

This is still missing the following features:
* Automatically update the graph when builds for an event/pipeline update. WORKAROUND: refresh page
* Display meaningful icons for the colorblind. WORKAROUND: Status on hover
* Display a more minified graph for events
* Mobile-friendly view

There are probably a number of other things that should be cleaned up, too.

screen shot 2017-10-23 at 3 26 21 pm

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -1,5 +1,50 @@
 import Component from '@ember/component';
+import { get, computed, set } from '@ember/object';
+import { all, reject } from 'rsvp';
 
 export default Component.extend({
-  classNames: ['pipelineWorkflow']
+  classNames: ['pipelineWorkflow'],
+  graph: computed('workflowGraph', {
+    get() {
+      const jobs = get(this, 'jobs');
+      const fetchBuilds = [];
+      const graph = get(this, 'workflowGraph');
+
+      // Hack to make page display stuff when a workflow is not provided
+      if (!graph) {
+        return reject(new Error('No workflow graph provided'));
+      }
+
+      // Preload the builds for the jobs
+      jobs.forEach((j) => {
+        const jobName = get(j, 'name');
+
+        const node = graph.nodes.find(n => n.name === jobName);
+
+        // push the job id into the graph
+        if (node) {
+          node.id = get(j, 'id');
+          fetchBuilds.push(get(j, 'builds'));
+        }
+      });
+
+      return all(fetchBuilds).then(() => {
+        const builds = [];
+
+        // preload the "last build" data for each job for the graph to consume
+        jobs.forEach(j => builds.push(get(j, 'lastBuild')));
+
+        // set values to consume from templates
+        set(this, 'builds', builds);
+        set(this, 'visGraph', graph);
+
+        return graph;
+      });
+    }
+  }),
+
+  init() {
+    this._super(...arguments);
+    set(this, 'builds', []);
+  }
 });
--- a/app/components/workflow-graph-vis/component.js
+++ b/app/components/workflow-graph-vis/component.js
@@ -8,12 +8,12 @@ export default Component.extend({
   nodes: computed({
     get() {
       const nodes = getWithDefault(this, 'workflowGraph.nodes', []);
-      let list = nodes.map(n => ({ id: `${n.name}`, label: n.name }));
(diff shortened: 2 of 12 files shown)


app/components/workflow-graph-vis/component.js
```diff
@@ -0,0 +1,138 @@
...
+
+export default Component.extend({
+  router: service(),
+  nodes: computed({
+    get() {
+      const nodes = getWithDefault(this, 'workflowGraph.nodes', []);
+      let list = nodes.map(n => ({ id: `${n.name}`, label: n.name }));
+      const builds = get(this, 'builds');
+
+      if (builds) {
```
jithin1987: should check for `Array.isArray(builds) && builds.length`

app/components/pipeline-workflow/component.js
```diff
@@ -1,5 +1,50 @@
...
+        const node = graph.nodes.find(n => n.name === jobName);
+
+        // push the job id into the graph
+        if (node) {
+          node.id = get(j, 'id');
+          fetchBuilds.push(get(j, 'builds'));
+        }
+      });
+
+      return all(fetchBuilds).then(() => {
```
jithin1987: where is data from `fetchBuilds` used ?

petey: fetchBuilds is the list of promises that `all()` is observing to determine if the builds are loaded. This data is saved into the job models by way of those promises and is subsequently accessed when asking each job for `lastBuild` on line 35.

app/components/workflow-graph-vis/component.js
```diff
@@ -0,0 +1,138 @@
...
+        }
+      },
+      physics: {
+        enabled: false
+      }
+    };
+
+    const nodeSet = get(this, 'nodes');
+    const edgeSet = get(this, 'edges');
+    const el = this.$('.vis-network-graph')[0];
```
tkyi: el?

app/components/workflow-graph-vis/component.js
```diff
@@ -0,0 +1,138 @@
...
+      if (Array.isArray(builds) && builds.length) {
+        list = nodes.map((n) => {
+          const obj = { id: `${n.name}`, label: n.name };
+          const build = builds.find(j => `${j.get('jobId')}` === `${n.id}`);
+
+          if (build) {
+            obj.status = get(build, 'status');
+            obj.buildId = get(build, 'id');
+            obj.sha = get(build, 'sha');
+            obj.title = obj.status;
```
tkyi: What is `title` used for?

app/components/workflow-graph-vis/component.js
```diff
@@ -0,0 +1,138 @@
...
+import { get, getWithDefault, computed, set } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@ember/component';
+
+export default Component.extend({
+  router: service(),
+  nodes: computed({
+    get() {
+      const nodes = getWithDefault(this, 'workflowGraph.nodes', []);
+      let list = nodes.map(n => ({ id: `${n.name}`, label: n.name }));
```
tkyi: Does `n.name` need to be in a template string here?

app/components/workflow-graph-vis/component.js
```diff
@@ -0,0 +1,138 @@
...
+      const builds = get(this, 'builds');
+
+      if (Array.isArray(builds) && builds.length) {
+        list = nodes.map((n) => {
+          const obj = { id: `${n.name}`, label: n.name };
```
tkyi: Also here

app/components/workflow-graph-vis/component.js
petey: This is a DOM element.

app/components/workflow-graph-vis/component.js
petey: This provides the hover on the build bubbles.

PR #233, opened 2017-11-04
fix: new workflow graph breaks validator ui

petey:
Context:
========
After the workflow updates, the validator failed to display jobs for projects that did not explicitly define workflow

Objective:
-----------
Use the new workflowGraph property instead of workflow to display the workflow and job information

screen shot 2017-11-03 at 5 18 46 pm


d2lam:
Wow pretty nice 👍

PR #236, opened 2017-11-15
feat(782): New d3-based workflow graph

petey:
Context:
========
The vis-based workflow graphs lack some important information about build status. We are somewhat limited in what we can do WRT customizing the output graph.
Vis utilizes canvas which does not provide hooks that allow us to test whether the graph was drawn properly.

Objective:
----------
Build a new workflow view from scratch with d3. The intention is to have a graph that somewhat resembles a railroad diagram or subway map. 

* Display icons for build statuses (WIP - Final icons not yet available)
* Display contextual information about triggers ( dashed line -> event did not trigger, solid line -> trigger occurred )
* Allow click event to bubble up from component, if a handler is defined
* Maybe most importantly, allow the component and the resulting svg graph to be tested

screen shot 2017-11-15 at 6 47 50 pm

Misc:
------
Todo:
- [x] Use final icons
- [ ] ~~Dynamically scale graph nodes/edges based on canvas size~~
- [x] Minified view for event list
- [x] Animate running job


petey:
Update:
--------
* I'm making the graph math numbers a little less magic and base a lot of the positioning based on the font-size of the icons. The new math makes it a lot easier for us to tweak the positions between elements, e.g. the length of the edges.
* Added a flag to produce a "minified" graph for the event view, this should draw a graph roughly half the size of of the full one, at it's present configuration.
* I have a new icon font for the graph to replace the current one, but haven't added it yet.
* Fixed colors not appearing on edges, as intended, and made the icon text sharper.
* Centered job names under icons, and truncate job names that are too long

app/utils/graph-tools.js
```diff
@@ -0,0 +1,133 @@
...
+      if (n.name === start) {
+        n.status = 'STARTED_FROM';
+      }
+    });
+  }
+
+  // Calculate graph rows and columns with edges
+  for (let i = 0; i < edges.length; i += 1) {
+    const edge = edges[i];
+    const nextEdge = edges[i + 1];
```
minz1027: This part has an assumption that `edges` with same `src` or `dest` are close with each other. If I have a yaml defined like this: https://github.com/minz1027/hook-demo/blob/master/screwdriver.yaml, with`[~pr, ~commit]` in the first job and fourth job, will it draw the edges for `~pr` and `~commit` correctly?

app/components/workflow-graph-d3/component.js
```diff
@@ -0,0 +1,163 @@
...
+    svg.selectAll('link')
+      .data(data.edges)
+      .enter()
+      .append('path')
+      .attr('class', d => `graph-edge ${d.status ? `build-${d.status.toLowerCase()}` : ''}`)
+      .attr('stroke-dasharray', d => (!d.status ? 5 : 500))
+      .attr('stroke-width', 2)
+      .attr('fill', 'transparent')
+      .attr('d', (d) => {
+        const calcPos = (pos, spacer) =>
```
jithin1987: can you define this function outside of `attr`

app/utils/graph-tools.js
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      if (n.name === start) {
+        n.status = 'STARTED_FROM';
+      }
+    });
+  }
+
+  // Calculate graph rows and columns with edges
+  for (let i = 0; i < edges.length; i += 1) {
```
jithin1987: loop to `edges.length - ` and avoid `if (nextEdge) {` at line 90

app/utils/graph-tools.js
```diff
@@ -0,0 +1,133 @@
...
+
+  // Double check all nodes have positions (handle detached jobs with no edges)
+  nodes.forEach((n) => {
+    if (!n.pos) {
+      maxHeight += 1;
+      n.pos = { x: 0, y: maxHeight };
+    }
+  });
+
+  // For auto-scaling canvas size
```
jithin1987: is this still required with svg ?

app/utils/graph-tools.js
petey: It does draw `pr` and `commit` correctly, but `publish` and `other_publish` are on top of each other.

app/utils/graph-tools.js
petey: It is trivial to define a workflow with only one edge. That would cause this logic to break.

app/utils/graph-tools.js
petey: It seems like setting the SVG canvas to a percentage width breaks scrollbars when `overflow: auto` is set on the parent and the parent smaller than the SVG.

app/utils/graph-tools.js
petey: The overlapping issue is easily solved by changing the order the jobs are defined, but this isn't ideal. I'm open to suggestions on how to better calculate the rows and columns for the graph.

tkyi:
Let's revisit the bugs in later PRs.

PR #238, opened 2017-11-17
fix: new algorithm for graph rows/columns

petey:
Context:
========
We determined that the algorithm for calculating rows/columns for the workflow graph would end up displaying some bubbles on top of each other, or in other ways be difficult to read.

Objective:
----------
Change the algorithm for calculating rows and columns.

Goes from this:
screen shot 2017-11-17 at 1 55 41 pm

to this:
screen shot 2017-11-17 at 1 55 20 pm

### DIFF
--- a/app/utils/graph-tools.js
+++ b/app/utils/graph-tools.js
@@ -82,18 +82,10 @@ const decorateGraph = (inputGraph, builds, start) => {
 
   nodes.forEach((n) => {
     // Set root nodes on left
-    if (n.name.startsWith('~')) {
+    if (n.name.startsWith('~') || !n.pos) {
       n.pos = { x: 0, y: y[0] };
       y[0] += 1;
-      // recursively walk the graph from root
-      walkGraph(graph, n.name, 1, y);
-    }
-
-    // Set detached nodes on left
-    if (!n.pos) {
-      n.pos = { x: 0, y: y[0] };
-      y[0] += 1;
-      // recursively walk the graph from detached node
+      // recursively walk the graph from root/ detached node
       walkGraph(graph, n.name, 1, y);
     }
 
--- a/app/utils/graph-tools.js
+++ b/app/utils/graph-tools.js
@@ -22,12 +22,45 @@ const STATUS_MAP = {
 const node = (nodes, name) => nodes.find(o => o.name === name);
 
 /**
+ * Find a build for the given job id
+ * @method build
+ * @param  {Array} builds   List of build objects
+ * @param  {String} jobId   The job id of the build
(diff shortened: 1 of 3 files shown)


app/utils/graph-tools.js
```diff
@@ -42,90 +75,58 @@ const decorateGraph = (inputGraph, builds, start) => {
...
-  // Decorate job status, and build info
-  if ((Array.isArray(builds) || builds instanceof DS.PromiseArray) && get(builds, 'length')) {
-    nodes.forEach((n) => {
-      const build = builds.find(j => j && `${get(j, 'jobId')}` === `${n.id}`);
-
-      if (build) {
-        const status = get(build, 'status');
+  nodes.forEach((n) => {
+    // Set root nodes on left
+    if (n.name.startsWith('~') || !n.pos) {
```
jithin1987: is it guaranteed that detached nodes will always be at end of nodes list ?

petey: Not really, but they should always appear after `~pr` and `~commit`. https://github.com/screwdriver-cd/workflow-parser/blob/master/lib/getWorkflow.js#L18

app/utils/graph-tools.js
```diff
@@ -21,14 +21,47 @@ const STATUS_MAP = {
...
+ * Walks the graph to find siblings and set their positions
+ * @method walkGraph
+ * @param  {Object}  graph Raw graph definition
+ * @param  {String}  start The job name to start from for this iteration
+ * @param  {Number}  x     The column for this iteration
+ * @param  {Array}   y     Accumulator of column depth
+ */
+const walkGraph = (graph, start, x, y) => {
+  if (!y[x]) { y[x] = 0; }
+  const nodeNames = graph.edges.filter(e => e.src === start).map(e => e.dest);
```
jithin1987: is graph guaranteed not to have circles ? if it has circles then it will loop forever.

petey: config-parser should error when a graph contains a cycle.

app/utils/graph-tools.js
jithin1987: what about triggers ? Is there a chance that a job X which is reachable from  `~sd...`  appear before `~sd...` and gets processed as a detached job ?

petey: Yes, this looks a little weird as the `~sd@` job appears on the same line as the "detached" one. It is still kind of legible. I could try sorting the nodes by name to remove this problem.

petey: I'm now sorting the graph nodes to guarantee all the `~jobs` are processed first.

PR #239, opened 2017-11-22
fix(803): event model does not properly compute when it is finished

petey:
Context:
========
The event model is using the old workflow list instead of the new workflowGraph to compute whether it has completed all expected builds. This results in events with the new workflow from never having the isComplete flag true. This may result in unexpected api traffic for completed events.

Objective:
-----------
Use the new workflow graph to compute the number of expected jobs for an event. Pipelines configuring the old workflow will still have a new workflowGraph, so this change is backwards compatible.

Misc:
-----
* Removed the node sort in decorateGraph, as it is no longer necessary, and leads to confusion when comparing output and input.
* Added a event.status to tell the aggregate status of the event
* Wrapped status and isComplete in ObjectPromiseProxy since they depend on build information that comes as a PromiseArray.

### DIFF
--- a/app/event/model.js
+++ b/app/event/model.js
@@ -1,6 +1,12 @@
-import { computed } from '@ember/object';
+import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
+import ObjectProxy from '@ember/object/proxy';
+import { computed, get } from '@ember/object';
 import { sort, not } from '@ember/object/computed';
 import DS from 'ember-data';
+import graphTools from 'screwdriver-ui/utils/graph-tools';
+
+const { graphDepth } = graphTools;
+const ObjectPromiseProxy = ObjectProxy.extend(PromiseProxyMixin);
 
 export default DS.Model.extend({
   causeMessage: DS.attr('string'),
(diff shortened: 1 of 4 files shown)


app/event/model.js
```diff
@@ -20,47 +26,85 @@ export default DS.Model.extend({
...
+      return ObjectPromiseProxy.create({
+        promise: builds
+          .then((list) => {
+            // no builds yet
+            if (!get(list, 'length')) {
+              return false;
+            }
+
+            // Figure out how many builds we expect to run
+            const expectedBuilds = graphDepth(
```
tkyi: Can we move the `expectedBuilds` under the part where we figure out build failures? Since it doesn't seem like we use it for determining failed builds.

app/event/model.js
```diff
@@ -20,47 +26,85 @@ export default DS.Model.extend({
...
-
-      // last build failed, or all expected builds are successful
-      return true;
+      const builds = get(this, 'builds');
+
+      return ObjectPromiseProxy.create({
+        promise: builds
+          .then((list) => {
+            // no builds yet
+            if (!get(list, 'length')) {
```
tkyi: Is there any reason why we use `get(list, length)` syntax here when we use `list.length` on line 100?

PR #244, opened 2018-01-10
feat: add content-security-policy and other security headers

petey:
Context
========
The Content-Security-Policy HTTP response header helps you reduce XSS risks on modern browsers by declaring what dynamic resources are allowed to load via a HTTP Header.

Objective
---------
Add additional security headers in nginx serving platform

Misc
----
* [Content-Security-Policy](https://content-security-policy.com/) [[2](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)]
* [x-content-type-options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options)
* [x-frame-options](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options)
* [x-xss-protection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection)

Notes
------
I added numerous sha256 values to `style-src` due to dependencies adding inline styles. I'm afraid that this is somewhat brittle, and may bite us later when dependencies update, and we forget to update everything. I added these shas to avoid having to add `'unsafe-inline'` since that feels "unsafe". 

Running the docker container requires a new environment variable be added `AVATAR_HOSTNAME` to allow user images to be displayed. For github.com, this would be set to `*.githubusercontent.com`


petey:
I'm tempted to add another ENV variable to be required for the nginx config, e.g. `AVATAR_HOSTNAME`,  to be used in the `img-src` section.

prbinu:
Please test your CSP policy using https://csp-evaluator.withgoogle.com/

petey:
@prbinu Here's an example of the CSP policy for production:
Generated by `ember csp-headers --environment production`
```
style-src 'self' 'sha256-F7mkvbJTwqc33GcL7mev+qfhVmJ9akZNlMnRdJTmvjI=' 'sha256-Dn0vMZLidJplZ4cSlBMg/F5aa7Vol9dBMHzBF4fGEtk=' 'sha256-sA0hymKbXmMTpnYi15KmDw4u6uRdLXqHyoYIaORFtjU=' 'sha256-GK3QEnA3E0HIm/yl0gj7r7kmfONG84shoAcbrpNh9eg=' 'sha256-eU1D4PmGDxaJYeMSmyMsshvwPgzIiQpmijT8nLuPB8I=' 'sha256-47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU='; connect-src 'self' https://api.screwdriver.cd https://store.screwdriver.cd; img-src 'self' https: data: *.githubusercontent.com; default-src 'none'; script-src 'self'; font-src 'self'; media-src 'self';
```

The evaluator is concerned with `script-src 'self'` but Screwdriver doesn't host JSONP, Angular or user uploaded files.

prbinu:
LGTM

I would recommend to keep the sha256 for style-src directives.
Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

config/environment.js
```diff
@@ -5,6 +5,35 @@ module.exports = (environment) => {
...
+      'worker-src': [
+        'blob:'
+      ],
+      'img-src': [
+        "'self'",
+        'https:',
+        // ace editor
+        'data:',
+        // github avatars
+        '*.githubusercontent.com'
```
stjohnjohnson: avatars*.githubusercontent.com?

petey: FWIW, that configuration is only used when running `ember serve` which we don't use in any production environment.

d2lam:
Are we committing `package-json.lock` now? This is still using `node:6`.

petey:
This will run fine in node:8, but I can update that in another PR.

PR #247, opened 2018-01-19
fix: CSP errors on log page, and probably more

petey:
Context:
========
Lots of errors on build page:
```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'sha256-<snip>'". Either the 'unsafe-inline' keyword, a hash ('sha256-<snip>'), or a nonce ('nonce-...') is required to enable inline execution.
```

These do not currently appear to have a visual impact on the page.

Objective:
-----------
Since it is impossible to precalculate all the possible sha256 values that Glimmer [ember] and ACE Editor [validator] will generate, switch to use 'unsafe-inline' for style-src for now.


petey:
The build page is probably generating an one or more CSP errors per log line due to the log coloring we do.

PR #249, opened 2018-01-25
fix: force reload events model after starting a new build

jithin1987:
## Context

Reload timeout for events is configured to be 1.5 mins in production. This does not take into account user's clicking Start button to start a new event which results in users not seeing the new event unless they manually reload the page

## Objective

Reload events model after new event is started.

### DIFF
--- a/app/components/pipeline-events-list/component.js
+++ b/app/components/pipeline-events-list/component.js
@@ -4,7 +4,6 @@ import Component from '@ember/component';
 import ENV from 'screwdriver-ui/config/environment';
 
 export default Component.extend({
-  modelToReload: 'events',
   showMore: computed('events.length', 'numToShow', {
     get() {
       return get(this, 'numToShow') < get(this, 'events.length');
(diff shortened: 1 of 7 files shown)


app/components/pipeline-events-list/component.js
```diff
@@ -2,9 +2,8 @@ import { computed, get, set } from '@ember/object';
 import { sort } from '@ember/object/computed';
 import Component from '@ember/component';
 import ENV from 'screwdriver-ui/config/environment';
-import ModelReloaderMixin from 'screwdriver-ui/mixins/model-reloader';
 
-export default Component.extend(ModelReloaderMixin, {
+export default Component.extend({
   modelToReload: 'events',
```
petey: `modelToReload` isn't needed if `ModelReloaderMixin` isn't used

app/mixins/model-reloader.js
```diff
@@ -2,6 +2,8 @@ import { later, cancel } from '@ember/runloop';
 import Mixin from '@ember/object/mixin';
 import ENV from 'screwdriver-ui/config/environment';
 
+const FORCE_WAIT_TIMEOUT = 100;
```
petey: Can this magic number be moved to the environment config?

app/components/pipeline-events-list/component.js
jithin1987: done.

PR #252, opened 2018-01-30
fix: Fix templates bug when no template tags exist

ScottyFillups:
# Context

There is currently a bug on the templates page where templates with no template tags will not be loaded. This is because `this.get('template').getTemplateTags(params.name)` will `throw` an `404 Not Found` error, which isn't handled in the promise chain.

# Objective

Fix the bug by handling 404s.

### DIFF
--- a/app/templates/detail/route.js
+++ b/app/templates/detail/route.js
@@ -20,6 +20,6 @@ export default Route.extend({
       }
 
       return verPayload;
-    })
+    });
   }
 });
--- a/app/templates/detail/route.js
+++ b/app/templates/detail/route.js
@@ -7,17 +7,19 @@ export default Route.extend({
   model(params) {
     return RSVP.all([
       this.get('template').getOneTemplate(params.name),
-      this.get('template').getTemplateTags(params.name)
+      this.get('template').getTemplateTags(params.name).catch(err => err)
     ]).then((arr) => {
       const [verPayload, tagPayload] = arr;
 
-      tagPayload.forEach((tagObj) => {
-        const taggedVerObj = verPayload.find(verObj => verObj.version === tagObj.version);
+      if (tagPayload !== '404 Not Found') {
+        tagPayload.forEach((tagObj) => {
+          const taggedVerObj = verPayload.find(verObj => verObj.version === tagObj.version);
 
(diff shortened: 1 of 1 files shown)


app/templates/detail/route.js
```diff
@@ -7,15 +7,17 @@ export default Route.extend({
...
     return RSVP.all([
       this.get('template').getOneTemplate(params.name),
-      this.get('template').getTemplateTags(params.name)
+      this.get('template').getTemplateTags(params.name).catch(err => err)
     ]).then((arr) => {
       const [verPayload, tagPayload] = arr;
 
-      tagPayload.forEach((tagObj) => {
-        const taggedVerObj = verPayload.find(verObj => verObj.version === tagObj.version);
+      if (tagPayload !== '404 Not Found') {
```
tkyi: Should probably check if (tagPayload.status === 200), since `err` is most likely an error object, not a string

ScottyFillups: I tested it locally and it is a string; I'm unsure why.

tkyi: I see we usually reject with a String: https://github.com/screwdriver-cd/ui/blob/c2b8b5f0e5a8b8065c498e38b4fedfad048237ba/app/template/service.js#L44-L47

tkyi: Actually, I think it makes more sense to just return an empty array if no template tags exist rather than a 404.
So we can delete these lines: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/listTags.js#L27-L29

ScottyFillups: I agree. I've modified the code in the Screwdriver repo: https://github.com/screwdriver-cd/screwdriver/pull/863

Please don't merge this into master until further notice.

d2lam: @ScottyFillups  You can put [DNM] to the title of this PR if it's not supposed to be merged. Or add the tag `status/blocked` to this PR.

ScottyFillups:
The bug was fixed by making modifications to the API: https://github.com/screwdriver-cd/screwdriver/pull/863

Rather than making the UI handle a 404 error, it's passed an empty instead (ie. the current code should work currently as is).

PR #253, opened 2018-01-31
DNM: feat: require auth for build page views

petey:
Context
=======
Security requirements have been set requiring users to be authenticated to view logs and artifacts

Objective
----------
Add auth requirement to view the build page. Send Authentication headers to api calls to get logs and artifacts.

### DIFF
--- a/tests/unit/build-artifact/service-test.js
+++ b/tests/unit/build-artifact/service-test.js
@@ -1,5 +1,6 @@
 import { moduleFor, test } from 'ember-qunit';
 import Pretender from 'pretender';
+import Service from '@ember/service';
 let server;
 
 const manifest = `.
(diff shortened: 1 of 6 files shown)


tests/unit/build-artifact/service-test.js
```diff
@@ -35,12 +36,24 @@ const getManifest = () => {
...
+    authenticated: {
+      token: 'banana'
+    }
+  }
+});
+
 moduleFor('service:build-artifact', 'Unit | Service | build artifact', {
   // Specify the other units that are required for this test.
-  // needs: ['service:foo']
+  // needs: ['service:session'],
```
tkyi: Should this be uncommented?

petey: No. I'm mocking it instead of using the actual service.

PR #258, opened 2018-02-27
feat(806): Allow users to restart event jobs and detached pipelines

petey:
Context:
========
We want to allow users to restart jobs in a given event and retain that event's context in the new event. For example: restart a flakey job to continue the pipeline, or redeploy an artifact after a regression. Issue https://github.com/screwdriver-cd/screwdriver/issues/806

Objective:
-----------
Provide a mechanism that ties to the graph that allows users to start jobs. This generates a dropdown that can be populated with links and other information that allows users to go to the build details page, or restart a build/detached pipeline.

Tooltip:

Confirmation on restart:

Misc:
------
Restrictions:
* Build must be in the most recent event (this is subject to change)
* Build must have previously executed -or- be the root node of a detached workflow

Additional:
------------
Sort of addresses https://github.com/screwdriver-cd/screwdriver/issues/817 and https://github.com/screwdriver-cd/screwdriver/issues/878 by providing a standard html link to build details.

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -1,9 +1,7 @@
 import Component from '@ember/component';
 import { get, computed, set, setProperties } from '@ember/object';
 import { all, reject } from 'rsvp';
-import graphTools from 'screwdriver-ui/utils/graph-tools';
-
-const { isRoot } = graphTools;
+import { isRoot } from 'screwdriver-ui/utils/graph-tools';
 
 export default Component.extend({
   classNames: ['pipelineWorkflow'],
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -1,7 +1,9 @@
 import Component from '@ember/component';
 import { get, computed, set, setProperties } from '@ember/object';
 import { all, reject } from 'rsvp';
-import { isRoot } from 'screwdriver-ui/utils/graph-tools';
+import graphTools from 'screwdriver-ui/utils/graph-tools';
+
+const { isRoot } = graphTools;
 
 export default Component.extend({
   classNames: ['pipelineWorkflow'],
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -1,9 +1,13 @@
 import Component from '@ember/component';
-import { get, computed, set } from '@ember/object';
+import { get, computed, set, setProperties } from '@ember/object';
 import { all, reject } from 'rsvp';
+import graphTools from 'screwdriver-ui/utils/graph-tools';
+
+const { isRoot } = graphTools;
 
 export default Component.extend({
   classNames: ['pipelineWorkflow'],
+  showTooltip: false,
   graph: computed('workflowGraph', {
     get() {
       const jobs = get(this, 'jobs');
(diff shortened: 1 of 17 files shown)


app/components/pipeline-workflow/component.js
```diff
@@ -1,9 +1,13 @@
 import Component from '@ember/component';
-import { get, computed, set } from '@ember/object';
+import { get, computed, set, setProperties } from '@ember/object';
 import { all, reject } from 'rsvp';
+import graphTools from 'screwdriver-ui/utils/graph-tools';
+
+const { isRoot } = graphTools;
```
ScottyFillups: Why not `import { isRoot } from 'screwdriver-ui/utils/graph-tools';`? I don't see `graphTools` being used anywhere other than here.

app/pipeline/index/controller.js
```diff
@@ -82,6 +83,49 @@ export default Controller.extend(ModelReloaderMixin, {
...
+        parentBuildId,
+        parentEventId,
+        causeMessage
+      });
+
+      console.log('parentBuildId', parentBuildId);
+      console.log('parentEventId', parentEventId);
+      console.log('pipelineId', pipelineId);
+      console.log('startFrom', startFrom);
+      console.log('causeMessage', causeMessage);
```
ScottyFillups: Do we want these `console.log`s?

petey: I took those out, I think I may have forgotten to check that in.

app/components/pipeline-workflow/component.js
petey: good catch.

app/components/workflow-graph-d3/component.js
```diff
@@ -37,44 +37,39 @@ export default Component.extend({
...
-        return false;
-      }
+      const fn = get(this, 'graphClicked');
 
-      const fn = get(this, 'buildClicked');
-
-      // Properly handle job if one is passed
-      if (typeof fn === 'function') {
-        return fn(job);
+      if (!get(this, 'minified') && typeof fn === 'function') {
```
minz1027: I wonder when I click on the minified build bubble, what will happened?

app/components/workflow-tooltip/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+  transition: opacity .3s .1s;
+
+  &.show-tooltip {
+    opacity: 1;
+    visibility: visible;
+  }
+
+  .content {
+    position: relative;
+    font-size: 16px;
```
jithin1987: can we rely on `1rem` here. With 16px hard coded it won't be responsive with browser Font settings

app/components/workflow-tooltip/styles.scss
```diff
@@ -0,0 +1,65 @@
...
+    background-color: $sd-white;
+    border: 2px solid $grey-400;
+    border-radius: 2px;
+    box-shadow: 0 4px 16px 0 $box-shadow;
+    padding: 16px 18px;
```
jithin1987: relay on `em` if we are overriding node `font-size`

app/components/workflow-graph-d3/component.js
petey: it acts as though you clicked on the event, causing it to pop up to the big graph at the top.

app/components/workflow-tooltip/styles.scss
petey: Using `px` is consistent with our entire sass code base (with the exception of collections components), and is also consistent with the denali styles that this was pulled from. 

Apart from making the code consistent argument, I'm not entirely convinced that using relative size units has even been necessary since 2010, since the majority of users will increase page/font sizes using the zoom feature as opposed to digging into the browser's stylesheets. https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css

jithin1987: Using `rem` values provides much easier and cleaner means to scale up/down font-sizes (0.9rem/1.1rem) instead of using pixel values against each component. This also preserves user preference (it's a setting change in browsers), when user sets a permanent font-size instead of having to zoom in/out each time.

If this is how entire code base is then consistency wins.

PR #261, opened 2018-03-05
Fixed issue where last build status wasn't being displayed on dashboard

ethan-haynes:
How can it be reproduced?
  Viewing the `/dashboards/:id` page won't display last build status
What did you expect?
  To see the status of the last build
What actually occurred?
  There was no value being populated in the template

### DIFF
--- a/app/components/collection-view/component.js
+++ b/app/components/collection-view/component.js
@@ -50,19 +50,19 @@ export default Component.extend({
             ret.lastBuildStatus = '';
           }
 
-          ret.lastBuildIcon = (() => {
+          [ret.lastBuildIcon, ret.lastBuildStatusColor] = (() => {
             switch (ret.lastBuildStatus) {
             case 'started_from':
             case 'queued':
             case 'running':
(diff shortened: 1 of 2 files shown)


app/components/collection-view/component.js
```diff
@@ -44,10 +44,30 @@ export default Component.extend({
...
             ret.lastBuildTime = 0;
+            ret.lastBuildStatus = '';
           }
 
+          ret.lastBuildIcon = (() => {
+            switch (ret.lastBuildStatus) {
+            case 'started_from':
+            case 'queued':
+            case 'running':
+              return 'spinner fa-spin text-primary';
```
petey: `refresh` would be more consistent than `spinner` when compared to the pipeline graphs. The `text-primary` style isn't defined, so this isn't achieving any special coloring.

ethan-haynes:
Made the icon changes to keep them consistent. Also applied the class names to the text so that would be colored along with the icons. Since bootstrap is a dependency, used those color classes for the icons and text.

PR #262, opened 2018-03-05
fix: support restart from any past event

minz1027:
Support restart from any past event. The API part is done and in prod.

Related to: https://github.com/screwdriver-cd/screwdriver/issues/806

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -46,12 +46,10 @@ export default Component.extend({
       });
     }
   }),
-  displayRestartButton: computed('mostRecent', 'selectedEventObj', 'authenticated', {
+  displayRestartButton: computed('selectedEventObj', 'authenticated', {
     get() {
       // user is authenticated, and displaying builds in the most recent event
-      // TODO: remove restriciton for most recent event
-      return get(this, 'authenticated') &&
-        get(this, 'mostRecent') === get(this, 'selectedEventObj.id');
+      return get(this, 'authenticated');
     }
   }),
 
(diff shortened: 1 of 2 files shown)


app/components/pipeline-workflow/component.js
```diff
@@ -46,12 +46,10 @@ export default Component.extend({
       });
     }
   }),
-  displayRestartButton: computed('mostRecent', 'selectedEventObj', 'authenticated', {
+  displayRestartButton: computed('selectedEventObj', 'authenticated', {
```
petey: selectedEventObj is not needed here either.

app/components/pipeline-workflow/component.js
```diff
@@ -46,12 +46,10 @@ export default Component.extend({
...
     get() {
       // user is authenticated, and displaying builds in the most recent event
```
petey: please update the comment.

app/components/pipeline-workflow/component.js
```diff
@@ -46,12 +46,10 @@ export default Component.extend({
       });
     }
   }),
-  displayRestartButton: computed('mostRecent', 'selectedEventObj', 'authenticated', {
+  displayRestartButton: computed('authenticated', {
```
petey: Sorry to make you keep changing this, but since this really only maps to the value of authenticated, there is a simple shorthand that makes this easier.

```
import { alias } from '@ember/object/computed';
```

```
displayRestartButton: alias('authenticated'),
```

minz1027: This looks fancy! will address it

PR #264, opened 2018-03-16
fix: fixes for guest mode

petey:
Context:
========
A setting was added to screwdriver that allows a "guest mode" in which a user is authenticated as a guest and is allowed to access api endpoints using those credentials. This PR changes the screwdriver UI to require a user to authenticate to view any page, even if the user chooses to authenticate in guest mode. Those credentials will be passed along to any subsequent calls to the screwdriver api.

## Objective:
Prevent a guest user from receiving a blank page when navigating to pages where a valid credential is required to get data.

Gritty details:
------
* Add the AuthenticatedRouteMixin to all "root" routes (except login). This forces a user to log in for that route and any children of that route. When visiting a page (via a link) the user should be redirected back to that route after logging in.
* Cleanup of various modules using `this.get('property')` instead of `get(this, 'property')`, per ember style guides.
* When dealing with collections, use the Ember ArrayProxy methods instead of converting to an plain js array.
* Added additional metadata to the session data, including username, scmcontext, scope, and a flag that is set when the user is a guest.
* Fixed api requests being sent to retrieve collection data for dashboards when the user is not authenticated (or is a guest).
* Removed the ability for a guest user to visit certain routes.
* Fixed the error message that is displayed when a guest user attempts to perform an operation that is not permitted.
* Removed embedded title attributes from svg files which were being displayed instead of a more appropriate title defined in the template

## Related Links:
Issue: https://github.com/screwdriver-cd/screwdriver/issues/870


app/dashboard/show/controller.js
```diff
@@ -1,21 +1,22 @@
...
 
 export default Controller.extend({
   collection: alias('model'),
   actions: {
     removePipeline(pipelineId, collectionId) {
-      return this.store.findRecord('collection', collectionId)
+      return get(this, 'store').findRecord('collection', collectionId)
         .then((collection) => {
-          const pipelineIds = collection.get('pipelineIds') || [];
+          const pipelineIds = getWithDefault(collection, 'pipelineIds', []);
```
minz1027: `getWithDefault` is pretty cool! 👍

ScottyFillups: Reminds me of Scala `Option.getOrElse` 😄

petey: @ScottyFillups That sounds so threatening. Now, I'm imagining the Scala garbage collector coming around asking for protection money. "That's a pretty nice property you've got over there. It'd be a shame if something were to happen to it."

app/validator/route.js
```diff
@@ -1,4 +1,6 @@
 import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
 
-export default Route.extend({
+export default Route.extend(AuthenticatedRouteMixin, {
```
minz1027: For `validator`, do we need authentication? even the API endpoint requires no auth: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/validator.js

app/authenticators/screwdriver-api.js
```diff
@@ -41,17 +51,27 @@ export default Base.extend({
...
+
+        resolve(Object.assign({}, {
+          username,
+          scope,
+          scmContext,
+          isGuest: scope.includes('guest'),
+          token: data.token
+        }));
+
         return resolve(data);
```
tkyi: Do you still need this line?

petey: Good catch.

app/authenticators/screwdriver-api.js
```diff
@@ -26,7 +25,18 @@ function fetchToken() {
         withCredentials: true
       }
     })
-      .done(jwt => resolve(jwt))
+      .done((jwt) => {
+        // Add some data from the JWT to the session data
+        const { username, scope, scmContext } = decoder(jwt.token);
+
+        resolve(Object.assign({}, {
```
tkyi: do you need to return here too?

petey: It's probably not explicitly necessary, but I'll add it to be consistent.

app/validator/route.js
petey: Removed auth requirement.

PR #268, opened 2018-03-23
feat: add commands ui

yuichi10:
## Context
Recently sd-cmd wad published, and user want to see what kind of commands are able to be used as same as Templates.
So I create a page of commands.
Also there is no API to get tags of command, so in this PR, ui does not show tags.

## Objective
- Add command list page
- Add command detail page
- Able to change command infomation by version in detail page.

The view of list `/commands`

The view of detail `/commands/{namespace}/{name}`


app/components/command-list/template.hbs
```diff
@@ -0,0 +1,10 @@
+<h1>Commands <a href="https://docs.screwdriver.cd/user-guide/commands" title="More Information" target="_blank">{{fa-icon "question-circle"}}</a></h1>
+<div class="explain">Commands share a binaries (or scripts) across multiple jobs.</div>
```
tk3fftk: This should be just `binaries`.

yuichi10: I took this explain from [command design](https://github.com/screwdriver-cd/screwdriver/blob/master/design/commands.md#context). Also we can upload shell scripts, so it will be fine to write `(or scripts)`.

app/command/service.js
```diff
@@ -0,0 +1,51 @@
...
+    const url =
+      `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/commands/` +
+      `${encodeURIComponent(namespace)}/${encodeURIComponent(name)}`;
+
+    return this.fetchData(url);
+  },
+  getCommandTags(namespace, name) {
+    const url =
+      `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/commands/
+      ${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/tags`;
```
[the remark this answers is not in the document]
yuichi10: Yes the API to get tags list is not implemented yet and I do not use this function now. I will delete this function.

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
+<h4>Format: {{command.format}}</h4>
+
+{{#if isHabitat}}
+<div class="mode" title="This is the mode of the habitat.">
```
tkyi: I think it makes more grammatical sense to say `This is the habitat mode.`

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <span class="label">Mode:</span>
+  <span class="value">{{command.habitat.mode}}</span>
+</div>
+<div class="package" title="This is the package of the habitat.">
```
tkyi: `This is the habitat package.`

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <span class="label">Package:</span>
+  <span class="value">{{command.habitat.package}}</span>
+</div>
+<div class="habitat-command" title="This is the command of the habitat.">
```
tkyi: `habitat command`

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <span class="label">Command:</span>
+  <span class="value">{{command.habitat.command}}</span>
+</div>
+{{else if isDocker}}
+<div class="image" title="This is the image of the docker.">
```
tkyi: `docker image`

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <span class="label">Image:</span>
+  <span class="value">{{command.docker.image}}</span>
+</div>
+<div class="docker-command" title="This is the command of the docker.">
```
tkyi: `docker command`

app/components/command-format/template.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <span class="label">Command:</span>
+  <span class="value">{{command.docker.command}}</span>
+</div>
+{{else if isBinary}}
+<div class="file" title="This is the command of the binary.">
```
tkyi: `binary command`

app/components/command-list/template.hbs
tkyi: Can remove `a` from `a binaries`

app/components/command-format/template.hbs
yuichi10: Thank you! I will fix all of these English mistake.

app/components/command-format/template.hbs
yuichi10: Sorry I use `command` but it will be better to use `file`. 
I will fix to say `binary file`.

PR #269, opened 2018-03-28
feat(911): Show sourcePaths in validator [6]

tkyi:
## Context
Now that users can configure `sourcePaths` in their `screwdriver.yaml` files, we should reflect those paths in the validator.

## Objective
This PR adds `sourcePaths` in the validator UI.

Array of strings example:
screen shot 2018-03-29 at 1 26 01 pm
String example:
screen shot 2018-03-29 at 1 26 46 pm

## Related links
Blocked by https://github.com/screwdriver-cd/screwdriver/pull/940 (still needs to be tested and pulled into prod)
Related to https://github.com/screwdriver-cd/screwdriver/issues/911

### DIFF
--- a/app/components/validator-job/template.hbs
+++ b/app/components/validator-job/template.hbs
@@ -39,7 +39,7 @@
     </ul>
   </span>
 </div>
-<div class="sourcePaths" title="These are the source paths that will trigger a job upon modification.">
+<div class="sourcePaths" title="These are the source paths that will trigger a job upon modification. If you want to specify a directory, make sure it has a trailing slash (/).">
   <span class="label">Source Paths:</span>
   <span class="value">
     <ul>
--- a/app/components/validator-job/template.hbs
+++ b/app/components/validator-job/template.hbs
@@ -39,6 +39,18 @@
     </ul>
   </span>
 </div>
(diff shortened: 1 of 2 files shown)


app/components/validator-job/template.hbs
```diff
@@ -39,6 +39,18 @@
     </ul>
   </span>
 </div>
+<div class="sourcePaths" title="These are the source paths that will trigger a job upon modification.">
```
d2lam: Maybe specify that if they want "directory" then make sure it has a `/` at the end. Otherwise it assumes that it's a filename

tests/integration/components/validator-job/component-test.js
```diff
@@ -180,6 +183,24 @@ test('it renders a description', function (assert) {
...
 
+test('it renders sourcePaths', function (assert) {
+  this.set('jobMock', {
+    image: 'int-test:1',
+    description: 'This is a description',
+    secrets: [],
+    environment: {},
+    settings: {},
+    annotations: {},
+    sourcePaths: ['README.md', 'src/folder']
```
d2lam: this would be `src/folder/` to be accurate?

tkyi: Good catch; technically our data-schema examples are wrong now as well

PR #270, opened 2018-03-28
fix: fix logs overlap problem in chrome

minz1027:
It turned out that chrome has problem with `display: grid` which caused the log lines overlap problem.

Switched it back to `display: flex` fixed the problem.

e.g. build with more than 5000 lines of logs in a step
screen shot 2018-03-28 at 5 13 49 pm

Related to: https://github.com/screwdriver-cd/screwdriver/issues/949

### DIFF
--- a/app/components/build-log/styles.scss
+++ b/app/components/build-log/styles.scss
@@ -40,20 +40,37 @@
 .logs {
   font-size: 1em;
   font-family: monospace;
-  height: auto;
+  white-space: pre-line;
+  background-color: #333;
+  color: #eee;
+  display: flex;
+  flex-direction: column;
+  align-content: stretch;
   overflow: auto;
-  display: grid;
-  grid-template-columns: 1fr 10fr;
+  height: auto;
+
+  .line {
+    display: flex;
+    flex-direction: column;
+  }
 
   .time {
     color: $sd-text-med;
-    grid-column: 1;
-    user-select: none;
   }
 
   .content {
-    grid-column: 2;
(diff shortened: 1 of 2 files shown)


vengopal:
👍

app/components/build-log/styles.scss
```diff
@@ -40,20 +40,38 @@
 .logs {
   font-size: 1em;
   font-family: monospace;
-  height: auto;
+  white-space: pre-line;
+  background-color: #333;
```
tkyi: is there any reason why we're not using colors from https://github.com/screwdriver-cd/ui/blob/master/app/styles/screwdriver-colors.scss?

app/components/build-log/styles.scss
```diff
@@ -40,20 +40,38 @@
...
+  .line {
+    display: flex;
+    flex-direction: column;
+  }
 
   .time {
     color: $sd-text-med;
-    grid-column: 1;
-    user-select: none;
+    display: none;
```
ScottyFillups: Why is display set to `none` now?

app/components/build-log/styles.scss
minz1027: we don't actually have that color in the file. that file only has a black black. this one is lighter. Do you want to to add this color there? but probably won't reuse it any where

tkyi: Hm what were we using before? Or is this just a new color?
It's okay to leave it hardcoded if it's just used here.

PR #274, opened 2018-03-30
Show repo url beside template name in template page

pranavrc:
# Context

Currently, there's no way to find out a template's source repo without using the search function.

# Objective
 
- [x] Show link to source repo beside template name on the template page

image

# Related links

Issue that tracks this - https://github.com/screwdriver-cd/screwdriver/issues/522

### DIFF
--- a/app/components/template-header/component.js
+++ b/app/components/template-header/component.js
@@ -1,10 +1,18 @@
 import Component from '@ember/component';
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
   templateToRemove: null,
+  scmUrl: null,
   isRemoving: false,
+  store: service(),
   init() {
     this._super(...arguments);
+    this.get('store').findRecord('pipeline', this.template.pipelineId).then((pipeline) => {
+      this.set('scmUrl', pipeline.get('scmRepo.url'));
+    }).catch(() => {
+      this.set('scmUrl', null);
+    });
   },
   actions: {
     setTemplateToRemove(template) {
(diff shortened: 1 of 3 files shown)


app/components/template-header/component.js
```diff
@@ -1,10 +1,16 @@
...
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
   templateToRemove: null,
+  scmUrl: null,
   isRemoving: false,
+  store: service(),
   init() {
     this._super(...arguments);
+    this.get('store').findRecord('pipeline', this.template.pipelineId).then((pipeline) => {
```
Zhongtang: What will happen if the pipeline was already deleted?

pranavrc: Ideally a template without a pipeline should never exist because:
- A template without a pipeline cannot be deleted.
- The pipeline cannot be recreated because the template already exists and publish will always fail.
This needs an API fix. I've let @ScottyFillups know, he's fixing it.

ScottyFillups: Regarding template and tag cleanup after pipeline removal:

Issue: https://github.com/screwdriver-cd/screwdriver/issues/969
PR: https://github.com/screwdriver-cd/screwdriver/pull/970

pranavrc:
After @d2lam's comments in https://github.com/screwdriver-cd/screwdriver/pull/970 we decided to just gray out the buttons next to the template if its pipeline doesn't exist.

image

tkyi:
I kind of feel like things shouldn't show up if they're not usable. (Aka the source link should not show up if the pipeline doesn't exist.)
Delete icon should also not show up if the person doesn't have permissions to delete.

ScottyFillups:
Regarding Tiff's comment, I am in agreement of:

> Delete icon should also not show up if the person doesn't have permissions to delete.

99% of the time (I'm exaggerating, but bear with me 😛 ), you'll be viewing a template that's not yours. That means that the delete icon is unnecessary 99% of the time, and just taking up space. IMHO it's better to only show it when it's needed. If a user doesn't see a delete button, it's obvious why: they don't have admin privileges for the template. However, I have mixed feelings on the following suggestion:


> [...] the source link should not show up if the pipeline doesn't exist

While it is clear why the delete button would be missing, it's not so clear why the source link would be missing. I'm worried users would mistake this for a bug; it's rare for a pipeline that publishes templates to be removed.

tkyi:
Agree with Philip, but we can do that in a different PR. Will approve this one to unblock this one for now.
Tracking that issue here: https://github.com/screwdriver-cd/screwdriver/issues/967

PR #275, opened 2018-04-02
fix(911): Fix PR icon for non-running PR

tkyi:
## Context
We don't set the build status to CREATED for PRs that shouldn't be run. The builds are not even created.

## Objective
This PR fixes the icon switch statement to default to the banned icon, since we'll never reach the `CREATED` case.

## Related links
Related to https://github.com/screwdriver-cd/ui/pull/272
Original issue: https://github.com/screwdriver-cd/screwdriver/issues/911

### DIFF
--- a/app/components/pipeline-pr-view/component.js
+++ b/app/components/pipeline-pr-view/component.js
@@ -21,11 +21,12 @@ export default Component.extend({
       case 'SUCCESS':
         icon = 'check';
         break;
-      case 'CREATED':
-        icon = 'fa-ban';
+      case 'ABORTED':
+      case 'FAILURE':
+        icon = 'times';
         break;
       default:
-        icon = 'times';
+        icon = 'fa-ban';
         break;
       }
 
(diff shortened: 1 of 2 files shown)


app/components/pipeline-pr-view/component.js
```diff
@@ -21,11 +21,12 @@ export default Component.extend({
       case 'SUCCESS':
         icon = 'check';
         break;
-      case 'CREATED':
-        icon = 'fa-ban';
+      case 'ABORTED':
```
d2lam: Should we use the pause icon to be consistent with the pipeline view?

tests/integration/components/pipeline-pr-view/component-test.js
```diff
@@ -36,7 +36,7 @@ test('it renders a created (but not running) PR', function (assert) {
     name: 'PR-1234',
     builds: [{
       id: '1234',
-      status: 'CREATED'
+      status: null
```
d2lam: In the case we are talking about, I think there will be the job but no builds. So idk if this will work, since this assume there is a build and there is no status.

tkyi: Changed the test to reflect this.

app/components/pipeline-pr-view/component.js
tkyi: 🤷‍♀️  Not sure what makes the most sense. The pipeline graph is also using green and red while the PR icons are either green or blue.
I'm going to leave it for now in this PR. We can revisit this in the future.

PR #283, opened 2018-04-26
fix(1028): Parent meta exists when restarting a build from build page

tkyi:
## Context
Parent meta is not used when a user tries to Restart a build from the build detail page.

## Objective
This PR uses parent meta to start a build.

## Related links
Resolves https://github.com/screwdriver-cd/screwdriver/issues/1028

### DIFF
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -28,21 +28,13 @@ export default Controller.extend({
 
     startBuild() {
       const buildId = get(this, 'build.id');
-      const parentBuildId = get(this, 'build.parentBuildId');
-      const event = get(this, 'event');
-      const parentEventId = get(event, 'id');
-      const startFrom = get(this, 'job.name');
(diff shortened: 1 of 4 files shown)


app/pipeline/build/controller.js
```diff
@@ -24,20 +27,31 @@ export default Controller.extend({
...
+      const event = get(this, 'event');
+      const parentEventId = get(event, 'id');
+      const startFrom = get(this, 'job.name');
+      const pipelineId = get(this, 'pipeline.id');
+      const token = get(this, 'session.data.authenticated.token');
+      const user = get(decoder(token), 'username');
+      const causeMessage =
+        `${user} clicked restart for job "${startFrom}" for sha ${get(event, 'sha')}`;
       const newEvent = this.store.createRecord('event', {
+        buildId,
```
minz1027: You only need the buildId and causeMessage. The rest the API will figure out

tkyi: Updated

PR #292, opened 2018-05-15
add link to pr

lusol:
adding link to PR for pull requests running

screen shot 2018-05-14 at 10 31 46 pm

### DIFF
--- a/app/components/build-banner/template.hbs
+++ b/app/components/build-banner/template.hbs
@@ -33,6 +33,19 @@
     <span class="banner-value">{{buildContainer}}</span>
     <span class="banner-label">Container</span>
   </li>
+  {{#if event.pr.url}}
+    <li class="pr">
+      <span class="banner-value">
+        <div class="pr-url-holder">
+          <a href="{{event.pr.url}}" class="sha">{{inline-svg "github" class="pr-img"}}</a>
+        </div>
+      </span>
+      <span class="banner-label">Pull Request</span>
+    </li>
+  {{else}}
+    <li class="empty">
+    </li>
+  {{/if}}
   <li class="call-to-action">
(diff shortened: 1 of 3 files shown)


app/components/build-banner/template.hbs
```diff
@@ -33,6 +33,19 @@
     <span class="banner-value">{{buildContainer}}</span>
     <span class="banner-label">Container</span>
   </li>
+  {{#if event.pr.url}}
+    <li class="pr">
+      <span class="banner-value">
```
minz1027: Instead of adding a new banner column, how about just link the job name to PR? Currently it's not clickable

pranavrc:
Add integration tests to `tests/integration/components/build-banner/component-test.js` and tests that use the events model, like the pipeline controller maybe

app/components/build-banner/template.hbs
jithin1987: I think that's not very intuitive for users to identify with job name as a link to PR page. I think we should be having enough space to display this info. as a new column. If not we can revisit.

pranavrc: Think there should be a blue link with PR-<number> next to the github icon in the column. Just like there's a blue link next to the user's avatar in the USER column. Otherwise it's not obvious that it's a link to the PR.

lusol:
woops, fixed that

pranavrc:
Can you update the screenshot to show what it looks like in each of the if cases (with coverage/without)?

lusol:
Without coverage
screen shot 2018-05-16 at 3 53 46 pm

With coverage
screen shot 2018-05-16 at 3 55 24 pm

lusol:
@jithin1987  YOU ARE VERY CORRECT, fixed it.

PR #297, opened 2018-05-31
feat: restore builds route

jithin1987:
## Context

Adds route `/builds/buildId` to UI which will redirect to actual url for the build. This is useful when debugging build errors when all you have is the build Id.

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -10,6 +10,7 @@ const Router = EmberRouter.extend({
 /* eslint-disable array-callback-return */
 Router.map(function route() {
   this.route('home', { path: '/' });
+  this.route('builds', { path: '/builds/:build_id' });
   this.route('pipeline', { path: '/pipelines/:pipeline_id' }, function secretsRoute() {
     this.route('secrets');
     this.route('build', { path: 'builds/:build_id' });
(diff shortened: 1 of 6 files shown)


app/router.js
```diff
@@ -10,6 +10,7 @@ const Router = EmberRouter.extend({
 /* eslint-disable array-callback-return */
 Router.map(function route() {
   this.route('home', { path: '/' });
+  this.route('builds', { path: '/builds/:build_id' });
```
tkyi: should this be `this.route('build', ` for consistency/since it takes you to one build?

jithin1987: that's inconsistent with rest api urls, ideally it should match. `/builds/1` takes you to a single build but it's not `build/1`

tkyi: Ah so it should be `this.route('pipelines',` for line 14 and `this.route('builds',` for line 16

jithin1987: yep, that's my 2 cents on this.

tkyi: Are we going to fix it?

jithin1987: later, that needs directory renaming and more unnecessary changes at this point.

PR #299, opened 2018-06-01
feat(926): Display template namespace if it exists

tkyi:
## Context
Should display namespace/name if namespace exists and is not "default". Also showing namespace and name in the template detail page now.

## Objective
This PR should fix it so it still shows namespace and name.

Example with namespace:
screen shot 2018-05-31 at 5 24 27 pm

Example with default namespace:
screen shot 2018-05-31 at 5 24 13 pm

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/926


app/templates/index/route.js
```diff
@@ -12,6 +12,12 @@ export default Route.extend({
       templates.forEach((t) => {
         if (!names[t.name]) {
           names[t.name] = 1;
+          // construct full template name
+          t.fullName = t.name;
```
jithin1987: can we define this in a common place, second time this logic is repeated.

app/components/template-section/template.hbs
```diff
@@ -1,8 +1,8 @@
 <article class="template">
   <header>
-    {{#link-to "templates.detail" template.name}}<h4>{{template.name}}</h4>{{/link-to}}
+      {{#link-to "templates.detail" template.fullName}}<h4>{{template.fullName}}</h4>{{/link-to}}
```
jithin1987: an extra space

PR #300, opened 2018-06-05
fix bug where older pr events couldn't display dropdown properly

lusol:
There was a bug where older pr events would prevent dropdowns of PRs from displaying properly

### DIFF
--- a/app/pr-events/service.js
+++ b/app/pr-events/service.js
@@ -6,7 +6,7 @@ import ENV from 'screwdriver-ui/config/environment';
 export default Service.extend({
   session: service(),
   /**
-   * Calls the logs api service to fetch logs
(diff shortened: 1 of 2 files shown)


tkyi:
Probably should add a test to check for this case

app/pr-events/service.js
```diff
@@ -26,18 +25,15 @@ export default Service.extend({
           type: 'pr'
         },
         contentType: 'application/json',
-        crossDomain: true,
-        xhrFields: {
-          withCredentials: true
```
tkyi: Pretty sure you need credentials for this call; should probably put this section back.
Can add something like this to your test:
https://github.com/screwdriver-cd/ui/blob/master/tests/unit/template/service-test.js#L5-L11

PR #301, opened 2018-06-05
fix build page spamming api constantly when build is queued

petey:
Context
=======
When a build page is opened for a build that is still enqueued, the API will be spammed several times a second. This happens because the build-banner component calls `reloadBuild()` when it is either queued or running during the willRender event of the component lifecycle. When the build info is fetched, it causes the willRender event to fire again, and this loop can be extremely short. The current reload event only allows firing once per render loop, but this loop is very fast.

Objective
----------
Hit the API only as often as `BUILD_RELOAD_TIMER`, by throttling all requests to the reload handler.

Misc
------
Fix tests broken in firefox due to nav-banner component writing a value after it has been destroyed.

### DIFF
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -1,4 +1,4 @@
-import { later, once } from '@ember/runloop';
+import { later, throttle } from '@ember/runloop';
 import { get, computed } from '@ember/object';
 import { reads, mapBy } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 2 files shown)


app/pipeline/build/controller.js
```diff
@@ -1,4 +1,4 @@
-import { later, once } from '@ember/runloop';
+import { later, throttle } from '@ember/runloop';
```
tkyi: cool! 👍

PR #303, opened 2018-06-08
can press enter to create pipeline

lusol:
You can now create a pipeline by hitting the enter key on your key board instead of having to press the create button with your mouse.


app/components/pipeline-create-form/template.hbs
```diff
@@ -9,6 +9,7 @@
   key-up="scmChange"
   placeholder="Enter your repository url (eg: git@github.com:screwdriver-cd/ui.git#master)"
   value=scmUrl
+  enter="canSave"
```
jithin1987: can't you just do `saveData` here ? why new action ?

app/components/pipeline-create-form/component.js
```diff
@@ -20,6 +20,11 @@ export default Component.extend({
   }),
 
   actions: {
+    canSave() {
```
jithin1987: i think you don't need this, just hook up to `saveData` directly on enter.

PR #304, opened 2018-06-11
feat(937): Display child pipeline secrets

Filbird:
## Context
Child pipelines inherit secrets from their parent pipeline and can override them.

## Objective
The secrets view for child pipelines should show secrets inherited from their parent pipeline. These secrets should be differentiated and should not be allowed to be updated/deleted. 

Instead, these secrets can be overridden by creating a secret of the same name from the child pipeline. Conversely the secrets can be reverted back to their parent pipeline's values.

screen shot 2018-06-12 at 10 56 40 am

## Reference

### DIFF
--- a/app/components/secret-view/component.js
+++ b/app/components/secret-view/component.js
@@ -7,9 +7,28 @@ export default Component.extend({
   originalAllowInPR: null,
   buttonAction: computed('newValue', 'secret.allowInPR', 'originalAllowInPR', {
     get() {
+      const secret = this.get('secret');
+      const pipeline = this.get('pipeline');
+
+      if (secret.get('pipelineId') === pipeline.get('configPipelineId')) {
+        return 'Override';
+      }
+
       return (this.get('newValue')
         || this.get('originalAllowInPR') !== this.get('secret.allowInPR')) ?
         'Update' : 'Delete';
+    }
+  }),
+  passwordPlaceholder: computed({
+    get() {
+      const secret = this.get('secret');
+      const pipeline = this.get('pipeline');
+
+      if (secret.get('pipelineId') === pipeline.get('configPipelineId')) {
+        return 'Inherited from config pipeline';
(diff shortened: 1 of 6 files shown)


Filbird:
@tkyi Sounds reasonable! Where do you figure the best place to display that description is? Underneath `Secrets` or perhaps underneath the table?

jithin1987:
@Filbird same place as we display current message. https://github.com/screwdriver-cd/ui/blob/0ffe5365253fd5f7167852563fd3df69280bb5b4/app/components/pipeline-secret-settings/template.hbs#L9  Replace it with something of the line "Secrets are inherited from [parent pipeline](Link to parent pipeline secrets)". You have the option to update a Secret with new value.

Filbird:
@jithin1987  Updated 😄 
screen shot 2018-06-13 at 11 50 25 am

app/components/secret-view/component.js
```diff
@@ -20,16 +45,27 @@ export default Component.extend({
     modifySecret() {
       const secret = this.get('secret');
 
-      if (this.get('buttonAction') === 'Delete') {
-        secret.destroyRecord();
-      } else {
+      if (this.get('buttonAction') === 'Delete'
+        || this.get('buttonAction') === 'Revert') {
+        return secret.destroyRecord().then(() => {
+          this.get('secrets').store.unloadRecord(secret);
```
minz1027: Why we still need to unloadRecord after `destroyRecord`? based on the doc, after the destroy it should be gone both locally and backend?

Filbird: I had thought so as well, but then I was running into an issue similar to : https://github.com/emberjs/data/issues/4972.

In our case, if we revert then override the same secret we get a duplicate id error.

minz1027: ah, got it! thx for explaining!

PR #306, opened 2018-06-11
Ember 3.1.x Upgrade

petey:
Context
=======
* Keeping up-to-date, ember 3.1 was release 4/13.
* ember-cli-babel 5.x has been deprecated. A number of required addons were using this plugin.
* Bower support in ember has been deprecated for quite some time, and bower itself has recommended moving away since 10/2017 

Objective
----------
* Update to emberjs 3.1.x
* Update all addons to their most recent published versions. (resolve use of ember-cli-babel 5.x)
* Remove all remaining Bower dependencies (transitive dependencies from jstree, highlight addons)

Important
----------
* `ember-cli-jstree` has the required changes on their master branch, but have been lax in publishing version 1.0.10 so this module is pinned to HEAD of the master branch
* `ember-inline-svg` has an open PR fixing the deprecations, but this hasn't been merged, and there's been little activity from the maintainer. Pinned to the HEAD of the master branch for the PR.


.eslintrc.js
```diff
@@ -35,8 +35,9 @@ module.exports = {
     // node files
     {
       files: [
-        'testem.js',
         'ember-cli-build.js',
+        'testem.js',
+        'blueprints/*/index.js',
```
jithin1987: we don't have any blueprints yet, is this change required ?

petey: I don't expect screwdriver will ever have blueprints, but this change is part of the defaults provided by `ember init`. I'd prefer to keep things close to the defaults so it isn't as big a diff to manage with every update.

petey: I can reproduce the build errors when installing dependencies with `npm` as opposed to `yarn`, which works fine locally. Seems to be related to `ember-cli-jstree`, which is pinned to the repo's HEAD. This may have something to do with the way that dependency is brought in by npm.

I'm not actually expecting this to be merged until those two pinned dependencies have produced new versions with the appropriate changes.

PR #307, opened 2018-06-12
feat(937): Hide pipeline options for child pipeline

Filbird:
## Context
Child pipelines should only be able to be updated/deleted via changing the `screwdriver.yaml` of its parent pipeline.

## Objective
Hide the "Pipeline" and "Danger zone" sections from the pipeline options UI.

screen shot 2018-06-12 at 10 54 11 am

## Reference

### DIFF
--- a/app/components/pipeline-options/template.hbs
+++ b/app/components/pipeline-options/template.hbs
@@ -1,8 +1,7 @@
 {{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
 
 <div class="row">
-  {{#if pipeline.configPipelineId}}
-  {{else}}
+  {{#unless pipeline.configPipelineId}}
     <div class="col-xs-12 col-md-8">
       <section class="pipeline">
         <h3>Pipeline</h3>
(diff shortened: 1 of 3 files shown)


app/components/pipeline-options/template.hbs
```diff
@@ -1,41 +1,44 @@
...
-          </div>
-          <div class="row">
-            <div class="col-xs-10">
-              <div>
-                {{input
-                  class="text-input"
-                  key-up="scmChange"
-                  value=scmUrl
-                }}
+  {{#if pipeline.configPipelineId}}
```
tkyi: would it be cleaner to use `unless`?
```
{{#unless pipeline.configPipelineId}}
{{/unless}}
```

app/components/pipeline-options/template.hbs
```diff
@@ -99,35 +102,38 @@
...
-        <li>
-          {{#if showDangerButton}}
-          <div class="row">
-            <div class="col-xs-1 col-md-8">
-              <h4>Remove this pipeline</h4>
-              <p>Once you remove a pipeline, there is no going back.</p>
-            </div>
-            <div class="col-xs-1 col-md-4 right">
-              <a href="#" {{action "showRemoveButtons"}} class="trash">{{fa-icon "trash"}}</a>
+  {{#if pipeline.configPipelineId}}
```
tkyi: same comment as above

PR #308, opened 2018-06-12
feat(937): add child pipelines tab

minz1027:
## Context
Only config pipeline will see the child pipeline tab. And they can start all child pipelines.

screen shot 2018-06-12 at 3 11 16 pm

## To Do
Will wait for the design mock to see what to add to the child pipeline list and add a link to guide when we have the guide

## Reference


app/components/pipeline-list/template.hbs
```diff
@@ -0,0 +1,26 @@
...
+    <button {{action "startAll"}} class="start-button">Start All</button>
+  </div>
+  <ul class="col-xs-12 col-md-8">
+    {{#each pipelines as |pipeline|}}
+      <li class="appId">{{#link-to "pipeline" pipeline.id}}{{pipeline.appId}}{{/link-to}}</li>
+    {{/each}}
+  </ul>
+{{else}}
+  <div class="num-results">
+    <span>No child pipeline created</span>
```
tkyi: No child pipeline`s` created

Should we link to some documentation here?

tkyi:
Wondering if we should have a title on this tab. Not sure if it's necessary.

app/components/pipeline-list/template.hbs
minz1027: I will add it when we have the guide. :)

Filbird:
Thoughts on renaming `childpipelines` to use spinal-case? e.g. `child-pipelines`.

minz1027:
@Filbird addressed!

tkyi:
Jw since we decided on the whole hiding-things-from-users-if-they-can't-do-anything

In that case I think it should be okay for people to see the Child Pipelines tab but the Start All button should be greyed out or should not show if they do not have permission

minz1027:
Not like the child pipeline page, we don’t know if user has permission when we render the page and we don’t have an endpoint for that.

PR #309, opened 2018-06-14
Show full job names

awick:
Show the full job names by increasing the X distance based on the max job name length.  If job names aren't shown then still use 2*ICONSIZE.

This is for https://github.com/screwdriver-cd/screwdriver/issues/1124

### DIFF
--- a/app/components/workflow-graph-d3/component.js
+++ b/app/components/workflow-graph-d3/component.js
@@ -177,7 +177,7 @@ export default Component.extend({
 
         path.moveTo(startX, startY);
         // curvy line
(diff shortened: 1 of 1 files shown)


tkyi:
Can you include before/after screenshots?

awick:
Old: 
image

New:
image


I'm just using https://github.com/awick/sd-test for my test, do you have a good pathological file to test with? :)

tkyi:
Hmm I don't really have a good example. But yeah I have noticed that we tend to have a lot of real estate horizontally so it shouldn't be too big of an issue to make this change atm.

@jithin1987 @catto What do you think?

PR #310, opened 2018-06-15
feat(937): Add link to parent on child pipeline header

Filbird:
## Context
Users viewing a child pipeline's page should be able to directly navigate to its parent pipeline.

The "Child pipelines" tab should be on the leftmost column for parent pipelines.

### Update
Move the link to the top left portion of the header, add a fa-icon to accompany it.
screen shot 2018-06-15 at 4 45 49 pm

Additionally move the child pipelines tab to the left most column.

screen shot 2018-06-15 at 4 46 01 pm

---
Add a link to the parent pipeline on the child pipeline header.

screen shot 2018-06-14 at 6 48 59 pm

Pretty barebones, later on we can style it appropriately.

## Reference


russdeluca:
Will users only get to the child page from the parent?  If not, I'd just have the link text be "Parent pipeline" instead of "Back to parent pipeline".   Also, either way, I recommend using a descriptive class name, so something like "parent-pipeline" instead of "back", which makes it easy later on to understand exactly what is being styled in the CSS.

Filbird:
Thanks for your input @russdeluca!

Users can also get to the child pipeline via search, similar to regular pipelines. 

As for the text and CSS, I mostly opted to follow the convention set in the templates UI.
https://github.com/screwdriver-cd/ui/blob/master/app/components/template-header/template.hbs#L1
However, seeing as the two are not quite alike, I agree that it makes more sense for "Parent pipeline" to be treated as a regular part of the header than as a bread crumb.

Any thoughts @jithin1987  @minz1027 @tkyi ?

app/components/pipeline-header/template.hbs
```diff
@@ -1,3 +1,6 @@
 {{#link-to "pipeline" pipeline.id}}<h1>{{pipeline.appId}}</h1>{{/link-to}}
 <a href="{{pipeline.hubUrl}}" class="branch">{{fa-icon "code-fork"}} {{pipeline.branch}}</a>
-<span class="scm">{{fa-icon scmContext.scmIcon}} {{scmContext.scm}}</span>
\ No newline at end of file
+<span class="scm">{{fa-icon scmContext.scmIcon}} {{scmContext.scm}}</span>
+{{#if pipeline.configPipelineId}}
+  {{#link-to "pipeline" pipeline.configPipelineId class="back"}}Back to parent pipeline{{/link-to}}
```
jithin1987: just `Parent Pipeline` is good. From search it will be confusing.

app/components/pipeline-header/styles.scss
```diff
@@ -34,3 +34,7 @@ a.active {
   color: $grey-800;
   text-decoration: underline;
 }
+
+.back {
+  float: right;
```
jithin1987: why is this needed. I in general don't prefer floats, since they takes nodes out of normal document flow

tkyi:
@jithin1987 I think it's weird to have Child Pipelines as the first tab. This is still a pipeline and whenever changes are made, they will still want to see Events first. Will this page default to opening on Events still?

jithin1987:
> I think it's weird to have Child Pipelines as the first tab. This is still a pipeline and whenever changes are made, they will still want to see Events first. Will this page default to opening on Events still?

Purpose of External SD Config is to support building child pipelines, and it's natural for the owner to see pipelines for children and hence it's given more prominent. Config pipeline should default to "child pipelines" view. If owner of the Config pipeline want to see events for the config pipeline user can still click the Events tab.

Also config pipeline shouldn't be forced to have build pipeline of it's own, it's forcing the owner of config repo to implement a build pipeline which might be a duplicate of work user has already done. If user has only provided a `screwdriver.yaml` and not replicated the directory structure/files we need take care of not spamming events  page with failed events.

PR #312, opened 2018-06-18
fix: set coverage query startTime to first step startTime

minz1027:
Cannot use `buildStart` attr directly cuz it's formatted by ember and our API didn't expect that date format.  Change it to be the startTime of the first step instead.
https://github.com/screwdriver-cd/ui/blob/master/app/build/model.js#L53

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -75,7 +75,7 @@ export default Component.extend({
 
   coverageInfoCompute() {
     // Set coverage query startTime to build start time since user can do coverage during user step
-    const buildStartTime = this.get('buildStart');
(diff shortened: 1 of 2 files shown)


app/components/build-banner/component.js
```diff
@@ -75,9 +75,11 @@ export default Component.extend({
 
   coverageInfoCompute() {
     // Set coverage query startTime to build start time since user can do coverage during user step
-    const buildStartTime = this.get('buildStart');
+    const buildStartTime = this.get('buildSteps')[0].startTime;
     const coverageStepEndTime = this.get('coverageStepEndTime');
 
+    console.log('start step:', this.get('buildSteps')[0]);
```
d2lam: remove this

PR #314, opened 2018-06-19
feat(614): Add Pipeline Token UI

kumada626:
## Context
We added pipeline token endpoints to the API ( https://github.com/screwdriver-cd/screwdriver/pull/1123 ).
Then, I add the pipeline token UI to create, delete and refresh the pipeline tokens.

## Objective
Add a token-list to the pipeline's secrets tab.

This creates, delete and refresh pipeline tokens in the same way of User api tokens.
In the app/application/adapter.js, overrides some method to generate pipeline token's endpoints.

## References
issue: https://github.com/screwdriver-cd/screwdriver/issues/614
API: https://github.com/screwdriver-cd/screwdriver/pull/1123


app/application/adapter.js
```diff
@@ -152,5 +153,71 @@ export default DS.RESTAdapter.extend({
 
     // Pass-through to super-class
     return this._super(status, headers, data, requestData);
+  },
+  /**
+   * Overriding default adapter because pipelinetoken's  endpoint is differnt
```
tkyi: Overriding default adapter because the pipeline token's endpoint is different

app/pipeline/secrets/controller.js
```diff
@@ -7,6 +7,10 @@ export default Controller.extend({
   errorMessage: '',
   pipeline: reads('model.pipeline'),
   secrets: reads('model.secrets'),
+  pipelinetokens: reads('model.tokens'),
```
tkyi: camel case => pipelineTokens?

app/application/adapter.js
kumada626: Fixed.

app/pipeline/secrets/controller.js
kumada626: Yes, that should be camel case. I fixed it.

app/application/adapter.js
```diff
@@ -152,5 +153,71 @@ export default DS.RESTAdapter.extend({
 
     // Pass-through to super-class
     return this._super(status, headers, data, requestData);
+  },
+  /**
+   * Overriding default adapter because pipeline token's  endpoint is differnt
```
tk3fftk: (small comment) this line has double space between "pipelinetoken's" and "endpoint".

kumada626: Oh……fixed.

PR #315, opened 2018-06-19
Fix duration for parallel/running workflow

tanoda:
Pipeline page can show longer duration than real with parallel workflow.
I fix this by compute duration using (latest end time) - (earliest create time) for completed event, and (now) - (earliest create time) for running event.

references: [current build.totalDurationMS implementation](https://github.com/screwdriver-cd/ui/blob/4b299ce233419170828edb62e2eb24eba721c262/app/build/model.js#L75-L79)

### DIFF
--- a/app/event/model.js
+++ b/app/event/model.js
@@ -38,7 +38,19 @@ export default DS.Model.extend(ModelReloaderMixin, {
   }),
   duration: computed('builds.[]', 'isComplete', {
     get() {
-      return get(this, 'builds').reduce((val = 0, item) => val + get(item, 'totalDurationMS'));
+      const builds = get(this, 'builds');
+      const firstCreateTime = builds.map(item => get(item, 'createTime')).sort()[0];
+      let lastEndTime = new Date();
+
+      if (get(this, 'isComplete')) {
+        lastEndTime = builds.map(item => get(item, 'endTime')).sort().pop();
+      }
+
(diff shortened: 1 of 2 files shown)


app/event/model.js
```diff
@@ -38,7 +38,15 @@ export default DS.Model.extend(ModelReloaderMixin, {
   }),
   duration: computed('builds.[]', 'isComplete', {
     get() {
-      return get(this, 'builds').reduce((val = 0, item) => val + get(item, 'totalDurationMS'));
+      const builds = get(this, 'builds');
+      const firstCreateTime = builds.map(item => get(item, 'createTime')).sort()[0];
+      const lastEndTime = builds.map(item => get(item, 'endTime')).sort().pop();
+
+      if (!firstCreateTime || !lastEndTime) {
+        return 0;
```
jithin1987: I think `0` will look weird until event turns to `isComplete`.  Can we improve duration displayed for event with current running builds ?

Lets say 2 out of 3 builds have been finished and 3rd build is running, is there enough data to show `startTime/endTime of current running step in build 3` - `startTime of 1st build` (or at least `startTime of build 3` -  `startTime of 1st build`)

tanoda: Thanks for comment. I'll look into it.

tanoda: fixed and added test case.

PR #316, opened 2018-06-19
feat: Add last updated time to templates and commands

tkyi:
## Context

Users would be interested in when a command or template was published. 

## Objective
This PR adds that info to the templates/commands landing and detail pages.

### Templates

screen shot 2018-06-19 at 1 35 36 pm

screen shot 2018-06-19 at 1 35 58 pm

### Commands

screen shot 2018-06-19 at 1 35 12 pm
screen shot 2018-06-19 at 1 35 20 pm

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/522#issuecomment-309585283


app/commands/detail/route.js
```diff
@@ -1,6 +1,8 @@
 import { inject as service } from '@ember/service';
 import RSVP from 'rsvp';
 import Route from '@ember/routing/route';
+import templateHelper from 'screwdriver-ui/utils/template';
```
jithin1987: `import { getLastUpdatedTime } from 'screwdriver-ui/utils/template';` here and elsewhere

app/components/command-header/template.hbs
```diff
@@ -4,6 +4,7 @@
 <p>{{command.description}}</p>
 <ul>
   <li>Released by: <a href="mailto:{{command.maintainer}}">{{command.maintainer}}</a></li>
+  {{#if command.lastUpdated}}<li>Last publish: {{command.lastUpdated}}</li>{{/if}}
```
jithin1987: `published`

app/components/command-section/template.hbs
```diff
@@ -5,4 +5,8 @@
   <p class="description">
     {{command.description}}
   </p>
-</article>
\ No newline at end of file
+  <ul>
+    <li>Released by: <a href="mailto:{{command.maintainer}}">{{command.maintainer}}</a></li>
+    {{#if command.lastUpdated}}<li>Last publish: {{command.lastUpdated}}</li>{{/if}}
```
jithin1987: `published`

app/commands/detail/route.js
```diff
@@ -10,6 +12,13 @@ export default Route.extend({
     ]).then((arr) => {
       const [verPayload] = arr;
 
+      verPayload.forEach((verObj) => {
```
jithin1987: rather than repeating this logic in different places I would rather modify command Service to handle in a central place, same goes for `templates`

app/components/command-header/template.hbs
tkyi: I know npm does "Last Publish" (https://www.npmjs.com/package/screwdriver-coverage-base) but Docker hub does "last pushed" (https://hub.docker.com/r/screwdrivercd/launcher/)
I can changed everything to Last published if that makes more sense to you

app/commands/detail/route.js
tkyi: I'm using the function from templateHelper, per your suggestion last time I tried doing something similar with the template getFullName. Not sure what you want to do instead

app/commands/detail/route.js
tkyi: That doesn't work. Tests fail with something like:
```
not ok 570 Chrome 67.0 - Unit | Utility | template: it gets the name as full name when namespace is default
    ---
        actual: >
            null
        stack: >
            TypeError: (0 , _template.getFullName) is not a function
                at Object.<anonymous> (http://localhost:7357/assets/tests.js:11059:44)
                at runTest (http://localhost:7357/assets/test-support.js:4108:30)
                at Test.run (http://localhost:7357/assets/test-support.js:4094:6)
                at http://localhost:7357/assets/test-support.js:4306:12
                at Object.advance (http://localhost:7357/assets/test-support.js:3746:26)
                at begin (http://localhost:7357/assets/test-support.js:5695:20)
                at http://localhost:7357/assets/test-support.js:4693:6
        message: >
            Died on test #1     at Module.callback (http://localhost:7357/assets/tests.js:11057:19)
                at Module.exports (http://localhost:7357/assets/vendor.js:111:32)
                at requireModule (http://localhost:7357/assets/vendor.js:32:18)
                at TestLoader.<anonymous> (http://localhost:7357/assets/test-support.js:24726:11)
                at TestLoader.require (http://localhost:7357/assets/test-support.js:24716:27)
                at TestLoader.loadModules (http://localhost:7357/assets/test-support.js:24708:16)
                at loadTests (http://localhost:7357/assets/test-support.js:25387:22): (0 , _template.getFullName) is not a function
        Log: |
    ...

```

app/commands/detail/route.js
jithin1987: index route calls `getAllCommands` and detail route calls `getOneCommand` from Command service, both of which calls `fetchData` https://github.com/screwdriver-cd/ui/blob/master/app/command/service.js#L14 which finally calls API to get command data https://github.com/screwdriver-cd/ui/blob/master/app/command/service.js#L33 
In both cases an array is returned, so you can move this logic of calling `getLastUpdatedTime` inside the `.done` handler in Service at one place rather than doing it twice in each route.

jithin1987:
@tkyi this is what I am thinking of https://github.com/screwdriver-cd/ui/commit/47b608dc2b87b74bfb4251bf4f16b2cc3e78c562#diff-3c0db6d46f7d78df6da38560f3950c26

Those changes are for template.

tkyi:
@jithin1987 That looks like a namespace change, did you mean to send that to Brian?

I've updated my PR with changes you requested in the `.done` section of `fetchData`

PR #319, opened 2018-06-21
feat(1040): add supported images to template view

jimgrund:
## Context

Display supported images and their aliases in the template view when a template includes multiple images.  

image

## References

* Issue: https://github.com/screwdriver-cd/screwdriver/issues/1040


app/components/template-header/template.hbs
```diff
@@ -9,8 +9,17 @@
 </h1>
 <h2>{{template.version}}</h2>
 <p>{{template.description}}</p>
-<ul>
+<ul id="template-details">
```
tkyi: We don't usually use `id`, can you switch to `class`

tkyi:
Can you post some screenshots?
Also need to make changes to the validator to show the images.

app/components/template-header/template.hbs
jithin1987: I would have a structure like this. There is no need to force fit an `ul` here.
```
<div class="template-details">
  {{#if template.namespace}}
  <div class="template-details--item">
    <span>Namespace:</span><span>{{template.namespace}}</span>
  </div>
  {{/if}}
  <div class="template-details--item">
    <span>Namespace:</span><span>{{template.name}}</span>
  </div>
  {{#if template.images}}
  <div class="template-details--item">
    <span>Supported Images:</span>
    <div>
      {{#each-in template.images as |alias path|}}
      <div>
        <span>{{alias}}: </span><span>{{path}}</span>
      </div>
      {{/each}}
    </div>
  </div>
  {{/if}}
</div>
```

With css

```
.template-details--item > div {
 display: inline-block;
}
```

app/components/template-header/styles.scss
```diff
@@ -1,16 +1,29 @@
...
-
   .template-label {
     margin: 0 5px;
     padding: 3px 5px;
     border-radius: 3px;
     background-color: $sd-light-blue;
     color: $sd-white;
   }
+
+  .back {
```
jithin1987: think a rebase issue, this was removed originally.

app/components/template-header/styles.scss
```diff
@@ -1,16 +1,29 @@
...
+    float: right;
+  }
+
+  .template-details--item > div {
+    margin-left: 10px;
```
jithin1987: can you remove this margin ? It should make the details item align to left along with description,

app/components/template-header/styles.scss
```diff
@@ -1,16 +1,25 @@
...
   .template-label {
     margin: 0 5px;
     padding: 3px 5px;
     border-radius: 3px;
     background-color: $sd-light-blue;
     color: $sd-white;
   }
+
+  .template-details--item > div {
+    margin-left: 0px;
```
jithin1987: Is there already a `margin-left` applied from some other rule ? if not you can just remove this rule. Also with `0` unit need not be mentioned - `margin-left: 0` If it can be omitted, then it's ideal

PR #328, opened 2018-07-02
fix(1025): Display error messages on pipeline options page

tkyi:
## Context
We're currently not displayed error messages on the UI whenever people do not have permissions to perform actions on the pipeline options page.

## Objective
This PR bubbles up the error message so it's clear that the user's attempt was invalid.

Trying to toggle the jobs:
screen shot 2018-07-02 at 3 03 03 pm

Trying to sync:
screen shot 2018-07-02 at 3 03 29 pm

Trying to delete the pipeline:
screen shot 2018-07-02 at 3 03 17 pm

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1025

### DIFF
--- a/app/pipeline/options/controller.js
+++ b/app/pipeline/options/controller.js
@@ -13,12 +13,14 @@ export default Controller.extend({
       const job = this.store.peekRecord('job', id);
 
       job.set('state', state);
-      job.save().catch(error => this.set('errorMessage', error.errors[0].detail || {}));
+      job.save()
+        .catch(error => this.set('errorMessage', error.errors[0].detail || ''));
     },
     removePipeline() {
       this.get('pipeline').destroyRecord().then(() => {
         this.transitionToRoute('home');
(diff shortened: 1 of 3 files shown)


app/pipeline/options/controller.js
```diff
@@ -13,12 +13,12 @@ export default Controller.extend({
       const job = this.store.peekRecord('job', id);
 
       job.set('state', state);
-      job.save();
+      job.save().catch(error => this.set('errorMessage', error.errors[0].detail || {}));
```
jithin1987: shouldn't default be an empty string ? `this.set('errorMessage', error.errors[0].detail || '')`

Assuming `errors[0]` is guaranteed to be an object on error. Same for other two changes as well.

jithin1987:
While we are at this pipeline options page can you take a look at this as well ? https://github.com/screwdriver-cd/screwdriver/issues/765

app/pipeline/options/controller.js
tkyi: updated

PR #333, opened 2018-07-12
feat(1040): update validator and template display

jimgrund:
## Context

Relocate supported images from template-header to the content section of the template display page.
Update validator to display supported images.

image

image

## References

* Issue: https://github.com/screwdriver-cd/screwdriver/issues/1040


app/components/validator-job/template.hbs
```diff
@@ -1,30 +1,44 @@
 {{#if collapsible}}
 <h4 class="job" {{action 'nameClick'}}>{{fa-icon (if isOpen "minus-square" "plus-square")}} {{name}}{{#if index }}.{{index}}{{/if}}</h4>
 {{/if}}
-<div class="annotations" title="These are the job-level annotations that the user has defined.">
-  <span class="label">Annotations:</span>
+{{#if template.description}}
```
tkyi: I don't think we need to display this under Contents since it's already shown at the top of the page

tkyi:
Do you have a screenshot with `Settings`?

jimgrund:
Screenshot updated to include full template view including Settings

app/components/validator-job/template.hbs
jimgrund: Are you referring to the `label` displaying as "Template Description"?  This was a bit of confusion for me as it's valid for someone to have a `description` in the config section of the template (maybe unlikely) as well as overall for the template.  This would've had two Descriptions display and without additional layout changes to differentiate the `config` elements it would've been confusing to end-user.

tkyi: Hm I never noticed that
Strange. Not sure why you would need a description per job

tkyi: I think that's something we could remove (from the data-schema) in the future, but for this pass your PR looks fine. 👍

PR #334, opened 2018-07-16
fix commands with same namespace duplicating

lusol:
Fix bug where when you filter commands by namespace, commands with the same name will show up multiple times (due to them having different versions).


app/command/service.js
```diff
@@ -57,7 +57,19 @@ export default Service.extend({
             }
           });
 
-          return resolve(commands);
+          let unique = {};
+
+          let uniqueCommands = commands.filter((c) => {
+            if (c.name in unique) {
```
tkyi: shouldn't filter by only name, needs to be `c.namespace`/`c.name`

I think this is much simpler to read
```
const names = {};
  const result = [];

  commands.forEach((c) => {
    const fullName = `${c.namespace}/${c.name}`;

    if (!names[fullName]) {
      names[fullName] = 1;
      result.push(c);
    }
  });

  return Promise.resolve(result);
```

app/command/service.js
```diff
@@ -57,7 +57,21 @@ export default Service.extend({
             }
           });
 
-          return resolve(commands);
+          let unique = {};
+
+          let uniqueCommands = commands.filter((c) => {
+            let fullname = `${c.namespace}/${c.name}`;
```
tkyi: should be fullName for consistency

tk3fftk:
Related: https://github.com/screwdriver-cd/screwdriver/issues/1096

app/command/service.js
```diff
@@ -57,7 +57,21 @@ export default Service.extend({
             }
           });
 
-          return resolve(commands);
+          let unique = {};
```
jithin1987: putting this logic in `fetchData()` will affect output of `getCommandTags` &  `getOneCommand` as well. This will affect the Command detail page which needs all versions of a given Command to display `versions`  info at bottom of page. You could move this logic to a `then` handler of `getAllCommands` like how Templates are doing https://github.com/screwdriver-cd/ui/blob/master/app/template/service.js#L36

lusol: @jithin1987 DONE DUDE

PR #335, opened 2018-07-19
fix: fix event duration

minz1027:
### Context
Currently, if you have multiple workflow inside the pipeline. The event duration is wrong even only partial flow is triggered. Because it's using number of jobs inside the workflowGraph to determine if an event finishes or not. 

### Solution
There is no good way to determine if an event is done with sourcePaths unless we store sourcePaths info in the event itself. Instead of checking number of builds, in this PR, If no new builds added in 2 reloads(each reload has 1 min delay), we consider the event as complete.

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1173

### DIFF
--- a/app/event/model.js
+++ b/app/event/model.js
@@ -1,11 +1,9 @@
 import { computed, observer, get, set } from '@ember/object';
 import { sort, not } from '@ember/object/computed';
 import DS from 'ember-data';
-import graphTools from 'screwdriver-ui/utils/graph-tools';
 import ENV from 'screwdriver-ui/config/environment';
 import ModelReloaderMixin from 'screwdriver-ui/mixins/model-reloader';
-
-const { graphDepth } = graphTools;
+import { isActiveBuild } from 'screwdriver-ui/utils/build';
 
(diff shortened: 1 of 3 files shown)


app/event/model.js
```diff
@@ -101,49 +102,47 @@ export default DS.Model.extend(ModelReloaderMixin, {
...
         if (runningBuild) {
           set(this, 'isComplete', false);
+          set(this, 'numBuilds', numBuilds);
+          set(this, 'reloadWithoutNewBuilds', 0);
+
+          return false;
+        }
+
+        // Nothing is running now, check if new builds added during reload
+        const newBuilds = get(this, 'numBuilds') === 0 ? 0 : numBuilds - get(this, 'numBuilds');
```
jithin1987: can't this be `numBuilds - get(this, 'numBuilds')` ?

minz1027: That is to avoid one more reload to determine if event is complete. When the first time it loads, get(this, 'numBuilds') has a default value of `0`, so it's always considered there are new builds. But not true for completed event.

PR #337, opened 2018-07-23
Draggable Pipeline Graph in Job Page

lusol:
When on a job page, users are able to view the entire pipeline which the job belongs to. The user can navigate between the different jobs on the pipeline by clicking on individual nodes on the draggable graph without having to go back to the pipeline page. The current job node which the user is on will always be highlighted.

screen shot 2018-08-02 at 10 40 29 am

at bottom left

Related to https://github.com/screwdriver-cd/screwdriver/issues/958


app/components/workflow-graph-d3/component.js
```diff
@@ -86,6 +96,8 @@ export default Component.extend({
     });
   },
   draw() {
+    console.log('test')
```
tkyi: console.log?

tkyi:
Is this PR ready for review? Can you update the description and add screenshots/a demo?

tkyi:
This is not the UX that the mock showed. I think this PR still needs some work.

app/pipeline/build/route.js
```diff
@@ -16,7 +16,8 @@ export default Route.extend({
       job,
       event,
       pipeline: this.get('pipeline'),
-      jobs: jobs.filter(j => !/^PR-/.test(j.get('name')))
+      jobs: jobs.filter(j => !/^PR-/.test(j.get('name'))),
+      events: this.get('pipeline.events')
```
jithin1987: this will change with @tkyi pagination PR, since we pushed events down to `events` route. But lets push this out with existing scenario and worry about this later.

app/components/workflow-graph-d3/component.js
```diff
@@ -131,16 +142,48 @@ export default Component.extend({
       // create the icon graphic
       .insert('text')
       .text(d => icon(d.status))
-      .attr('font-size', `${ICON_SIZE}px`)
+      .attr('font-size', (e) => {
+        let routeParams = this.get('routeParams');
+        let buildId = this.get('build.id');
+
+        if (get(this, 'dragsize')
```
jithin1987:
```
       let routeParams = this.get('routeParams');
        let buildId = this.get('build.id');
        if (get(this, 'dragsize')
          && ((routeParams && routeParams.build_id && routeParams.build_id === e.buildId)
            || (buildId === e.buildId))) {
```

Make this a helper function `isDragGraphActiveNode(this, e)`, this logic is being repeated multiple places.

PR #341, opened 2018-08-03
fix(1194): fix guest mode

jithin1987:
Clear out collection data when switching between guest and user login


app/authenticators/screwdriver-api.js
```diff
@@ -62,15 +78,9 @@ export default Base.extend({
           return reject();
         }
 
-        const { username, scope, scmContext } = jwt;
+        const authData = getData(token, jwt);
```
tkyi: These var names are kind of confusing for me, since the jsdocs for this function say the first argument is the JWT and the second one is the decoded token

I think it would be clearer to me if you either:
- take in args as a config to make it more explicit/clear
- rename the variables that are passed in here

tests/helpers/inject-scm.js
```diff
@@ -18,9 +18,10 @@ const scms = [
 
 /**
  * Inject scm service to an Ember Object
- * @param {Object} self - Ember object generated by ember-qunit moduleFor()
+ * @param {Object}    self    - Ember object generated by ember-qunit moduleFor()
+ * @param {oldFormat} boolean - Denotes if test is following old RFC
```
tkyi: JSdocs should be `{Boolean} oldFormat`

app/authenticators/screwdriver-api.js
jithin1987: done

tests/helpers/inject-scm.js
jithin1987: fixed

lusol:
LGTM

PR #346, opened 2018-08-13
feat(1204): Use pagination for pipeline events [4]

tkyi:
## Context
Should call newly updated events endpoint each time we display more events on the pipeline page.

## Objective
This PR calls the events endpoint with pagination to make the page load a lot faster.

_Note: Not sure if we need to keep the UI backwards-compatible with the old API (with no pagination)._

## Related links
~Blocked by: https://github.com/screwdriver-cd/screwdriver/pull/1248~
Related to https://github.com/screwdriver-cd/screwdriver/issues/1204


app/pipeline/events/route.js
```diff
@@ -1,20 +1,28 @@
...
 export default Route.extend({
   routeAfterAuthentication: 'pipeline.events',
   beforeModel() {
     this.set('pipeline', this.modelFor('pipeline').pipeline);
   },
   model() {
-    return this.get('pipeline.jobs')
-      // Split jobs from workflow
-      .then(jobs => ({
+    return RSVP.all([
```
jithin1987: this breaks the `refresh` semantics used by `ModelReloaderMixin` (relays on `model.refresh`) which results in not loading PR's or events automatically without page reload.

app/pipeline/events/controller.js
```diff
@@ -132,6 +126,24 @@ export default Controller.extend(ModelReloaderMixin, {
...
+      })
+        .then((events) => {
+          const nextEvents = events.toArray();
+
+          if (Array.isArray(nextEvents)) {
+            if (nextEvents.length < ENV.APP.NUM_EVENTS_LISTED) {
+              this.set('moreToShow', false);
+            }
+
+            this.set('model.events', this.get('events').concat(nextEvents));
```
jithin1987: Once we fix the reload semantics of pipeline events, this will be problematic. I think we should collect paginated events in another field and then use `computed` property to combine initial events given via model and new events loaded through pagination.

```
  this.set('paginatedEvents', this.get('paginatedEvents').concat(nextEvents));
...
events: computed('initialEvents', 'paginatedEvents' 
...
```

app/pipeline/events/route.js
jithin1987: I was trying out this https://stackoverflow.com/a/35935554 but getting errors, need to dig deeper

app/pipeline/events/controller.js
```diff
@@ -13,15 +12,39 @@ export default Controller.extend(ModelReloaderMixin, {
...
+    get() {
+      const jobs = this.get('model.jobs');
 
+      return jobs.filter(j => !/^PR-/.test(j.get('name')));
+    }
+  }),
+  paginatedEvents: [],
+  initialEvents: computed('model.events', {
+    get() {
+      return this.get('model.events').toArray();
```
jithin1987: This will hold first page results, need to have dedup logic

app/pipeline/events/controller.js
```diff
@@ -1,31 +1,63 @@
...
+    return Promise.resolve();
+  },
   isShowingModal: false,
+  moreToShow: true,
   errorMessage: '',
-  pipeline: reads('model.pipeline'),
-  jobs: reads('model.jobs'),
-  events: reads('model.events'),
-  pullRequests: reads('model.pullRequests'),
+  jobs: computed('model.jobs', {
```
tkyi: do you need this?

app/pipeline/events/controller.js
```diff
@@ -1,31 +1,63 @@
...
+      return jobs.filter(j => !/^PR-/.test(j.get('name')));
+    }
+  }),
+  paginateEvents: [],
+  initialEvents: computed('model.events', {
+    get() {
+      let headEvents = this.get('headEvents') || [];
+      let modelEvents = this.get('model.events').toArray();
+
+      headEvents = headEvents.filter(e => !modelEvents.find(c => c.id === e.id));
```
tkyi: maybe add a comment here

tkyi:
I think what's confusing for me is all the events-related vars: `modelEvents`, `initialEvents`, `headEvents`, `nextEvents`, `paginateEvents`, `events`.

- `modelEvents` is the first page of events (as returned by the model)
- `initialEvents` comprises of the original model events and the newly reloaded model events, deduped.
- `headEvents` is the total first group of events in the first page (or a tracker for the last `initialEvents` before reload).
- `nextEvents` is the next page of events.
- `paginateEvents` is the events to show after the first page (based on the user clicking `more`), which consists of all the `nextEvents` concatted together.
- `events` is the total events to show (`initialEvents` + `paginateEvents`).

Not sure what would be a better naming convention, but just writing it down here to help me understand.

tkyi:
Can't approve my own PR but @jithin1987 your changes look good to me 👍

app/pipeline/events/controller.js
```diff
@@ -1,31 +1,63 @@
...
-  events: reads('model.events'),
-  pullRequests: reads('model.pullRequests'),
+  jobs: computed('model.jobs', {
+    get() {
+      const jobs = this.get('model.jobs');
+
+      return jobs.filter(j => !/^PR-/.test(j.get('name')));
+    }
+  }),
+  paginateEvents: [],
```
minz1027: I guess these are the paginated events starting from page 2?

app/pipeline/events/controller.js
```diff
@@ -1,31 +1,63 @@
...
+  nonPaginateEvents: computed('model.events', {
```
minz1027: These are actually the adjusted results for page1, adjusted if events in first page get changed? The naming of these variables is kinda confusing

PR #348, opened 2018-08-16
fix: sleep before checks

d2lam:
Builds failed intermittently because it's not available immediately after deploying. Need to sleep a bit before check


d2lam:
@jithin1987  We did. The builds always failed and we always have to rerun it. You can check the pattern here: https://cd.screwdriver.cd/pipelines/7/events

PR #362, opened 2018-09-24
feat(1283): custom doc link

DekusDenial:
## Context
Customize the doc link

## Objective
Have UI overriding the default doc link with whatever from env var if there is such

## Related links
Related to:
- https://github.com/screwdriver-cd/screwdriver/issues/1283


config/environment.js
```diff
@@ -49,6 +49,7 @@ module.exports = (environment) => {
       SDAPI_NAMESPACE: 'v4',
       SDSTORE_HOSTNAME: 'http://localhost:8081',
       SDSTORE_NAMESPACE: 'v1',
+      SDDOC_URL: 'http://docs.screwdriver.cd',
```
tkyi: Maybe also have it configurable under prod environment (https://github.com/screwdriver-cd/ui/blob/master/config/environment.js#L94-L98)? Not sure what value that would bring though

nginx.conf
```diff
@@ -63,7 +64,7 @@ http {
         # Dynamic configuration for the ember app
         location /assets/supplementary_config.js {
            content_by_lua_block {
-               ngx.say("window.SUPPLEMENTARY_CONFIG = { SDAPI_HOSTNAME: '", os.getenv("ECOSYSTEM_API"), "', SDSTORE_HOSTNAME: '", os.getenv("ECOSYSTEM_STORE"), "'  };")
+               ngx.say("window.SUPPLEMENTARY_CONFIG = { SDAPI_HOSTNAME: '", os.getenv("ECOSYSTEM_API"), "', SDSTORE_HOSTNAME: '", os.getenv("ECOSYSTEM_STORE"), "', SDDOC_URL: '", os.getenv("ECOSYSTEM_STORE"), "'  };")
```
tkyi: should it be `os.getenv("SDDOC_URL")`?

DekusDenial: oh well

config/environment.js
DekusDenial: actually, i was considering `http://docs.screwdriver.cd` as default since the cluster means to override it

PR #364, opened 2018-09-27
fix(1241): Check for branch-specific trigger for D3 graph

tkyi:
## Context
Need to highlight the branch-specific trigger when it triggers a build.

## Objective
This PR checks if the branch-specific node matches the startFrom and sets the status and icons appropriately.

screen shot 2018-09-27 at 11 56 47 am

It uses the same logic found here: https://github.com/screwdriver-cd/workflow-parser/blob/master/lib/getNextJobs.js#L24-L39

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1241


app/utils/graph-tools.js
```diff
@@ -104,6 +104,34 @@ const walkGraph = (graph, start, x, y) => {
...
+ * @return {Boolean}       True if the node matches the startFrom
+ */
+const isTrigger = (name, start) => {
+  // Set a status on the trigger node (if it starts with ~)
+  if (name === start && /^~/.test(name)) {
+    return true;
+  }
+
+  // Set status on trigger node if is branch specific trigger
+  const edgeSrcBranch = name.match(new RegExp('^~(pr|commit):/(.+)/$'));
```
jithin1987: can you take this out and define this regex as a constant outside of this function 
Eg: `const TRIGGER_REGEX = new RegExp('^~(pr|commit):/(.+)/$')`

PR #372, opened 2018-10-18
fix: replace rootUrl during runtime

minz1027:
rootURL is compiled during build time for index.html.
to override that during runtime, we can use nginx sub_filter to replace it.

http://nginx.org/en/docs/http/ngx_http_sub_module.html

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -56,8 +56,11 @@ http {
         add_header x-xss-protection '1; mode=block';
         add_header Content-Security-Policy $csp_header;
 
+        set_by_lua $rootURL 'return os.getenv("ROOT_URL") or ""';
 
         location / {
+            sub_filter "/assets" "$rootURL/assets";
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -56,8 +56,11 @@ http {
         add_header x-xss-protection '1; mode=block';
         add_header Content-Security-Policy $csp_header;
 
+        set_by_lua $rootURL 'return os.getenv("ROOT_URL") or ""';
```
tkyi: What version are we using? According to the set_by_lua docs,
>NOTE Use of this directive is discouraged following the v0.9.17 release. Use the set_by_lua_block directive instead.

should we be using `set_by_lua_block` instead?
https://github.com/openresty/lua-nginx-module#set_by_lua_block

minz1027: Sure, i can change it to  set_by_lua_block

PR #378, opened 2018-12-07
feat: Add cache invalidation to pipeline options page

tkyi:
## Context
It would be nice if users could invalidate their pipeline or job caches through the UI.

## Objective
This PR adds functionality for clearing job or pipeline cache.

screencapture-localhost-4200-pipelines-1-options-2018-12-06-16_18_20

_Note: not sure how/if we need to add event cache invalidation._

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1385

### DIFF
--- a/app/cache/service.js
+++ b/app/cache/service.js
@@ -1,0 +1,53 @@
+import $ from 'jquery';
+import { Promise as EmberPromise } from 'rsvp';
+import Service, { inject as service } from '@ember/service';
+import { get } from '@ember/object';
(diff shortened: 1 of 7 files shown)


DekusDenial:
So we need to check for admin permission or it’s automatically implied already on that page?

tkyi:
@DekusDenial When the request is sent, it will use the user's credentials and should return 401 if they're not authorized.

PR #380, opened 2018-12-15
fix(1335): Allow user to trigger any job

tkyi:
## Context
Users should be able to start from any job.

## Objective
This PR fixes a bug that set all jobs to `isRootNode = false` and allows users to start any job as long as it is not root.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1335

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -61,14 +61,15 @@ export default Component.extend({
     graphClicked(job, mouseevent, sizes) {
       const edges = get(this, 'directedGraph.edges');
       let isRootNode = true;
+      let isTrigger = job ? /^~/.test(job.name) : false;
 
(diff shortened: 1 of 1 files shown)


app/components/pipeline-workflow/component.js
```diff
@@ -60,15 +60,15 @@ export default Component.extend({
   actions: {
     graphClicked(job, mouseevent, sizes) {
       const edges = get(this, 'directedGraph.edges');
-      let isRootNode = false;
+      let isRootNode = true;
```
minz1027: don't thinks this will work. cuz isRoot will set it to back to false if it is not root. We should probably just remove this whole rootNode check logic

tkyi: hm still need `isRoot` to figure out where to place the toolTip, but can figure out if it's the first element by matching for `~`.

PR #388, opened 2019-01-16
fix: fix init logs when init failed

minz1027:
change init logs for failed init step

### DIFF
--- a/app/components/build-log/component.js
+++ b/app/components/build-log/component.js
@@ -86,9 +86,16 @@ export default Component.extend({
           }
 
           if (get(this, 'stepEndTime')) {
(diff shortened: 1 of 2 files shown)


app/components/build-log/component.js
```diff
@@ -86,9 +86,16 @@ export default Component.extend({
           }
 
           if (get(this, 'stepEndTime')) {
+            let msg = 'Image pull completed. Build init completed';
```
tkyi: maybe add another period at the end just for consistency? but not necessary

minz1027: Yeah, i missed it!

PR #390, opened 2019-01-21
feat(pipeline detail): UI updates

DekusDenial:
## Context
UI updates for pipeline events and PRs view

## Objective
see linked issues below for mocks/screenshots

**also got rid of some deprecation warnings from deps*

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1311
Related to https://github.com/screwdriver-cd/screwdriver/issues/1419


app/styles/column-view.scss
```diff
@@ -0,0 +1,67 @@
...
+
+a {
+  color: $brand-600;
+}
+
+.view {
+  display: flex;
+  padding-bottom: 10px;
+  border-bottom: 1px solid $grey-400;
+  // justify-content: space-between;
```
jithin1987: please remove if not required?

app/styles/column-view.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.status {
+  font-size: 2rem;
+  padding-right: 5px;
+  color: $sd-failure;
```
jithin1987: why `sd-failure` color for status ? Perhaps overridden elsewhere ?

app/utils/graph-tools.js
```diff
@@ -235,4 +239,38 @@ const decorateGraph = (inputGraph, builds, start) => {
...
 
-export default { node, icon, decorateGraph, graphDepth, isRoot, isTrigger };
+/**
+ * Filter to the subgraph in which the root is the start from node
+ * @param   {Array}   [{nodes}]   Array of graph vertices
+ * @param   {Array}   [{edges}]   Array of graph edges
+ * @param   {String}  [startNode] Starting/trigger node
+ * @returns {Object}              Nodes and edges for the filtered subgraph
+ */
+const subgraphFilter = ({ nodes, edges }, startNode) => {
```
jithin1987: is this for the list view ? If so then perhaps on a second pass, we can take into account status of the builds in the path as well - say if build didn't run then don't show it. But this is a good start.

app/styles/column-view.scss
DekusDenial: yup

app/utils/graph-tools.js
DekusDenial: yeah, this is supposed to be before the status got deduced. Good start indeed

PR #398, opened 2019-02-01
fix(1460): Add build blocked info in init step [3]

tkyi:
Adding `blockedStartTime` to build stats to track gap of time between time when build is enqueued and when it starts running.

Note: Need to merge changes in /screwdriver pipeline to prod first.

Related to https://github.com/screwdriver-cd/screwdriver/issues/1460

### DIFF
--- a/app/components/build-log/component.js
+++ b/app/components/build-log/component.js
@@ -77,6 +77,14 @@ export default Component.extend({
             n: 1
           });
 
+          if (buildStats.blockedStartTime) {
+            initLogs.push({
+              t: new Date(buildStats.blockedStartTime).getTime(),
+              m: 'Build blocked, putting back into queue.',
+              n: 1
+            });
+          }
+
           if (buildStats.hostname) {
             initLogs.push({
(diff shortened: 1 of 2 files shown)


app/components/build-log/component.js
```diff
@@ -77,6 +77,14 @@ export default Component.extend({
             n: 1
           });
 
+          if (buildStats.blockedStartTime) {
+            initLogs.push({
+              t: new Date(buildStats.blockedStartTime).getTime(),
+              m: 'Build blocked, putting back into queue.',
```
jithin1987: can we get information on the blocked build here ? `Build blocked by build LINK_TO_BUILD_X`

tkyi: would need to add another field to `build.stats` like `blockedByIds` or something and call models function `build.getBlockedByIds` in API to set it (https://github.com/screwdriver-cd/models/blob/master/lib/build.js#L37)

tkyi: can work on that in second pass

PR #399, opened 2019-02-01
fix: event commit msg & workflow node ordering

DekusDenial:
* fix: truncate commit message
* fix: workflow graph node ordering


catto:
I think this PR fixes 
https://github.com/screwdriver-cd/screwdriver/issues/1359

catto:
@jithin1987 Sure! I'll check it when I'm available

PR #400, opened 2019-02-01
fix: Show breakdown of total build time

tkyi:
## Context
It can be confusing for users to see the Duration text in the banner since they might think the build took all that time, when actually the Duration is made up of queue time, blocked by time, image pull time, and build time.

## Objective
This PR adds a title when hovering over build duration banner and expands to show a slight build duration breakdown when clicked. Clicking again hides the breakdown.

_Note: This PR might not be a good idea since we still don't have a good way to break down image pull time vs blocked by time. Not sure if it would be better to wait until we have those times before adding any info to the banner._

Hover example with title:
screen shot 2019-02-01 at 2 13 33 pm

Expansion example:
screen shot 2019-02-05 at 2 06 44 pm

This PR also fixes so it will not show undefined for build status on hover over a build in the workflow graph (e.g.: `main - undefined`). It will just show job name until status is generated.

### DIFF
--- a/app/build/model.js
+++ b/app/build/model.js
@@ -75,7 +75,7 @@ export default DS.Model.extend({
   }),
   imagePullDuration: computed('stats.imagePullStartTime', 'startTime', {
     get() {
-      return durationText.call(this, 'stats.imagePullStartTime', 'stats.startTime');
+      return durationText.call(this, 'stats.imagePullStartTime', 'startTime');
     }
   }),
   truncatedMessage: computed('commit.message', {
--- a/app/build/model.js
+++ b/app/build/model.js
@@ -75,14 +75,6 @@ export default DS.Model.extend({
       return durationText.call(this, 'stats.imagePullStartTime', 'startTime');
     }
   }),
-  truncatedMessage: computed('commit.message', {
-    get() {
-      const msg = this.get('commit.message');
-      const cutOff = 150;
-
-      return msg.length > cutOff ? `${msg.substring(0, cutOff)}...` : msg;
-    }
-  }),
   buildDuration: computed('startTime', 'endTime', {
     get() {
       return durationText.call(this, 'startTime', 'endTime');
--- a/app/build/model.js
+++ b/app/build/model.js
@@ -63,16 +63,13 @@ export default DS.Model.extend({
       return `${dt} ago`;
     }
   }),
-  queuedDuration: computed('createTime', 'stats.blockedStartTime', {
+  // Queue time and blocked time are merged into blockedDuration
+  blockedDuration: computed('createTime', 'stats.imagePullStartTime', {
     get() {
-      return durationText.call(this, 'createTime', 'stats.blockedStartTime');
+      return durationText.call(this, 'createTime', 'stats.imagePullStartTime');
     }
   }),
-  blockedDuration: computed('stats.{blockedStartTime,imagePullStartTime}', {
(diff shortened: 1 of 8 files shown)


jithin1987:
Can we modify it to display  "Blocked time, Setup time (pretty much the image pull duration) & Build time" ? 

Case 1 - No blocked time
"Setup time: 1 minutes"
"Build time: 10 minutes"

Case 1 - With blocked time
"Blocked time: 5 minutes"
"Setup time: 1 minutes"
"Build time: 10 minutes"

jithin1987:
Another aspect to ponder is should we even display the total build time (queue + blocked + execution)
ie under Duration display the build time alone and then on hover + click show the queue & blocked time.

app/build/model.js
```diff
@@ -63,12 +63,35 @@ export default DS.Model.extend({
...
   }),
-  queuedDuration: computed('createTime', 'startTime', {
+  queuedDuration: computed('createTime', 'stats.blockedStartTime', {
     get() {
-      return durationText.call(this, 'createTime', 'startTime');
+      return durationText.call(this, 'createTime', 'stats.blockedStartTime');
     }
   }),
-  buildDuration: computed('createTime', 'endTime', {
+  blockedDuration: computed('stats.{blockedStartTime,imagePullStartTime}', {
```
jithin1987: how does older builds which does not have this info show up ?

jithin1987: My suggestion is that https://github.com/screwdriver-cd/ui/blob/097fe7d2cccbb620f646a3ac05f6b4d0f9487716/app/build/model.js#L35-L38 should return `N/A` if we don't have data

app/components/build-banner/template.hbs
```diff
@@ -33,7 +33,13 @@
     </li>
   {{/if}}
   <li class="duration">
-    <span class="banner-value">{{duration}}</span>
+    <details>
+      <summary><a><span class="banner-value" title="Total duration: {{duration}}, Queue time: {{queueDuration}}, Blocked time: {{blockDuration}}, Image pull time: {{imagePullDuration}}, Build time: {{buildDuration}}">{{duration}}</span></a></summary>
+        <li class="subsection"><span class="banner-value">{{queueDuration}} in queue</span></li>
+        <li class="subsection"><span class="banner-value">{{blockDuration}} blocked</span></li>
```
jithin1987: should be displayed only if build has  `blockDuration` ie it entered blockedBy state.

app/components/build-banner/template.hbs
```diff
@@ -33,7 +33,13 @@
     </li>
   {{/if}}
   <li class="duration">
-    <span class="banner-value">{{duration}}</span>
+    <details>
+      <summary><a><span class="banner-value" title="Total duration: {{duration}}, Queue time: {{queueDuration}}, Blocked time: {{blockDuration}}, Image pull time: {{imagePullDuration}}, Build time: {{buildDuration}}">{{duration}}</span></a></summary>
+        <li class="subsection"><span class="banner-value">{{queueDuration}} in queue</span></li>
```
jithin1987: this should be negligible, we can skip this for now.

tkyi: updated

app/components/build-banner/template.hbs
tkyi: addressed

PR #401, opened 2019-02-06
fix: Check if value exists before calling getTime

tkyi:
Build page is failing to load with error:
```
Uncaught TypeError: n.getTime is not a function
```

This PR should check for element before calling getTime() on it.

### DIFF
--- a/app/build/model.js
+++ b/app/build/model.js
@@ -16,6 +16,10 @@ function calcDuration(start, end) {
     endTime = this.get(end);
   }
 
+  if (typeof endTime === 'string') {
+    endTime = new Date(end);
+  }
+
   startTime = this.get(start);
 
   if (!startTime || !endTime) {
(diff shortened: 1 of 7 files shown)


app/build/model.js
```diff
@@ -16,6 +16,10 @@ function calcDuration(start, end) {
     endTime = this.get(end);
   }
 
+  if (typeof endTime === 'string') {
+    endTime = new Date(end);
+  }
+
   startTime = this.get(start);
```
jithin1987: above change needs to be done for `startTime` as well, I see the same issue locally this time against start time from `imagePullDuration` call

app/build/model.js
```diff
@@ -16,6 +16,14 @@ function calcDuration(start, end) {
     endTime = this.get(end);
   }
 
+  if (typeof startTime === 'string') {
```
jithin1987: this needs to be moved after line 27, which overwrites this value here.

PR #407, opened 2019-02-13
fix: get most recent running event

DekusDenial:
fix: https://github.com/screwdriver-cd/screwdriver/issues/1475


DekusDenial:
not at all, cuz the fallback here is the original logic

PR #410, opened 2019-02-20
fix: add side icons for FROZEN and COLLAPSED

minz1027:
the icon for `FROZEN` is not exactly the same as what we have in the graph cuz that one is from icomoon, not font awesome

https://fontawesome.com/icons/snowflake?style=solid

screen shot 2019-02-20 at 3 02 19 pm
screen shot 2019-02-20 at 3 02 02 pm

### DIFF
--- a/app/utils/build.js
+++ b/app/utils/build.js
@@ -29,6 +29,13 @@ const statusIcon = (status, isLight) => {
   case 'UNSTABLE':
     icon = 'exclamation-circle';
     break;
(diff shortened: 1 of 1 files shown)


jithin1987:
we can update icomoon one to use the icon from front awesome ?

minz1027:
@jithin1987  the font awesome one is not in a circle, it will look weird in a graph

PR #412, opened 2019-02-27
fix(1452): display tests results on build page

minz1027:
- display tests results on build page
- override the sonar coverage / tests results if set in meta

screen shot 2019-02-27 at 2 58 53 pm

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1452

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -8,6 +8,7 @@ export default Component.extend({
   classNames: ['build-banner', 'row'],
   classNameBindings: ['buildStatus'],
   coverage: service(),
+  coverageInfo: {},
   coverageStep: computed('buildSteps', {
     get() {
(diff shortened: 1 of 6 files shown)


app/components/build-banner/component.js
```diff
@@ -71,15 +72,38 @@ export default Component.extend({
     }
   }),
 
+  overrideCoverageInfo() {
+    const buildMeta = this.get('buildMeta');
+
+    // override coverage info if set in build meta
+    if (buildMeta && buildMeta.tests) {
```
pranavrc: Probably nitpicking, but I'd just use `hoek.reach` over the nested if statements. LGTM otherwise

minz1027: lol, i wanted to use that too, but it's not part of the dependencies now. don't want to introduce it only because of this change

PR #413, opened 2019-02-28
fix(1452): add validation for coverage meta

minz1027:
add validation for coverage meta to make it consistent

Related to: https://github.com/screwdriver-cd/screwdriver/issues/1452

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -77,15 +77,17 @@ export default Component.extend({
 
     // override coverage info if set in build meta
     if (buildMeta && buildMeta.tests) {
+      const coverage = String(buildMeta.tests.coverage);
+      const tests = String(buildMeta.tests.results);
       let coverageInfo = this.get('coverageInfo');
 
-      if (buildMeta.tests.coverage) {
-        coverageInfo.coverage = `${buildMeta.tests.coverage}%`;
+      if (coverage.match(/^\d+$/)) {
+        coverageInfo.coverage = `${coverage}%`;
         coverageInfo.coverageUrl = '#';
       }
 
-      if (this.get('buildMeta').tests.results) {
-        coverageInfo.tests = buildMeta.tests.results;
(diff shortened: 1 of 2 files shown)


app/components/build-banner/component.js
```diff
@@ -77,15 +77,17 @@ export default Component.extend({
 
     // override coverage info if set in build meta
     if (buildMeta && buildMeta.tests) {
+      const coverage = String(buildMeta.tests.coverage);
```
tkyi: wondering if it's better to stay consistent with prefixing special meta elements like meta PR comments (https://docs.screwdriver.cd/user-guide/metadata)
so this would be `buildMeta.meta.tests.coverage`

minz1027: this is following @jithin1987's requirement here: https://github.com/screwdriver-cd/screwdriver/issues/1452. hmm i don't think it's necessary to have one extra layer, for pr comment, that is to be consistent with v3. And builds meta is actually under `build` directly instead of `meta.build`, the doc is misleading or wrong... lol

app/components/build-banner/component.js
```diff
@@ -77,15 +77,17 @@ export default Component.extend({
...
+      const tests = String(buildMeta.tests.results);
       let coverageInfo = this.get('coverageInfo');
 
-      if (buildMeta.tests.coverage) {
-        coverageInfo.coverage = `${buildMeta.tests.coverage}%`;
+      if (coverage.match(/^\d+$/)) {
```
jithin1987: do we have standard validation libraries in Ember for this ? cc @DekusDenial

DekusDenial: yes, but see no point in making this a computed prop

app/components/build-banner/component.js
minz1027: updated the doc for default meta already

tkyi: 😭 
okay 👍

PR #416, opened 2019-03-11
fix: add start button for restricted PR

DekusDenial:
Since the job & build models are lazily fetched, meaning, you will see the jobs first before their builds info got loaded into the models, the start button won't appear until the next render to accommodate for this flow.

Currently, two possible situations that a PR job that has no build.
1. PR's changed files are not in the sourcePaths of any of the jobs in that PR
2. Restricted PR setting causes no build is ever created (thus https://github.com/screwdriver-cd/screwdriver/issues/1528)

Because we do not mark PR job (whether it's restricted or no build simply becuz of no changed file), the UI has to know when to display the start button only for the (2) case.

There could be more edge cases involving PR restriction. But this PR only address the most common ones.

Screen Shot 2019-03-11 at 5 21 43 PM

blocked by: https://github.com/screwdriver-cd/screwdriver/pull/1536
related: https://github.com/screwdriver-cd/screwdriver/issues/1528

### DIFF
--- a/app/build/model.js
+++ b/app/build/model.js
@@ -75,6 +75,10 @@ export default DS.Model.extend({
   }),
   endTimeWords: computed('endTime', {
     get() {
+      if (!this.get('endTime')) {
+        return null;
+      }
+
       return `${durationText.call(this, 'endTime', 'now')} ago`;
     }
   }),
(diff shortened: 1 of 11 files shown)


tkyi:
Wondering if you should always be able to Start/Stop all PR builds from the pipeline page 🤔

DekusDenial:
@tkyi 
Start? Yes
Stop? Prolly not due to the likelihood peeps stopping builds mid-flight or fat finger still a thing...lol

DekusDenial:
unfortunately there are some issues with this PR that I need to fix in another PR

PR #421, opened 2019-03-22
Fix: Make clickable area for collections larger

cythrawll:
The collections flyout has a hover effect over the whole "button" but only the text is clickable.  Makes it harder to toggle between collections.  This style tweak should make the whole button area interactive.


cythrawll:
Hold off trashcans positioning broken.

cythrawll:
okay trashcans are operational

jithin1987:
Thank you for the fixes, much appreciated. Keep them coming :)

PR #423, opened 2019-03-25
fix: show 7 chars of sha

minz1027:
show 7 chars of sha to be consistent with git


tkyi:
Also in `build-banner/component.js` and `build/route.js`

minz1027:
@tkyi addressed!

PR #426, opened 2019-03-28
fix(1545): event reload is not enough in some cases.

jithin1987:
This change modifies serializer to use existing build step data if call is from `findHasMany`

cc @minz1027 @DekusDenial

### DIFF
--- a/app/build/serializer.js
+++ b/app/build/serializer.js
@@ -1,4 +1,5 @@
 import { merge } from '@ember/polyfills';
+import { get } from '@ember/object';
 import DS from 'ember-data';
 
 export default DS.RESTSerializer.extend({
@@ -11,6 +12,18 @@ export default DS.RESTSerializer.extend({
   normalizeResponse(store, typeClass, payload, id, requestType) {
     if (payload.build) {
       payload.build.statusMessage = payload.build.statusMessage || null;
+    }
+
+    if (requestType === 'findHasMany' && Array.isArray(payload.builds)) {
+      payload.builds
+        .filter(b => store.hasRecordForId('build', b.id))
+        .forEach((b) => {
+          let storeBuild = store.peekRecord('build', b.id);
+
+          if (storeBuild) {
+            b.steps = get(storeBuild, 'steps').toArray();
+          }
+        });
     }
 
     return this._super(store, typeClass, payload, id, requestType);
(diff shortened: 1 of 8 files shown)


app/build/serializer.js
```diff
@@ -13,6 +14,18 @@ export default DS.RESTSerializer.extend({
       payload.build.statusMessage = payload.build.statusMessage || null;
     }
 
+    if (requestType === 'findHasMany' && Array.isArray(payload.builds)) {
```
jithin1987: our issue is findHasMany and findRecord calls return different data from API. Since findRecord has complete steps  data, prioritize steps data from existing data in store.

DekusDenial: so if i understand correctly, the issue u mentioned is about going from the pipeline's build detail page back to pipeline's event page?

jithin1987: `findHasMany` is what job.builds & event.builds use which results in call to events/id/builds & jobs/id/bulds api call which does not return proper steps data.

So we can ignore steps data coming from these API calls, since if present in Ember store then it has to be either 
1. same data without proper steps info from another `findHasMany` call or
1. build data from a proper /builds/id api call which has proper steps data.

minz1027:
need to fix a test case!

app/build/serializer.js
DekusDenial: right, and the only place that call out to `findRecord` is from pipeline's build page. So can I assume that it's fine for pipeline's event page to have stale steps data from its corresponding `findHasMany` call, even though it may be stale for very short moment?

app/mixins/model-reloader.js
```diff
@@ -1,8 +1,13 @@
 import { later, cancel } from '@ember/runloop';
+import DS from 'ember-data';
 import Mixin from '@ember/object/mixin';
 import ENV from 'screwdriver-ui/config/environment';
 
 export default Mixin.create({
+  /**
+   * Parameter to indicate reloading is paused
+   */
+  isPaused: DS.attr('boolean', { defaultValue: false }),
```
DekusDenial: maybe just leave it as `false` cuz there is a controller use this mixin as well

d2lam:
Can we remove package lock and add to git ignore? It adds a lot of noise each time.

jithin1987:
ideally package.json should be checked in. That's one of the reason it was created. 

@DekusDenial  any thoughts ?

app/build/serializer.js
jithin1987: yep, step data from events/id/builds is static, it never changes.

PR #431, opened 2019-04-03
feat(939): Allow builds to label event using meta.label

tkyi:
## Context
It would be easier to rollback to a specific event if there was a more human-readable way to label/distinguish between events.

## Objective
This PR displays `meta.label` in the pipeline events page as "label". Defaults to showing nothing if the label does not exist.

In this example `meta.label` is set to `VERSION_1.2.3`:
Screen Shot 2019-04-03 at 4 26 04 PM

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/939

### DIFF
--- a/app/components/pipeline-event-row/styles.scss
+++ b/app/components/pipeline-event-row/styles.scss
@@ -6,4 +6,10 @@
       border: none;
     }
   }
+
+  .label {
+    display: inline-block;
+    float: right;
+    color: $grey-600;
+  }
 }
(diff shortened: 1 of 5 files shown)


tkyi:
@DekusDenial I'm definitely a frontend n00b so I welcome any/all suggestions. 🙇‍♀️ 

Does this look better?
Screen Shot 2019-04-05 at 1 33 18 PM

PR #434, opened 2019-04-08
fix(build metrics): build metrics UI update

DekusDenial:
- added the steps chart
- added a filter for successful data points only
- added more preset range options
- added cursor line for better visualize interaction
- added more vibrant colors to color palette
- reduced number of monkey-patched prototype methods from c3
- custom tooltip and legend
- added click to lock down tooltip and fixated on selected data
- added "copy" button to tooltip to copy tooltip's text content to clipboard
- added "only" button to legend to hide all but the selected data
- unified the custom date selector into one calendar view and select dates only
- separated steps chart interaction from the rest of the charts (events and builds)
- refactored based on similarity between the charts
- added deep-linking to build as well as its steps (see tooltip on builds and steps charts)

package-lock changed because I have updated some deps

update to: https://github.com/screwdriver-cd/screwdriver/issues/1412
fix: https://github.com/screwdriver-cd/screwdriver/issues/962

### DIFF
--- a/app/pipeline/build/step/route.js
+++ b/app/pipeline/build/step/route.js
@@ -1,0 +1,18 @@
+import BuildRoute from '../route';
+
+export default BuildRoute.extend({
+  model(params) {
+    this.controllerFor('pipeline.build').set('preselectedStepName', params.step_id);
+
+    return this._super(this.paramsFor('pipeline.build'));
+  },
+  afterModel(model) {
+    this._super(model);
+
+    const stepName = this.controllerFor('pipeline.build').get('preselectedStepName');
+
+    if (!model.build.get('steps').findBy('name', stepName)) {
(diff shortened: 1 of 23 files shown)


app/pipeline/build/step/route.js
```diff
@@ -0,0 +1,18 @@
+import BuildRoute from '../route';
+
+export default BuildRoute.extend({
+  model(params) {
+    this.controllerFor('pipeline.build').set('preselectedStepName', params.step_id);
+
+    return this._super(this.paramsFor('pipeline.build'));
+  },
+  afterModel(model) {
+    this._super(model);
```
jithin1987: what does this do ?

DekusDenial: Since this route really just the same as the build route, it will need the model from that route to resolve some conditions

PR #435, opened 2019-04-09
fix(1516): Reload after stop build

tkyi:
Should reload builds after stopping one.

Related to https://github.com/screwdriver-cd/screwdriver/issues/1516

### DIFF
--- a/app/pipeline/events/controller.js
+++ b/app/pipeline/events/controller.js
@@ -286,6 +286,7 @@ export default Controller.extend(ModelReloaderMixin, {
         build.set('status', 'ABORTED');
 
         return build.save()
+          .then(() => job.hasMany('builds').reload())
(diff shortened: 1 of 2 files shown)


app/pipeline/events/controller.js
```diff
@@ -286,7 +286,8 @@ export default Controller.extend(ModelReloaderMixin, {
         build.set('status', 'ABORTED');
 
         return build.save()
-          .catch(e => this.set('errorMessage', Array.isArray(e.errors) ? e.errors[0].detail : ''));
+          .catch(e => this.set('errorMessage', Array.isArray(e.errors) ? e.errors[0].detail : ''))
+          .finally(() => job.hasMany('builds').reload());
```
jithin1987: can you change this to `then` ? we don't need to reload in-case something went wrong.

PR #438, opened 2019-04-10
fix(build metrics): add UTC display option

DekusDenial:


DekusDenial:
sure, this can be merged cuz it's separate concept

PR #440, opened 2019-04-16
fix(1058): Pass in causeMessage to event create

tkyi:
## Context
When an event is started by a periodic build, it can be hard for users to know if the build was started manually by a user, through a commit, or by Screwdriver automatically.

## Objective
This PR passes in `causeMessage` when a new event is created manually.

## Related links
Related to https://github.com/screwdriver-cd/screwdriver/issues/1058

### DIFF
--- a/app/pipeline/events/controller.js
+++ b/app/pipeline/events/controller.js
@@ -220,10 +220,13 @@ export default Controller.extend(ModelReloaderMixin, {
     startMainBuild() {
       this.set('isShowingModal', true);
 
+      const token = get(this, 'session.data.authenticated.token');
+      const user = get(decoder(token), 'username');
       const pipelineId = this.get('pipeline.id');
       const newEvent = this.store.createRecord('event', {
         pipelineId,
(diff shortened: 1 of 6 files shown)


app/pipeline/events/controller.js
```diff
@@ -220,10 +220,13 @@ export default Controller.extend(ModelReloaderMixin, {
     startMainBuild() {
       this.set('isShowingModal', true);
 
+      const token = get(this, 'session.data.authenticated.token');
+      const user = get(decoder(token), 'username');
```
jithin1987: Also can we set the `user` displayed as this user, instead of the commit user ?

Another aspect to see if during external trigger. During external trigger `causeMessage` is clear, but perhaps the user should also be set as `scheduler`

jithin1987: Unless we are already doing it.

tkyi: `user` should already be this user.

PR #441, opened 2019-04-19
feat(1285): Can stop all builds in a selected event

tkyi:
## Context

It can be annoying for users to have to manually stop each build in an event. It would be nice if they could stop all running/queued/blocked builds with one click or API call.

## Objective

This PR adds a Stop button to the main pipeline event graph so a user can stop all running builds with one click. The button only shows if there are builds running.

Example:
Screen Shot 2019-04-19 at 3 54 47 PM

On hover it will show a text hint:
Screen Shot 2019-04-19 at 3 54 57 PM

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1285

### DIFF
--- a/app/components/pipeline-graph-nav/styles.scss
+++ b/app/components/pipeline-graph-nav/styles.scss
@@ -11,6 +11,24 @@
     line-height: 33px;
   }
 
+  .btn.event__stop {
+    float: right;
+    border: 1px solid $sd-aborted;
+    background-color: #fff;
+    color: $sd-aborted;
+    width: 100px;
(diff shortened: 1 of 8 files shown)


jithin1987:
Does this take care of PR chain events as well ? If not will that be a separate change ?

tkyi:
Should handle PR chain events as well. The endpoint just calls `event.getBuilds`
https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/events/stopBuilds.js#L80

PR #443, opened 2019-04-25
fix: PR list drop down to show only valid builds

jithin1987:
## Changes

1. Always navigate to step url unless build is in success state.
2. Fix for some SHAs not clickable in PR build detail page when selecting a past build.

### DIFF
--- a/tests/unit/pipeline/build/controller-test.js
+++ b/tests/unit/pipeline/build/controller-test.js
@@ -6,6 +6,7 @@ import Pretender from 'pretender';
 import Service from '@ember/service';
 import wait from 'ember-test-helpers/wait';
 import sinon from 'sinon';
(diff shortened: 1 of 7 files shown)


tests/unit/pipeline/build/controller-test.js
```diff
@@ -280,3 +281,34 @@ test('it reloads a build', function (assert) {
     assert.ok(true);
   });
 });
+
+sinonTest('it changes build step', function (assert) {
```
tkyi: why `sinonTest` vs `test`?

jithin1987: because this test uses `stub` https://github.com/screwdriver-cd/ui/pull/443/files#diff-f98bd7f1412616760bfcea1bb5d502ecR289

PR #446, opened 2019-05-09
feat(1455): allow create and update with rootDir (6)

minz1027:
- allow create and update with rootDir
for both create and update, we always send both fields to the API since api's logic is to construct the scmUri based on `checkoutUrl` and `rootDir`.

Blocked by: https://github.com/screwdriver-cd/screwdriver/pull/1601

Screen Shot 2019-05-10 at 12 05 08 PM
Screen Shot 2019-05-10 at 12 04 51 PM
Screen Shot 2019-05-10 at 2 13 18 PM

Related to https://github.com/screwdriver-cd/screwdriver/issues/1455


tkyi:
probably a separate title for the source dir would be more clear

minz1027:
@tkyi will hide the rootDir config under a checkbox

app/pipeline/model.js
```diff
@@ -21,6 +23,16 @@ export default DS.Model.extend({
   metrics: DS.hasMany('metric', { async: true }),
 
   appId: alias('scmRepo.name'),
-  branch: alias('scmRepo.branch'),
+  branch: computed('scmRepo.{branch,rootDir}', {
```
tkyi: hm will be confusing when it shows up on Collections/Search page?

minz1027: i think it's okay, if we added another column for this, it will take up a lot of the space while for most pipelines, the value is going to be empty

jithin1987: we should check UX for how it shows up, in collections page for pipelines which do have rootDir.

tests/integration/components/pipeline-options/component-test.js
```diff
@@ -95,19 +95,57 @@ module('Integration | Component | pipeline options', function(hooks) {
...
     await triggerKeyEvent('.text-input', 'keyup', 'SPACE');
 
-    assert.dom('.text-input').hasValue(scm);
+    assert.dom('.scm-url').hasValue(scm);
+
+    await click('button.blue-button');
+  });
+
+  test('it updates a pipeline with rootDir', async function(assert) {
+    const scm = 'git@github.com:foo/bar.git';
```
jithin1987: can you add one `assert.expect(6)` here, since some actions are async which will validate if all assertions were indeed called.

minz1027: sure

minz1027: addressed

PR #461, opened 2019-06-20
fix: add "trusted" to templates & commands

DekusDenial:
## Context

Add visual aid to `trusted` templates and commands

## Objective

prepend the svg for `trusted` to the name of template/command
and fix search on lowercase match

Screen Shot 2019-06-20 at 3 33 10 PM
Screen Shot 2019-06-20 at 3 33 23 PM

## References

https://github.com/screwdriver-cd/screwdriver/issues/1427

blocked by: 
https://github.com/screwdriver-cd/data-schema/pull/340
https://github.com/screwdriver-cd/datastore-sequelize/pull/52

## License


app/commands/detail/controller.js
```diff
@@ -16,6 +16,9 @@ export default Controller.extend({
       return this.commands[0];
     }
   }),
+  trusted: computed('commands.[]', function computeTrusted() {
+    return this.commands.reduce((trusted, c) => trusted || !!c.trusted, false);
```
adong: Does this line of code means if all commands are `trusted`, then mark as `trusted`? 

Do you mind to switch to `Array.every`? Reduce is less verbose than `.every`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

app/application/controller.js
```diff
@@ -7,7 +7,7 @@ export default Controller.extend({
   collections: [],
   session: service(),
   scmContexts: alias('model'),
-  showCollections: computed('collections.[]', function() {
+  showCollections: computed('collections.[]', function showCollections() {
```
adong: My bad, it was a warning. Actually, I'm little confused why we have to give name to anonymous function. Maybe we can discuss on relaxing such rule.

app/commands/detail/controller.js
DekusDenial: my bad, should have used `Array.some`. Will make the change now

adong: Yeah, let `babel` handles the transpile. 😃

app/application/controller.js
DekusDenial: i guess SD primarily bases its rule on airbnb. I dont mind relaxing it and dont think debugging will be harder without it

PR #462, opened 2019-06-21
 fix: add admin UI to (un)mark template/command as "trusted"

DekusDenial:
## Context

Screwdriver admins will need a UI to (un)mark template/command as `trusted`

## Objective

This PR adds the required UI components

Screen Shot 2019-06-21 at 4 19 09 AM
Screen Shot 2019-06-21 at 4 18 57 AM

- Not the prettiest UI but it's better than doing separate solo pipeline/manual script/direct sql
- While api endpoints require token with admin scope, UI does have the jwt with right privilege for admins. No need to acquire another token for this.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1427

## License

### DIFF
--- a/app/command/service.js
+++ b/app/command/service.js
@@ -105,8 +105,41 @@ export default Service.extend({
             message = `${response.status} ${response.responseJSON.error}`;
           }
 
-          if (response.status === 401) {
+          if (response.status === 403) {
             message = 'You do not have the permissions to remove this command.';
+          }
+
+          return reject(message);
+        });
+    });
+  },
+  updateTrust(namespace, name, trusted) {
+    const url =
+      `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/commands/` +
+      `${encodeURIComponent(namespace)}/${encodeURIComponent(name)}/trusted`;
+    const ajaxConfig = {
+      method: 'PUT',
+      dataType: 'json',
+      url,
+      contentType: 'application/json',
+      crossDomain: true,
+      xhrFields: {
+        withCredentials: true
+      },
+      headers: {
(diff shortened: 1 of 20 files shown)


app/command/service.js
```diff
@@ -105,10 +105,43 @@ export default Service.extend({
             message = `${response.status} ${response.responseJSON.error}`;
           }
 
-          if (response.status === 401) {
+          if (response.status === 403) {
```
adong: Would it be `[401, 403].includes(response.status)` like line# 141??

app/command/service.js
```diff
@@ -105,10 +105,43 @@ export default Service.extend({
...
+      data: JSON.stringify({ trusted })
+    };
+
+    return new EmberPromise((resolve, reject) => {
+      $.ajax(ajaxConfig)
+        .done(content => resolve(content))
+        .fail(response => {
+          let message = `${response.status} Request Failed`;
+
+          if (response.status === 401 || response.status === 403) {
```
adong: Prefer `[401, 403].includes(response.status)`

app/template/service.js
```diff
@@ -105,10 +105,43 @@ export default Service.extend({
             message = `${response.status} ${response.responseJSON.error}`;
           }
 
-          if (response.status === 401) {
+          if (response.status === 403) {
```
adong: Not sure, if `401` should included here.

DekusDenial: ideally should be either, but i haven't tested it, honestly.

app/command/service.js
DekusDenial: at some point i thought of that but backed down. Will surely do that in the future. Great advice!

DekusDenial:
@adong yw, sure this repo needs a heavy refactoring as well. This is shamefully really a quick plug. 😭

PR #466, opened 2019-07-23
make timestamps non-selectable

raphaeleidus:
## Context

Copying multiple lines from a log is really annoying to do in a readable way because the timestamps are selectable.

## Objective

by making the timestamps unselectable highlighting allows you to get just the useful information

## References
_  | Screenshot
------------ | -------------
Before | Screen Shot 2019-07-23 at 1 10 13 PM
After | Screen Shot 2019-07-23 at 1 10 24 PM

## License

### DIFF
--- a/app/components/build-log/styles.scss
+++ b/app/components/build-log/styles.scss
@@ -52,6 +52,7 @@
     display: inline-block;
     vertical-align: top;
     width: 70px;
+    user-select: none;
   }
 
   .content {


adong:
Good idea, and build has failed. I'm kicking off another build

raphaeleidus:
Still failing but I can't find what the failure is, and I dont see why an SCSS change would cause a failure

tkyi:
I thought timestamps would be a useful thing
I think we thought it made sense since they are also selectable in Jenkins console logs

adong:
@raphaeleidus How about we make this as optional toggle button? So some can have timestamp copied and some wont?

jithin1987:
> Still failing but I can't find what the failure is, and I dont see why an SCSS change would cause a failure
@raphaeleidus  it's failing on a test which does copy of the logs

```
 not ok 54 Chrome 75.0 - [926 ms] - Integration | Component | build log: it starts fetching more log for a chosen completed step
00:08:24     --- 
```

Try running this test alone locally to see what's happening. Perhaps the css change is impacting mouse events which is preventing event handlers related to scrolling from  firing.

jithin1987:
Closing this. Please re-open once you have a fix for failing test.

PR #469, opened 2019-07-26
feat: add query parameter for activeTab for artifacts and steps

adong:
## Context

Users want to share which active tab they are actively viewing.

## Objective

This PR adds a `query parameter` to preserve what tab is currently `active`, so that users can share links 

`Steps`:  http://localhost:4200/pipelines/1/builds/2

image

Or 
`Artifacts`: http://localhost:4200/pipelines/1/builds/2?tab=artifacts

image

## License

### DIFF
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -20,6 +20,12 @@ export default Controller.extend({
   stepList: mapBy('build.steps', 'name'),
   isShowingModal: false,
   errorMessage: '',
+  activeTab: 'steps',
+  queryParams: [
+    {
+      activeTab: { type: 'string' }
+    }
+  ],
   prEvents: computed('model.{event.pr.url,pipeline.id}', {
     get() {
       if (this.get('model.event.type') === 'pr') {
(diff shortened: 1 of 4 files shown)


tkyi:
Will it be possible to deep link to a specific open artifact? like we deep link to specific steps in build logs

app/pipeline/build/controller.js
```diff
@@ -20,6 +20,12 @@ export default Controller.extend({
   stepList: mapBy('build.steps', 'name'),
   isShowingModal: false,
   errorMessage: '',
+  activeTab: 'steps',
```
tkyi: I think I remember during design discussions that we wanted the activeTab to be "artifacts" if the build had run successfully

adong:
Technical References:

I thought I could setup a query parameter in Controller `queryParams`, then use `setupController` in its routers to have a pre-defined default value. 

Ran into issues like 

1. https://github.com/emberjs/ember.js/issues/5465
2. https://discuss.emberjs.com/t/how-to-programatically-setup-default-query-params-after-model/13666/7

No success. 



Then used `Ember.run.next` inside Route's `setupController`, that worked. After read from https://github.com/emberjs/ember.js/issues/14606#issuecomment-260640665

image



So I ended up not fighting against the ember framework. Set a query parameter default value in controller, i.e. `steps`, knowing that default parameter wont show up in the url.

app/pipeline/build/controller.js
adong: Oh. I see, that requires some other changes.

adong:
Closed in favor of `path parameter` in https://github.com/screwdriver-cd/ui/pull/470

PR #472, opened 2019-08-01
feat: show template and commands in validator

kumada626:
## Context

We want to know what templates and commands are used in a job when using validator or visiting template details page.
Therefore, I added templates and commands link on the result of validation and template details page.

validator
スクリーンショット 2019-08-01 13 55 53

template details
スクリーンショット 2019-08-01 13 57 19

## Objective

If `SD_TEMPLATE_FULLNAME` env exists, a job uses template.
So I add link to template on job details.

And, validator searches all commands with regex `sd-cmd\s+exec\s+([\w-]+\/[\w-]+)` to find executed sd-cmd.
If sd-cmd is executed in a job, I add link to all executed sd-cmd on job details.

## License

### DIFF
--- a/app/components/validator-job/component.js
+++ b/app/components/validator-job/component.js
@@ -42,14 +42,14 @@ export default Component.extend({
       const commands = this.steps;
       const regex = /sd-cmd\s+exec\s+([\w-]+\/[\w-]+)/g;
       let sdCommands = [];
-      let matchRes;
 
       if (commands === []) {
         return [];
       }
 
       commands.forEach(c => {
-        matchRes = regex.exec(c.command);
+        let matchRes = regex.exec(c.command);
+
         while (matchRes !== null) {
           if (!sdCommands.includes(matchRes[1])) {
             sdCommands.push(matchRes[1]);
--- a/app/components/validator-job/component.js
+++ b/app/components/validator-job/component.js
@@ -5,6 +5,11 @@ export default Component.extend({
   classNameBindings: ['hasParseError', 'collapsible'],
   isOpen: true,
   collapsible: true,
+  getTemplateName: computed('job', {
+    get() {
+      return this.get('job.environment.SD_TEMPLATE_FULLNAME');
+    }
+  }),
   hasParseError: computed('job', {
     get() {
       return this.get('job.commands.0.name') === 'config-parse-error';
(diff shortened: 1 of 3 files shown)


app/components/validator-job/template.hbs
```diff
@@ -113,4 +113,16 @@
     </ul>
   </span>
 </div>
+<div class="sd-commands" title="These are the sd-commands that will be executed in the job.">
+  <span class="label">Sd-commands:</span>
```
wdstar: We should use plain `Commands:` label?
`Sd-command` seems not to be used except for `sd-command.yaml`.
https://docs.screwdriver.cd/user-guide/commands

kumada626: fix label name.

app/components/validator-job/component.js
```diff
@@ -32,6 +37,30 @@ export default Component.extend({
       return [];
     }
   }),
+  sdCommands: computed('job', {
+    get() {
+      const commands = this.steps;
+      const regex = /sd-cmd\s+exec\s+([\w-]+\/[\w-]+)/g;
+      let sdCommands = [];
+      let matchRes;
```
yoshwata: do you need to declare at here?
I think it is sufficient like below on L52
```
let matchRes = regex.exec(c.command);
```

tests/integration/components/validator-job/component-test.js
```diff
@@ -223,6 +240,27 @@ module('Integration | Component | validator job', function(hooks) {
     assert.dom('.sourcePaths .value ul li:last-child').hasText('src/folder/');
   });
 
+  test('it renders sd-commands', async function(assert) {
+    this.set('jobMock', {
+      image: 'int-test:1',
+      commands: [
+        { name: 'step1', command: 'echo hello' },
+        { name: 'step2', command: 'sd-cmd exec foo/bar@0.0.1 foobar' }
```
yoshwata: I'd like you to use 2 or more sd-cmd in step because UI seems to show some lines of command [here](https://github.com/screwdriver-cd/ui/pull/472/files#diff-35e77da0f7868ba13c3ed69f8d861762R120)

app/components/validator-job/component.js
kumada626: fixed.

tests/integration/components/validator-job/component-test.js
kumada626: Fixed testcase to use 2 kinds of sd-cmd 3 times.

PR #476, opened 2019-09-09
feat: update restart modal sentences

yuichi10:
## Context
Sometime user get confused when the job's sha is not the pipeline's branch. It happen when user set  branch specific job such as `requires: [~commi, ~commit:developnt]`.

I try to make it easy to understand what kind job user will start.

## Objective
Add commit message and able to jump commit change page.

before:
スクリーンショット 2019-09-09 12 10 16

after:
スクリーンショット 2019-09-09 12 01 31

## reference
related https://github.com/screwdriver-cd/screwdriver/issues/1706

## License

### DIFF
--- a/app/components/pipeline-workflow/template.hbs
+++ b/app/components/pipeline-workflow/template.hbs
@@ -42,8 +42,8 @@
       }}
         <h3>Are you sure?</h3>
         <p>
-          You are about to start the job <code>{{tooltipData.job.name}}</code> in a new event with the same
-          context of the selected event for sha <code>#{{selectedEventObj.truncatedSha}}</code>
+          Job: <code>{{tooltipData.job.name}}</code><br>
+          Commit:<a href={{selectedEventObj.commit.url}}><code>{{selectedEventObj.commit.message}}</code>({{selectedEventObj.truncatedSha}})</a> 
         </p>
         <div class="row">
           <div class="col-xs-6">
--- a/app/components/pipeline-workflow/template.hbs
+++ b/app/components/pipeline-workflow/template.hbs
@@ -25,10 +25,10 @@
       translucentOverlay=true
       containerClass="detached-confirm-dialog"
     }}
-      <h3>Are you sure?</h3>
+      <h3>Are you sure to start?</h3>
       <p>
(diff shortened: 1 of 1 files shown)


app/components/pipeline-workflow/template.hbs
```diff
@@ -42,8 +42,8 @@
       }}
         <h3>Are you sure?</h3>
         <p>
-          You are about to start the job <code>{{tooltipData.job.name}}</code> in a new event with the same
-          context of the selected event for sha <code>#{{selectedEventObj.truncatedSha}}</code>
+          Job: <code>{{tooltipData.job.name}}</code><br>
+          Commit:<a href={{selectedEventObj.commit.url}}><code>{{selectedEventObj.commit.message}}</code>({{selectedEventObj.truncatedSha}})</a>
```
tk3fftk: Do we need to truncate a commit message which is very long?

app/components/pipeline-workflow/template.hbs
```diff
@@ -42,8 +42,8 @@
       }}
         <h3>Are you sure?</h3>
```
tk3fftk: I think we should add more information to this text (e.g. `Are you sure to start?`) if we remove the description text.

yuichi10: I'll fix to ask user`Are you sure to start?`.

app/components/pipeline-workflow/template.hbs
yuichi10: Now every message is shown. I think we need to truncate a commit message.
How about 60 characters? 
It is because good commit summary should around 50 characters.

yuichi10: UI already have truncatedMessage value.
https://github.com/screwdriver-cd/ui/blob/master/app/event/model.js#L85
When I check the UI, 150 characters commit message looks not bad.

スクリーンショット 2019-09-10 17 21 41

app/components/pipeline-workflow/template.hbs
yuichi10: I fixed.

yuichi10:
Thank you!

PR #477, opened 2019-09-12
fix: default pipeline

adong:
## Context

To avoid confusion for users that they cannot delete default pipeline (`My Pipelines`), knowing that API will drop if the collection `type` attribute is `default`.

## Objective

Remove delete option to default pipeline

## References

Before: 
image

After:
image

## License


adong:
Weird. Tests passed locally. 
image

Rebased with latest master, and will trigger build again.

PR #479, opened 2019-09-17
feat: Enable URL with specific version or tag of commands or templates

ibu1224:
## Context

When the version is clicked on the template/command details screen, the content is rewritten by javascript. Since there is no link to the details screen of each version, they cannot be shared. So instead of rewriting content we changed for a link to each template and command.

## Objective

The version of the template/command details screen is changed to `href` like below.

#### Precondition
```
command: [
  {id:1, namespace:'foo', name:'bar', version:'1.0.0'}
  {id:2, namespace:'foo', name:'bar', version:'2.0.0 - latest'}
]  
```
#### Behavior
1. `/foo/bar`
 - It shows the screen of `latest (2.0.0)`
2. `/foo/bar/1.0.0`
 - It shows the screen of `1.0.0`
3. `/foo/bar/latest`
 - It shows the screen of `latest (2.0.0)`
4. `/foo/bar/9.9.9` or `/foo/bar/nonexisttag`
 - It shows the screen of `404`
## License

### DIFF
--- a/app/commands/detail/route.js
+++ b/app/commands/detail/route.js
@@ -12,10 +12,10 @@ export default Route.extend({
       const [verPayload, tagPayload] = arr;
 
       if (params.version) {
-        const exsistVersion = verPayload.filter(t => t.version === params.version);
-        const exsistTag = tagPayload.filter(c => c.tag === params.version);
+        const versionExists = verPayload.filter(t => t.version === params.version);
+        const tagExists = tagPayload.filter(c => c.tag === params.version);
 
-        if (exsistTag.length === 0 && exsistVersion.length === 0) {
+        if (tagExists.length === 0 && versionExists.length === 0) {
           this.transitionTo('/404');
         }
       }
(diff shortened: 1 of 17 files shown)


app/commands/detail/route.js
```diff
@@ -11,6 +11,15 @@ export default Route.extend({
     ]).then(arr => {
       const [verPayload, tagPayload] = arr;
 
+      if (params.version) {
+        const exsistVersion = verPayload.filter(t => t.version === params.version);
```
tkyi:
```suggestion
        const existVersion = verPayload.filter(t => t.version === params.version);
```
or `existingVersion`

app/commands/detail/route.js
```diff
@@ -11,6 +11,15 @@ export default Route.extend({
...
+        const exsistTag = tagPayload.filter(c => c.tag === params.version);
```
tkyi:
```suggestion
        const existTag = tagPayload.filter(c => c.tag === params.version);
```
or `existingTag`

app/commands/detail/route.js
ibu1224: Thanks! I fixed!

PR #480, opened 2019-09-23
feat(947): Implement UI for user default collection

Code-Beast:
## Context

Would be nice if users can find all pipeline create by themselves in a default collection.

## Objective

This PR implements new mocks for dashboard page. Main features are as follows:
1. A default collection called "My Pipelines" will be created when user firstly sign in which user can rename but is not allowed to delete. 
image

2. User can see a guide page when a new collection is created but no pipelines are added into it yet where user can be redirected to the pipeline creation page or screwdriver documentation page.
image

3. User can choose to view pipelines in a collection in `card mode` or `list mode`. 
* Card Mode
image
* List Mode
image

4. User can delete single pipeline from a collection using the close button when hovering the cursor on a pipeline card in `card mode` or a table row in `list mode`.
* Card Mode
image
* List Mode
image

5. User can do bulk operation (`remove`/ `move to` other collection) for pipelines in a collection.
* Card Mode
image
* List Mode
image

6. User can add pipelines to the current collection using the `add` button.
image

7. User can change the collection name and description using the `setting` button.
image

8. User can copy the link to the current collection to the clipboard using the `link` button.
image

## References

Related to [947](https://github.com/screwdriver-cd/screwdriver/issues/947)

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #484, opened 2019-10-04
fix: improve dashboard/show page's performance and make /home redirect to default collection page

Code-Beast:
## Context
Two issues:
1. Default collection is created after other previously created collections, so if we want to redirect to the default collection dashboard page, we can no longer use last collection in the returned collection array.
2. Currently, in dashboard/show page, in order to enable events-thumbnail component to generate events graph, route.js has to fetch events as well as their builds (for generating duration information) which is very inefficient and can be improved by using pipeline metrics which contains event duration info that is computed in the backend.

## Objective
1. Improve dashboard/show page's performance
2. Redirect /home to default collection page
3. Fix branch info overflow issue in pipeline card

## References
[UI feat-480](https://github.com/screwdriver-cd/ui/pull/480)
[Models fix-397](https://github.com/screwdriver-cd/models/pull/397)

## License


app/components/events-thumbnail/component.js
```diff
@@ -58,7 +58,7 @@ export default Component.extend({
       .attr('y', event => 40 - y(event.duration));
 
     bars.each(function addColor(bar) {
-      this.classList.add(bar.statusColor.toLowerCase());
+      this.classList.add(bar.statusColor);
```
adong: Should we keep the logics of converting to lowercases regardless? @Code-Beast ?

app/components/pipeline-card/styles.scss
```diff
@@ -93,34 +93,41 @@
   }
 
   .branch-info {
-    display: flex;
-    justify-content: space-between;
+    // display: flex;
```
tkyi: commented code

app/components/pipeline-card/styles.scss
```diff
@@ -93,34 +93,41 @@
...
-      & a {
-        width: 78%;
-      }
-
-      & > div {
-        width: 22%;
-      }
-    }
-  }
+  // @media screen and (max-width: 1440px) {
```
tkyi: same here

app/components/events-thumbnail/component.js
Code-Beast: statusColor is always lowercase in my logic

PR #489, opened 2019-10-17
fix(1651): Allow admins to delete templates and commands with no pipelines

tkyi:
## Context

Sometimes the pipeline is removed before people get a change to delete the template/command. Even if the pipeline doesn't exist, a Screwdriver admin should be able to remove the template/command through the UI.

## Objective

This PR always shows the trash icon on the template/command pages. The backend should be able to handle the case where a pipeline doesn't exist but the user is a Screwdriver admin gracefully: https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/templates/index.js#L39-L41

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1651

## License

### DIFF
--- a/app/components/command-header/template.hbs
+++ b/app/components/command-header/template.hbs
@@ -3,11 +3,11 @@
   {{command.namespace}}/{{command.name}}
   {{#if scmUrl}}
     <a href={{scmUrl}}>{{fa-icon "code-fork" title="Source code"}}</a>
-    <a href="#" {{action "setCommandToRemove" command}} class="remove">{{fa-icon "trash" title="Delete command"}}</a>
   {{else}}
-    {{fa-icon "code-fork" title="The pipeline for this command does not exist."}} {{fa-icon "trash" title="Cannot delete command; pipeline could not be found."}}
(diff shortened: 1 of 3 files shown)


app/components/command-header/template.hbs
```diff
@@ -5,7 +5,7 @@
     <a href={{scmUrl}}>{{fa-icon "code-fork" title="Source code"}}</a>
     <a href="#" {{action "setCommandToRemove" command}} class="remove">{{fa-icon "trash" title="Delete command"}}</a>
   {{else}}
-    {{fa-icon "code-fork" title="The pipeline for this command does not exist."}} {{fa-icon "trash" title="Cannot delete command; pipeline could not be found."}}
+    {{fa-icon "code-fork" title="The pipeline for this command does not exist."}} <a href="#" {{action "setCommandToRemove" command}} class="remove">{{fa-icon "trash" title="Delete command"}}</a>
```
jithin1987: can we show it only if it's admin ?

tkyi: updated

PR #490, opened 2019-10-18
fix: change word from move-to to copy-to and fix metrics error bubbling up

adong:
## Context

1) Label `Move To` actually keeps the original pipeline inside the original collection, which should be. We need to change the label from `Move To` to `Copy To`.

2) Metrics fetch no longer exist pipelineId will result to 404 pages.

## Objective

This PR Changes the label from `Move To` to `Copy To`

Before: 

image

After:
image

2)  While it still shows errors, but this time the errors are caught by Ember.RSVP, instead of bubbling up, and will not 404 pages
image

## License

### DIFF
--- a/app/components/collection-view/template.hbs
+++ b/app/components/collection-view/template.hbs
@@ -27,7 +27,7 @@
 
       <BsModal
         @class="add-pipeline-modal"
-        @open={{showAddPipelineModal}} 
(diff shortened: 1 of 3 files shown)


adong:
Hmmm, though myself curious why left side will result response as uncaught 🤔. 
image

PR #491, opened 2019-10-21
fix(550): change color code

wahapo:
## Context
I received a report that it was difficult to distinguish between green and red on https://github.com/screwdriver-cd/screwdriver/issues/550.
Please wait a little. I will check whether really there is an effect.

## Objective
I adjusted the green and red color codes with reference to https://github.com/screwdriver-cd/screwdriver/issues/550. I simulating the vision of individuals with protanopia.

### pipeline
- Current
    pipeline-current
- Modify
    pipeline-after

### metrics
- Current
    metrics-current
- Modify
    metrics-after

### danger zone
- Current
    danger-current
- Modify
    danger-after

### metrics
There is a problem that the text color of the link looks black. I didn't realize it is a link. I used the same background color as this(nav) part.
- Current
    Collections-current
- Modify
    Collections-after

## References
https://github.com/screwdriver-cd/screwdriver/issues/550
http://www.dic-global.com/en/110th/universal/

## License


tkyi:
I like the idea of making the page more accessible to people with color vision deficiency. 👍 

Maybe @chasturansky could take a look?

tkyi (approved):
It's been a long time waiting on approvals but I think we should just go for it
we currently don't have an active designer and I think it's a good idea

wahapo:
I forgot. Thank you so much!

PR #495, opened 2019-10-29
fix: Reverse eventsInfo array because of pipeline model change

Code-Beast:
## Context
Events-thumbnail components show outdated events info because of an API bug. Now the API is fixed, we need to make corresponding changes for the UI.

## Objective
Make events-thumbnail show latest events info.

## References
[screwdriver feat(947)](https://github.com/screwdriver-cd/ui/pull/480)

## License

### DIFF
--- a/app/components/collection-view/component.js
+++ b/app/components/collection-view/component.js
@@ -171,7 +171,7 @@ export default Component.extend({
           };
 
           if (metrics && metrics.length) {
-            const lastEvent = metrics.get('lastObject');
+            const lastEvent = metrics.get('firstObject');
             const lastEventStartTime = new Date(lastEvent.createTime);
             const lastEventStartYear = lastEventStartTime.getFullYear();
             const lastEventStartMonth = (lastEventStartTime.getMonth() + 1)
--- a/app/components/events-thumbnail/component.js
+++ b/app/components/events-thumbnail/component.js
@@ -41,7 +41,7 @@ export default Component.extend({
     // Fixed number of bars
     if (this.events.length < MAX_NUM_EVENTS_SHOWN) {
       this.events = [
-        ...this.events,
+        ...this.events.reverse(),
         ...new Array(MAX_NUM_EVENTS_SHOWN - totalNumberOfEvents).fill({
           duration: 0,
           statusColor: 'build-empty'
--- a/tests/integration/components/collection-view/component-test.js
+++ b/tests/integration/components/collection-view/component-test.js
@@ -168,18 +168,6 @@ const mockCollections = [
 const mockMetricsMap = EmberObject.create({
   1: [
     {
-      status: 'SUCCESS',
-      duration: 23173
-    },
-    {
-      status: 'SUCCESS',
-      duration: 20011
-    },
-    {
-      status: 'FAILURE',
-      duration: 39234
-    },
-    {
       creatTime: 'Tue Oct 01 2019 15:55:52 GMT-0700 (Pacific Daylight Time)',
       status: 'FAILURE',
       duration: 42718,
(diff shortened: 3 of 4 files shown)


adong:
This would require another small fix. UI needs to pass in the `sort`ing order to be `descending` to `getMetrics`

Code-Beast:
> This would require another small fix. UI needs to pass in the `sort`ing order to be `descending` to `getMetrics`

I checked the data schema and found that pagination schema has a default 'descending', the sort field of query is 'descending' already.

Code-Beast:
> @Code-Beast it's card coded to `ascending`See https://github.com/screwdriver-cd/models/pull/404/files For timeseries it has to be ascending. But we wanted the latest page of data. So now UI has to call pipeline metrics by passing `descending` and then `reverse` it to convert back to time series of latest page of metrics. Also the time information in the card should pertain to the latest event (last entry in time series)

There's one thing I am not sure about. In the query schema of pipeline/metrics, paginate.sort has a default 'descending', thus it makes no difference whether or not we pass a 'sort' query parameter. (In screwdriver, the default sort is 'descending' while in pipeline model, for getMetrics field, the default sort is 'ascending').

app/components/collection-view/component.js
```diff
@@ -171,7 +171,7 @@ export default Component.extend({
           };
 
           if (metrics && metrics.length) {
-            const lastEvent = metrics.get('lastObject');
+            const lastEvent = metrics.get('firstObject');
```
jithin1987: shouldn't this be lastObject itself ? after `reverse` won't  `lastobject`  be the latest event ?

Code-Beast: reverse happens in events-thumbnail. Here in collection-view I am fetching from the original returned metrics, so it should be firstObject

jithin1987:
> There's one thing I am not sure about. In the query schema of pipeline/metrics, paginate.sort has a default 'descending',

@Code-Beast can you point me to where in code this is happening ? As far as I see from model logic, only `metrics` uses `ascending`  by default to generate the time series format.

Code-Beast:
> > There's one thing I am not sure about. In the query schema of pipeline/metrics, paginate.sort has a default 'descending',
> 
> @Code-Beast can you point me to where in code this is happening ? As far as I see from model logic, only `metrics` uses `ascending` by default to generate the time series format.

Sure. https://github.com/screwdriver-cd/data-schema/blob/master/api/pagination.js

jithin1987:
> Sure. https://github.com/screwdriver-cd/data-schema/blob/master/api/pagination.js

@Code-Beast that the base default if it's not definfed. But model is overriding it and had hard coded to `ascending`. See my PRs for relaxing it.

https://github.com/screwdriver-cd/models/pull/404/files
https://github.com/screwdriver-cd/screwdriver/pull/1818/files

Code-Beast:
> > Sure. https://github.com/screwdriver-cd/data-schema/blob/master/api/pagination.js
> 
> @Code-Beast that the base default if it's not defined elsewhere. But model is overriding it and had hard coded to `ascending`. See my PRs for relaxing it.
> 
> https://github.com/screwdriver-cd/models/pull/404/files
> https://github.com/screwdriver-cd/screwdriver/pull/1818/files

image
At this point, 'sort' in the query is the default, which is 'descending', and we pass it the parameter for pipeline.getMetrics();

image
When it is passed in here, it will override the default here, which is 'ascending' and give us the correct  return value for the pagination case.

That's why I think it won't affect the result whether you pass in a 'sort' query parameter when making the API call or not.

jithin1987:
As discussed offline, @Code-Beast  is right abut hapi defaulting to descending. But now it will break Metrics page, since it needs `ascending` sort order. @Code-Beast can you please add that to this PR ?

Code-Beast:
> As discussed offline, @Code-Beast is right abut hapi defaulting to descending. But now it will break Metrics page, since it needs `ascending` sort order. @Code-Beast can you please add that to this PR ?

image

Because only when metrics are fetched by pagination will the 'sort' query parameter be passed in pipeline.getMetrics() within config, current changes won't break metrics page because metrics are not fetched by pagination on that page.

PR #498, opened 2019-10-30
fix: semantic release

adong:
## License


jithin1987:
this is not required. Looks like PR title will suffice.

PR #500, opened 2019-10-30
fix: change collection to collection.id

adong:
## Context

Fix a small bug that adding pipeline to collection.

## Objective
This PR fixed the function arguments from model to `model.id`.

## License

### DIFF
--- a/app/pipeline/controller.js
+++ b/app/pipeline/controller.js
@@ -8,8 +8,8 @@ export default Controller.extend({
   pipeline: alias('model.pipeline'),
   collections: alias('model.collections'),
   actions: {
-    addToCollection(pipelineId, collectionId) {
-      return this.store.findRecord('collection', collectionId).then(collection => {
+    addToCollection(pipelineId, c) {
+      return this.store.findRecord('collection', c.id).then(collection => {
         const pipelineIds = collection.get('pipelineIds');
 
         if (!pipelineIds.includes(pipelineId)) {
--- a/app/pipeline/controller.js
+++ b/app/pipeline/controller.js
@@ -9,7 +9,11 @@ export default Controller.extend({
   collections: alias('model.collections'),
   actions: {
     addToCollection(pipelineId, collection) {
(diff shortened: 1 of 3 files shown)


app/pipeline/controller.js
```diff
@@ -8,8 +8,8 @@ export default Controller.extend({
   pipeline: alias('model.pipeline'),
   collections: alias('model.collections'),
   actions: {
-    addToCollection(pipelineId, collectionId) {
-      return this.store.findRecord('collection', collectionId).then(collection => {
+    addToCollection(pipelineId, c) {
+      return this.store.findRecord('collection', c.id).then(collection => {
```
jithin1987: you have the collection here. Then why fetch it again ?

adong: I thought it was for fetching new collection's data. Just removed it.

@Code-Beast 🤔

Code-Beast: it's the collection the pipeline is added to

app/pipeline/controller.js
```diff
@@ -8,16 +8,10 @@ export default Controller.extend({
   pipeline: alias('model.pipeline'),
   collections: alias('model.collections'),
   actions: {
-    addToCollection(pipelineId, collectionId) {
-      return this.store.findRecord('collection', collectionId).then(collection => {
-        const pipelineIds = collection.get('pipelineIds');
+    addToCollection(pipelineId, collection) {
+      collection.pipelineIds.push(pipelineId);
 
-        if (!pipelineIds.includes(pipelineId)) {
```
jithin1987: we should keep this. check.

app/pipeline/controller.js
adong: If it only added to the current user's collection, then maybe we dont need to fetch.

Code-Beast: @adong Oh, just notice it is addToCollection in pipeline/controller.js, I thought this was the addMultipleToCollection method in dashboard/show

PR #504, opened 2019-11-05
fix: The validator page shows unique command link 

ibu1224:
## Context
https://github.com/screwdriver-cd/ui/pull/503
In the above link, I added a version to template link and command link.
However, the command version number is not unique. 
Therefore, I changed the showed version uniquely.

## Objective

In this case, the validator page show the same command link.
スクリーンショット 2019-11-05 12 08 32
Therefore, I tried to check the link to show.

## License


app/components/validator-job/component.js
```diff
@@ -56,8 +56,12 @@ export default Component.extend({
       commands.forEach(c => {
         let matchRes = regex.exec(c.command);
 
+        let commandExist = sdCommands.filter(
+          command => command.command === matchRes[1] && matchRes[2]
```
wahapo: Isn’t it like this?
`command => command.command === matchRes[1] && command.version === matchRes[2]`

If there is no match on line 57, `matchRes` will be null. It seems to need to be moved into the while statement.

tests/integration/components/validator-job/component-test.js
```diff
@@ -280,7 +280,8 @@ module('Integration | Component | validator job', function(hooks) {
       commands: [
         { name: 'step1', command: 'sd-cmd exec bar/foo@latest' },
         { name: 'step2', command: 'sd-cmd exec foo/bar@0.0.1 foobar' },
-        { name: 'step3', command: 'sd-cmd exec bar/foo@stable' }
+        { name: 'step3', command: 'sd-cmd exec bar/foo@stable' },
+        { name: 'step4', command: 'sd-cmd exec foo/bar' }
```
yoshwata: I'd like to define some exactly same commands. Then check that there is no duplicate commands  in dom.

ibu1224: Thank you. I add test case.

app/components/validator-job/component.js
ibu1224: oh. I committed a mistake.
I fix this problem quickly.

ibu1224: I tried to move into 'while statement'.
But the system called 'eslint error Don't make functions within a loop.'
I checked 'filter method action'
As a result, it became as follows.
If matchRes is null, but commandExist return an empty array.
Therefore, I want to avoid putting it in the `while statement`.

wahapo: Oh, yeah, I was mistaken. I think that it is necessary to implement the following.
```
sdCommands.filter((command, matchRes) => { command.command === matchRes[1] ...
 });
```

wahapo: I tried it, it seems that I need to put an implementation like filter in the while statement.
In the below case, the contents of sdCommands did not become as expected.
```
commands = [
  { command: 'sd-cmd exec test/cmd1@stable; sd-cmd exec test/cmd1@stable' },
  { command: 'sd-cmd exec test/cmd1@stable; sd-cmd exec test/cmd2@latest' }
];
```

result of `sdCommands`:
```
[
  { command: 'test/cmd1', version: 'stable' },
  { command: 'test/cmd1', version: 'stable' }
]
```

ibu1224: ok I fixed.

ibu1224:
@tkyi  Please review 🙏

PR #505, opened 2019-11-08
fix(1833): event view does not convey context which triggered it

klu909:
## Context

Currently ui displays creator user name rather than commit author user name. Most of time both are the same, when event is create by another build, UI displays previous build's author rather than sd-builtbot. This lead to user confusion due to ambiguity. 

## Objective

UI now displays both committer and event creator, and in event panel correctly labels `committer` rather than `user`. And in case of event triggered by external pipeline, creator
will now link to according external build

Before:
Screen Shot 2019-11-12 at 3 19 08 PM
After:
1.when event is trigger by external pipeline
Screen Shot 2019-11-12 at 3 17 18 PM
2.when creator and committer is the same
Screen Shot 2019-11-12 at 3 13 50 PM
3.when creator and committer is different
Screen Shot 2019-11-12 at 3 16 34 PM

## References

https://github.com/screwdriver-cd/screwdriver/issues/1790

## License

### DIFF
--- a/app/components/pipeline-event-row/component.js
+++ b/app/components/pipeline-event-row/component.js
@@ -54,7 +54,7 @@ export default Component.extend({
 
   externalBuild: computed('event.{causeMessage,startFrom}', {
     get() {
-      // need to use underscores here because they used for link-to
+      // using underscore because router.js doesn't pick up camelcase
       /* eslint-disable camelcase */
       const build_id = this.get('event.causeMessage').match(/\d+$/)[0];
       const pipeline_id = this.get('event.startFrom').match(/^~sd@(\d+):[\w-]+$/)[1];
--- a/app/components/pipeline-event-row/component.js
+++ b/app/components/pipeline-event-row/component.js
@@ -37,6 +37,33 @@ export default Component.extend({
     }
   }),
 
+  isExternalTrigger: computed('event.startFrom', {
+    get() {
+      return this.get('event.startFrom').match(/^~sd@(\d+):([\w-]+)$/);
+    }
+  }),
+
+  isCommiterDifferent: computed('isExternalTrigger', 'event.{creator.name,commit.author.name}', {
+    get() {
+      const creatorName = this.get('event.creator.name');
+      const authorName = this.get('event.commit.author.name');
+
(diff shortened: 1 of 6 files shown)


adong:
@klu909 Could you attach some before/ after screenshots?

jithine:
I think we should display both Commit user and event cause user.

For example. From https://cd.screwdriver.cd/pipelines/1/events

Screen Shot 2019-11-08 at 11 47 28 AM

app/components/pipeline-event-row/component.js
```diff
@@ -37,6 +38,35 @@ export default Component.extend({
...
+  }),
+
+  externalBuildUrl: computed('event.{causeMessage,startFrom}', {
+    get() {
+      const buildId = this.get('event.causeMessage').match(/\d+$/);
+      const pipelineId = this.get('event.startFrom').match(/^~sd@(\d+):[\w-]+$/)[1];
+      let url = '#';
+
+      if (buildId && pipelineId) {
+        url = `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/pipelines/${pipelineId}/builds/${buildId}`;
```
jithine: This should be the UI end point

klu909: done

app/components/pipeline-event-row/component.js
```diff
@@ -37,6 +37,33 @@ export default Component.extend({
...
+
+  externalBuild: computed('event.{causeMessage,startFrom}', {
+    get() {
+      // need to use underscores here because they used for link-to
+      /* eslint-disable camelcase */
+      const build_id = this.get('event.causeMessage').match(/\d+$/)[0];
+      const pipeline_id = this.get('event.startFrom').match(/^~sd@(\d+):[\w-]+$/)[1];
+      /* eslint-enable camelcase */
+
+      return { build_id, pipeline_id };
```
jithine: we use camel case everywhere. so can you change this to `buildId` & `pipelineId`

jithine: sorry missed the comments.

PR #506, opened 2019-11-19
fix: ui patching when buildId is undefined for unconventional causeMessage

adong:
## Context

This PR fixing a runtime exception, and breaks the UI back button.

## Objective
This PR fix exactly that by adding check to `buildId` when using `{{link-to}}`, if `buildId` exists, it would route to pipeline with given build, otherwise will default to pipeline with given `pipelineId`

## License


vnugopal:
@jithine

PR #509, opened 2019-12-04
fix: add cause message to override during freezeWindow

adong:
## Context

UI work for PR: https://github.com/screwdriver-cd/data-schema/pull/360

## Objective

This allows users to give reason (to override causeMessage) to `force start` during the `freezeWindow` period. 
## References

image

## License

### DIFF
--- a/app/pipeline/events/controller.js
+++ b/app/pipeline/events/controller.js
@@ -277,9 +277,10 @@ export default Controller.extend(ModelReloaderMixin, {
           this.set('errorMessage', Array.isArray(e.errors) ? e.errors[0].detail : '');
         });
     },
-    startDetachedBuild(job, parameters) {
+    startDetachedBuild(job, options = {}) {
       const buildId = get(job, 'buildId');
       let parentBuildId = null;
+      const { parameters, reason } = options;
 
       if (buildId) {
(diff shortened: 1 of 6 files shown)


app/pipeline/events/controller.js
```diff
@@ -292,10 +293,14 @@ export default Controller.extend(ModelReloaderMixin, {
       const pipelineId = get(this, 'pipeline.id');
       const token = get(this, 'session.data.authenticated.token');
       const user = get(decoder(token), 'username');
-      const causeMessage = `Manually started by ${user}`;
+      let causeMessage = `Manually started by ${user}`;
       const prNum = get(event, 'prNum');
       let startFrom = get(job, 'name');
 
+      if (reason) {
```
jithine: is `reason` used only for frozen builds ? I feel it's too generic, let's say in future we allow users to enter reason for all manual starts ?

adong: Currently, the dialog only opens for `frozenBuilds`. Because we need to append the `[force start]` keyword. We can open `reason` option up, and it would require users to type `force start` correctly.

adong:
Add check if reason is null, disable `start` button if no reason given for `freezeWindow`
image

adong:
rebased with latest master

PR #511, opened 2019-12-20
fix: Display a detail page with just the major version

wahapo:
## Context
My screwdriver.cd has a `common/jq` command.
common:jq

For example, if I input the following in the validator form, the list of Commands displays `common/jq@1.5`.
validator

Clicking on the `common/jq@1.5` link will display the not found page.

## Objective
Even if only the major version is specified, the command details page can be accessed. And I fixed the same for templates page.
fix

## References
https://github.com/screwdriver-cd/ui/pull/479
https://github.com/screwdriver-cd/ui/pull/504

I fixed based on this.
https://github.com/screwdriver-cd/models/blob/b88a006b832cfafa96f1114b1149055cc108f1b7/lib/commandFactory.js#L136-L153

## License

### DIFF
--- a/app/commands/detail/route.js
+++ b/app/commands/detail/route.js
@@ -1,6 +1,7 @@
 import { inject as service } from '@ember/service';
 import RSVP from 'rsvp';
 import Route from '@ember/routing/route';
+import { compareVersions } from 'screwdriver-ui/helpers/compare-versions';
 
 export default Route.extend({
   command: service(),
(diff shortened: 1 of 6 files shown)


app/commands/detail/route.js
```diff
@@ -10,14 +10,39 @@ export default Route.extend({
...
+        const versionExists = verPayload.filter(t => t.version.startsWith(params.version));
         const tagExists = tagPayload.filter(c => c.tag === params.version);
 
         if (tagExists.length === 0 && versionExists.length === 0) {
           this.transitionTo('/404');
         }
+
+        if (versionExists.length > 0) {
+          // Sort commands by descending order
+          versionExists.sort((a, b) => {
```
tkyi: can this be put into a helper?
https://github.com/screwdriver-cd/ui/tree/master/app/helpers

wahapo: Thanks, I fixed it.

PR #517, opened 2020-01-23
fix(1903): split active step inside artifacts

adong:
## Context

Current implementation is that when we land to artifacts route, it will redirect to `/pipeline/1/step/` route if we encountered failure step.

## Objective

This PR will prevent the redirection if users are on artifacts route.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1903

## License


adong:
> do we already a test covering this ?

Yes, and it cause tests to fail. I'm fixing the tests.

PR #519, opened 2020-01-27
Add UTC time as a log format option.

jacobtolar:
## Context

It's often useful to see the time in UTC instead of local time.

## Objective
Adds another item in the rotation when clicking through the various timestamp formats to show the event time in UTC.

## References

https://github.com/stefanpenner/ember-moment#common-optional-named-arguments

```
timeZone | An optional String time zone, defaults to moment.timeZone (the default time zone)
```

## Tests
I don't see a great way to test this. Can the UI be started up with static data or do I need to set up a complete local cluster to verify this change?

## License

### DIFF
--- a/app/components/build-log/component.js
+++ b/app/components/build-log/component.js
@@ -5,7 +5,7 @@ import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import ENV from 'screwdriver-ui/config/environment';
 
-const timeTypes = ['datetime', 'elapsedBuild', 'elapsedStep'];
+const timeTypes = ['datetime', 'datetimeUTC', 'elapsedBuild', 'elapsedStep'];
 
 export default Component.extend({
   logService: service('build-logs'),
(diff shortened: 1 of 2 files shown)


tkyi:
We're working on adding docs to help developers get setup with their local dev env
https://github.com/screwdriver-cd/guide/pull/349/files?short_path=1bda71c#diff-1bda71ce1b5ff33fb6694abb0e43857d

jithine:
cc @adong

adong:
😤Oops, I made a mistake and merged without following `semantic-release`
@jacobtolar Please take a look at git commit message formats.

https://docs.screwdriver.cd/about/contributing/index#submitting-pull-requests

This PR changes will be reflected in the next release (another PR merges with semantic-release commit format).

jacobtolar:
Well, this didn't quite work as expected (the time is shown in my local timezone still when I select UTC Time). perhaps the object passed to momentjs isn't timezone aware or moment-format isn't doing what I thought it would. will play around with it a bit more when I get a chance

PR #521, opened 2020-01-28
fix(1925): support branch names with number sign

wahapo:
## Context
This PR is for this Issue.
https://github.com/screwdriver-cd/screwdriver/issues/1925

Number sign of the branch name must be encoded to `%23`.

**Current**

**After**

Link to below page.

## License

### DIFF
--- a/app/helpers/branch-url-encode.js
+++ b/app/helpers/branch-url-encode.js
@@ -1,0 +1,17 @@
+import { helper } from '@ember/component/helper';
+
+/**
+ * Encode name of branch in url for number sign
+ * @method branchUrlEncode
+ * @param  {Array}     url      url
+ * @return {String}             encoded url
+ */
+export function branchUrlEncode([url]) {
+  if (!url) {
+    return url;
+  }
+
+  return url.replace(/#/g, '%23');
+}
+
+export default helper(branchUrlEncode);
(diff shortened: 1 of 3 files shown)


tests/integration/helpers/branch-url-encode-test.js
```diff
@@ -0,0 +1,22 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';
+
+module('helper:branch-url-encode', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders a duration given two parsable times in HH:mm:ss format', async function(assert) {
```
ibu1224: The description is wrong?

app/helpers/branch-url-encode.js
```diff
@@ -0,0 +1,17 @@
...
+ * @method branchUrlEncode
+ * @param  {Array}     url      url
+ * @return {String}             encoded url
+ */
+export function branchUrlEncode([url]) {
+  if (!url) {
+    return url;
+  }
+
+  return url.replace(/#/g, '%23');
```
wahapo: Only the `#` in the branch name needs to be converted.
In the `url` shown here, the anchor `#` is not used, so we should be able to replace all `#` without any problems.

tests/integration/helpers/branch-url-encode-test.js
wahapo: thanks, I fixed.

yoshwata:
[ask] Does this PR works properly when pipeline uses SourceDirectory?
https://docs.screwdriver.cd/user-guide/configuration/sourceDirectory.html

And how it displayed as branch name?

wahapo:
@yoshwata 
Also in that case, properly build was executed. And the link seems to be as expected.

Link to the GitHub page below.

PR #526, opened 2020-02-05
fix(1910):Remove default pipelines from 'Add to Collection' list

klu909:
## Context

When trying to add a pipeline to a collection, the default collection (My Pipelines) shows up as an option. When trying to click that option, I get an error Could not add Pipeline to `My Pipelines`.

## Objective

Since users should not be allowed to modify the default collection, that option should not show up to the Add to Collection list.

## References

https://github.com/screwdriver-cd/screwdriver/issues/1910

## License

### DIFF
--- a/app/components/collection-dropdown/component.js
+++ b/app/components/collection-dropdown/component.js
@@ -1,13 +1,22 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
 
 export default Component.extend({
   classNames: ['collection-dropdown'],
   showCollectionModal: false,
+  collections: null,
   addCollectionError: null,
   addCollectionSuccess: null,
   pipeline: null,
   buttonText: undefined,
   buttonClass: undefined,
+  filteredCollections: computed('collections', {
+    get() {
+      const { collections } = this;
+
+      return collections.filter(collection => collection.type !== 'default');
+    }
+  }),
   actions: {
     openModal() {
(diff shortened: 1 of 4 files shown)


app/components/collection-dropdown/component.js
```diff
@@ -1,13 +1,22 @@
...
   addCollectionError: null,
   addCollectionSuccess: null,
   pipeline: null,
   buttonText: undefined,
   buttonClass: undefined,
+  filteredCollections: computed('collections', {
+    get() {
+      const { collections } = this;
+
+      return collections.filter(collection => collection.type !== 'default');
```
adong: How about?

```javascript
import { filter } from '@ember/object/computed';

filteredCollections: filter('collection.[]', function(collection) {
  return collection.type !== 'default';
})

```

Ref: 

Filter: https://api.emberjs.com/ember/3.9/functions/@ember%2Fobject%2Fcomputed/filter
Aggregation `.[]`: https://guides.emberjs.com/release/object-model/computed-properties-and-aggregate-data/

adong: Another suggestion is that: change the generic variable name from `filteredCollections` to `addiableCollections` or `collectiableCollections`? Not sure though, 🤷‍♂ 

I think you might come up with another name.

klu909: changed to `normalCollections`

PR #529, opened 2020-02-12
fix(1963): fix redirect route to step route on all build route redirections

adong:
## Context

If users visit a failed build with artifacts link directly, then users should not be redirected to the failed step of the build.

Please see issue: https://github.com/screwdriver-cd/screwdriver/issues/1963

## Objective

This PR adds check only redirect to step route for 'pipeline.build.step', 'pipeline.build.index', which will NOT redirect users when their intention is to visit the artifacts directly.

## References

issue: https://github.com/screwdriver-cd/screwdriver/issues/1963
FYI: 
I always refer to this doc :) 
Ember Route: https://alexdiliberto.com/posts/ember-route-hooks-a-complete-look/

## License

### DIFF
--- a/app/pipeline/build/route.js
+++ b/app/pipeline/build/route.js
@@ -58,16 +58,18 @@ export default Route.extend({
     }
   },
 
-  redirect(model /* , transition */) {
-    const name = getActiveStep(get(model, 'build.steps'));
+  redirect(model, transition) {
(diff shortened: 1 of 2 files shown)


adong:
> LGTM, is there a test in place for this redirect too?

Indeed, added test. Good call. Also, I think this is one example of using `sinon.spy` https://sinonjs.org/releases/v7.1.0/spies/

PR #530, opened 2020-02-13
fix(1937): Add scm login link to warning message on scm mismatch

yoshwata:
## Context

Fix https://github.com/screwdriver-cd/screwdriver/issues/1937

## Objective

Like @wahapo says on his issue, put SCM login link when context mismatch occurs is user friendly. The deliverables is below.

fix

I'm sorry that I can not get scmName and scmIcon from pipeline's model, it is not an object but string.

## References

https://github.com/screwdriver-cd/screwdriver/pull/1966 

## License


app/components/info-message-scm/template.hbs
```diff
@@ -0,0 +1,6 @@
+{{#if message}}
+  {{#bs-alert onDismissed=(action "clearMessage") type=type}}
+    {{! template-lint-disable no-triple-curlies }}
+    {{fa-icon icon}} <span>{{{message}}}</span> Please login <a href="#authenticate" {{action "authenticate" scmContext}}>{{scmContext}}</a>
```
tkyi:
```suggestion
    {{fa-icon icon}} <span>{{{message}}}</span> Please login to <a href="#authenticate" {{action "authenticate" scmContext}}>{{scmContext}}</a>
```

yoshwata: Thank you

app/pipeline/events/template.hbs
```diff
@@ -4,7 +4,12 @@
   {{/if}}
 {{/if}}
 
-{{info-message message=errorMessage type="warning" icon="exclamation-triangle"}}
+{{#if (eq errorMessage "This checkoutUrl is not supported for your current login host.")}}
```
adong: I'd advise that we create a variable inside `app/pipeline/events/component.hbs` to handle string comparisons, rather than in handlebars.

1) computed property changes only update corresponding handlebar changes (more performant)
2) easy to track the number of errors if we have more types of errors to come.

```
  scmError: computed('errorMessage', {
    get() {
      return this.errorMessage === "This checkoutUrl is not supported for your current login host.") {
    }
  });
```

By the way, 2 things I don't fully grasp

1) Where the magic string comes from in the backend.
2) Could we reuse component `info-message` without creating `info-message-scm`.

yoshwata: > Where the magic string comes from in the backend.

It comes from API when builds started, pipeline repository changed or pipeline created as long as I know. I opened [PR](https://github.com/screwdriver-cd/screwdriver/pull/1966/files) for build started case.

> Could we reuse component info-message without creating info-message-scm

Yes, I integrated them

yoshwata: > > I'd advise that we create a variable inside app/pipeline/events/component.js

I tried to make component which change the state from `component.js`, but I can not render `authenticate` action dynamically from js.

yoshwata:
I make change to show change/create pipeline page properly. 
create

PR #528, opened 2020-02-07
fix(1950): refactor fetchLogs api call

adong:
## Context

Minor UI optimization

## Objective

- increase `LOG_RELOAD_TIMER` from `1` to `3`s
- refactored `fetchLog` to `shuttle` api service
- add tests for `shuttle` service

- upgrade to `node:12` for https://github.com/screwdriver-cd/screwdriver/issues/1722

## References

Issue: https://github.com/screwdriver-cd/screwdriver/issues/1950

Ember Data / Ember Ajax / Ember Fetch proposal
https://gist.github.com/tomdale/203ea125acd0be8b5f1f

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #534, opened 2020-02-27
fix: route from pipeline/build page to any route will redirect within timeout

adong:
## Context

We had a PR: https://github.com/screwdriver-cd/ui/pull/528

That increased the reload time from `1s` to `3s`:
https://github.com/screwdriver-cd/ui/pull/528/files#diff-695693c5dabaffb064bde5b060195ff7L52

Which effects `timeout` here 
https://github.com/screwdriver-cd/ui/blob/1b8b32deb80a2c1744d1d61ff28de910d94a12c1/app/pipeline/build/controller.js#L116-L123

Cause UI to behave that: if users are on pipeline/build page, and go back to any route. Within the timeout period, users will be redirected back. This behavior was not intended, and was not visible when the timeout was set to `1s`. 

But this undesired behavior is visible now, and we need to address it. Hence this PR. Credits to @klu909 who spotted this. 

## Objective

This will only trigger `changeBuildStep` if it's from the desired routes, such as `'pipeline.build.step', 'pipeline.build.index'`. All other routes will not trigger build. 

## References

[Ember Router Service](https://api.emberjs.com/ember/3.12/classes/RouterService/properties/currentRoute?anchor=currentRoute)
## License

### DIFF
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -9,6 +9,7 @@ import ENV from 'screwdriver-ui/config/environment';
 import { getActiveStep } from 'screwdriver-ui/utils/build';
 
 export default Controller.extend({
+  router: service(),
   prEventsService: service('pr-events'),
   session: service('session'),
   loading: false,
@@ -127,6 +128,12 @@ export default Controller.extend({
   },
 
   changeBuildStep(name) {
+    const currentRouteName = this.router.currentRoute.name;
+
+    if (!['pipeline.build.step', 'pipeline.build.index'].includes(currentRouteName)) {
+      return;
+    }
+
     const build = this.get('build');
     const pipelineId = this.get('pipeline.id');
     let activeStep;
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -104,13 +104,12 @@ export default Controller.extend({
 
     // reload again in a little bit if queued
     if (!this.loading) {
-      if (status === 'QUEUED' || status === 'RUNNING') {
+      if (['QUEUED', 'RUNNING'].includes(status)) {
         later(
           this,
           () => {
             if (!build.get('isDeleted') && !this.loading) {
               this.set('loading', true);
-
               build.reload().then(() => {
                 this.set('loading', false);
                 throttle(this, 'reloadBuild', timeout);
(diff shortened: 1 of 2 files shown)


adong:
> shouldn't we stop it here
> 
> https://github.com/screwdriver-cd/ui/blob/1b8b32deb80a2c1744d1d61ff28de910d94a12c1/app/pipeline/build/controller.js#L114-L116
> 
> ?
> else won't the `reload` be still active, when it's not needed ?

Yeah, that would make sense. I tried that,

```diff
diff --git a/app/pipeline/build/controller.js b/app/pipeline/build/controller.js
index 4a1b3f7..cc77cc7 100644
--- a/app/pipeline/build/controller.js
+++ b/app/pipeline/build/controller.js
@@ -3,7 +3,6 @@ import { get, computed } from '@ember/object';
 import { reads, mapBy } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { jwt_decode as decoder } from 'ember-cli-jwt-decode';
 import Controller from '@ember/controller';
 import ENV from 'screwdriver-ui/config/environment';
 import { getActiveStep } from 'screwdriver-ui/utils/build';
@@ -104,12 +103,19 @@ export default Controller.extend({
 
     // reload again in a little bit if queued
     if (!this.loading) {
-      if (status === 'QUEUED' || status === 'RUNNING') {
+      if (['QUEUED', 'RUNNING'].includes(status)) {
         later(
           this,
           () => {
             if (!build.get('isDeleted') && !this.loading) {
               this.set('loading', true);
+              const currentRouteName = this.getWithDefault('router.currentRoute.name', '');
+
+              if (!['pipeline.build.step', 'pipeline.build.index'].includes(currentRouteName)) {
+                console.log('stop reloading');
+
+                return;
+              }
 
               build.reload().then(() => {
                 this.set('loading', false);
@@ -128,12 +134,6 @@ export default Controller.extend({
   },
 
   changeBuildStep(name) {
-    const currentRouteName = this.router.currentRoute.name;
-
-    if (!['pipeline.build.step', 'pipeline.build.index'].includes(currentRouteName)) {
-      return;
-    }
-
     const build = this.get('build');
     const pipelineId = this.get('pipeline.id');
     let activeStep;
```

and it will not update the event workflow 

image

Nor top-right corner mini graph though.

image

We are still using the event data, just not to redirect the page 😐

adong:
PS, how I create that diff preview is done by the following:

```
git diff ui/file/Path/file.js > a.txt
copy the content and create a code block with diff
    ```diff
    // DIFF CONTENT
    ```

```

adong:
Fixing the test cases by mocking `service:router`

```
    this.owner.unregister('service:router');
    this.owner.register('service:router', routerServiceMock);
```

Ref:  https://github.com/ember-cli/ember-cli-qunit/issues/203#issuecomment-592402777

PR #538, opened 2020-03-09
fix(1977): The pipeline admin deletes template/command

ibu1224:
## Context
I fix [issue](https://github.com/screwdriver-cd/screwdriver/issues/1977).
Currently, template / command can only be deleted by cluster admin.
It is inconvenient because the admin of the template/command cannot be deleted.
Therefore, I modify it so that the pipeline admin can  be deleted.

## Objective
Modify non-cluster admins can click on the trash icon.
If anyone other than the cluster admin and the pipeline admin tries to delete, an err message will be displayed.

## References
https://github.com/screwdriver-cd/screwdriver/issues/1977

## License

### DIFF
--- a/app/components/command-header/template.hbs
+++ b/app/components/command-header/template.hbs
@@ -6,8 +6,8 @@
   {{else}}
     {{fa-icon "code-fork" title="The pipeline for this command does not exist."}}
   {{/if}}
+  <a href="#" {{action "setCommandToRemove" command}} class="remove">{{fa-icon "trash" title="Delete command"}}</a>
   {{#if isAdmin}}
-    <a href="#" {{action "setCommandToRemove" command}} class="remove">{{fa-icon "trash" title="Delete command"}}</a>
     {{x-toggle
       size="medium"
       theme="light"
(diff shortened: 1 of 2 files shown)


app/components/command-header/template.hbs
```diff
@@ -17,8 +17,6 @@
       onLabel="Trust?"
       onToggle=(action "updateTrust" command.namespace command.name)
     }}
-  {{else}}
-    {{fa-icon "trash" title="Cannot delete command; pipeline could not be found."}}
```
kumada626: I think that it is better not to show the icon when user is not a cluster admin and the pipeline is not be found.

ibu1224: OK, I fix it.

app/components/command-header/template.hbs
```diff
@@ -18,7 +18,11 @@
       onToggle=(action "updateTrust" command.namespace command.name)
     }}
   {{else}}
-    {{fa-icon "trash" title="Cannot delete command; pipeline could not be found."}}
+    {{#if scmUrl}}
```
adong: I'm curious why it is an `if` condition on `scmUrl`. Is it intended to check if users are the authors of the given templates/commands?

ibu1224: @adong review thank you.
Here checking if pipeline exists.
For `user` authority If not found pipeline, it is not really good to display the trash icon.

PR #542, opened 2020-03-30
fix: fix build log display

kumada626:
## Context

Because `.wrap`'s height is `vh70`, bottom space of build log still remain like below.
bottom_space

Fill the build log to the bottom.

After:
after_space

## Objective

Fill the build log to the bottom.

## References

#541 

## License

### DIFF
--- a/app/components/build-log/styles.scss
+++ b/app/components/build-log/styles.scss
@@ -34,7 +34,7 @@
 }
 
 .wrap {
-  height: auto;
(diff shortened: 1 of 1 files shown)


adong:
`grid-template-rows` is relatively new, glad to see we are adopting it and make our life easier. 

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows

PR #556, opened 2020-06-01
fix(1559): Allow float for coverage meta

tkyi:
## Context

Users might need more granularity for their code coverage percentages. Currently, we filter to only display non-decimal coverage.

## Objective

This PR checks coverage is an int, then coverts to decimal to display.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1559

## License


tests/integration/components/build-banner/component-test.js
```diff
@@ -440,7 +440,7 @@ module('Integration | Component | build banner', function(hooks) {
     }}`);
 
     return settled().then(() => {
-      assert.dom('.coverage .banner-value').hasText('100%');
+      assert.dom('.coverage .banner-value').hasText('100.00%');
```
jithine, line 443: is it possible to omit `.00` ?

PR #559, opened 2020-06-15
fix(1967): play button should create a fresh build.

jithine:
## Context

1. disable  `skip further` until UI can ask for user choice
1. create fresh build for play button
1. fix restart build interaction 

## License


jithine:
Moved it to draft, since restart case it not working as expected

PR #567, opened 2020-07-17
fix(2152): Pass in pipelineId and prNum to Sonar coverage [2]

tkyi:
## Context

Need to pass in pipelineId and prNum for new Sonar coverage calls.

## Objective

This PR passes in pipelineId and prNum to get Sonar coverage info.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2152

## License

### DIFF
--- a/app/components/pipeline-list-view/component.js
+++ b/app/components/pipeline-list-view/component.js
@@ -112,6 +112,9 @@ export default Component.extend({
         stopBuild: this.get('stopBuild')
       };
 
+      const prRegex = /^PR-(\d+)(?::([\w-]+))?$/;
+      const prNumMatch = jobName.match(prRegex);
+
       let duration;
       let startTime;
       let status;
@@ -128,7 +131,9 @@ export default Component.extend({
           jobId,
           buildId,
           startTime: latestBuild.startTime,
-          endTime: latestBuild.endTime
+          endTime: latestBuild.endTime,
+          pipelineId: latestBuild.pipelineId,
+          prNum: prNumMatch ? prNumMatch[1] : null
         };
       }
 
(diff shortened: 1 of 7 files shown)


app/components/pipeline-list-view/component.js
```diff
@@ -128,7 +131,9 @@ export default Component.extend({
           jobId,
           buildId,
           startTime: latestBuild.startTime,
-          endTime: latestBuild.endTime
+          endTime: latestBuild.endTime,
+          pipelineId: latestBuild.pipelineId,
+          prNum: prNumMatch ? prNumMatch[1] : null
```
pritamstyz4ever, line 136: should we have a length check here as we are getting the second index?

tkyi: updated

app/components/pipeline-list-view/component.js
```diff
@@ -128,7 +131,9 @@ export default Component.extend({
           jobId,
           buildId,
           startTime: latestBuild.startTime,
-          endTime: latestBuild.endTime
+          endTime: latestBuild.endTime,
+          pipelineId: latestBuild.pipelineId,
+          prNum: prNumMatch && prNumMatch.length > 1 ? prNumMatch[1] : null
```
adong, line 136: Just want to point out some cool features.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining


```
prNumMatch?.[1] ?? null
```

jithine: Cool feature. Reduces boiler plate a lot. For ember it doesn't matter, it always uses babel.  But we should use it in api and other components after upgdating to node:12

PR #568, opened 2020-07-20
feat(1079): add option for auto deploy key generation [6]

supra08:
## Context

Fixes screwdriver-cd/screwdriver#1079

## Objective

This PR adds a checkbox for auto keys generation.

## References

screwdriver-cd/screwdriver#1079

## License


app/components/pipeline-create-form/component.js
```diff
@@ -30,6 +32,14 @@ export default Component.extend({
     }
   }),
 
+  hasAutoDeployEnabled: computed({
+    get() {
+      const scm = this.scmService.getScm('github:github.com');
```
adong, line 37: Since we intend to support 3 SCMs: `github`, `gitlab`, and `bitbucket` and be agnostic, we can use

https://github.com/screwdriver-cd/ui/blob/32d16fa559dac66f8f064a8e45d5a557a0b75a18/app/application/controller.js#L22

```
session: service(),



const currentContext = session.get('data.authenticated.scmContext');
```

supra08: sure! I'll get it fixed

app/pipeline/serializer.js
```diff
@@ -30,7 +30,8 @@ export default DS.RESTSerializer.extend({
   serializeIntoHash(hash, typeClass, snapshot) {
     return assign(hash, {
       checkoutUrl: snapshot.attr('checkoutUrl'),
-      rootDir: snapshot.attr('rootDir') || ''
+      rootDir: snapshot.attr('rootDir') || '',
+      autoKeysGeneration: snapshot.attr('autoKeysGeneration') || false
```
adong, line 34: Should this default to `''` empty string?

app/pipeline/model.js
```diff
@@ -7,6 +7,7 @@ export default DS.Model.extend({
   annotations: DS.attr(),
   checkoutUrl: DS.attr('string'),
   rootDir: DS.attr('string'),
+  autoKeysGeneration: DS.attr('string'),
```
adong, line 10: From: https://github.com/screwdriver-cd/ui/pull/568/files#diff-6b06326e0cb4622d4b9dfc45401cd687R34
```
      autoKeysGeneration: snapshot.attr('autoKeysGeneration') || false
```

Does `autoKeysGeneration: false` imply `DS.attr('boolean')` 🤔 ?

I see `autoDeployKeyGeneration` is `DS.attr('boolean')` already, what is the difference between `autoDeployKeyGeneration` vs `autoKeysGeneration`?

To me `autoKeysGeneration` is the expected payload that API expects, while `autoDeployKeyGeneration` is the actual string from user's input?

tests/unit/pipeline/serializer-test.js
```diff
@@ -130,7 +130,8 @@ module('Unit | Serializer | pipeline', function(hooks) {
 
       assert.deepEqual(payload, {
         checkoutUrl: 'git@example.com:foo/bar.git',
-        rootDir: ''
+        rootDir: '',
+        autoKeysGeneration: false
```
adong, line 134: If we change the default value from `false` to `''`, also need to update test here :)

app/pipeline/model.js
supra08: Oh my bad, it should be boolean only. 
Actually from the UI and API side, the flag dealing with this is `autoKeysGeneration` only. `autoDeployKeysGeneration` is the flag that the user sets in the config and what is fetched in the scm modules after parsing the config.

adong:
LGTM. Could you rebase from master again? Will approve the PR

PR #569, opened 2020-07-22
fix(2124):  Pipeline list view issues

adong:
## Context

Issues from https://github.com/screwdriver-cd/screwdriver/issues/2124

## Objective

- [x] Clicking on the “play” button doesn’t bring up the build parameter modal if it’s enabled.

- [x] If you are on a list view, then click collection and click on another pipeline, it will land on the list view again but with jobs from the previously selected pipeline as well.

- [x] The clickable buttons should be using different cursor when hover.

- [ ] If you click any of the sortable column header, it seems it will query /v4/coverage/info for each job.

## References

- [Using the `wait` helper in Ember integration tests](https://til.hashrocket.com/posts/851fb1e6f7-using-the-wait-helper-in-ember-integration-tests)

## License

### DIFF
--- a/app/components/pipeline-list-view/component.js
+++ b/app/components/pipeline-list-view/component.js
@@ -56,6 +56,10 @@ export default Component.extend({
     this.set('table', table);
   },
 
+  didDestroyElement() {
+    this.set('jobsDetails', []);
+  },
+
   getDuration(startTime, endTime, status) {
     const startDateTime = Date.parse(startTime);
     const endDateTime = Date.parse(endTime);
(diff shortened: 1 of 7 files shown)


app/components/pipeline-list-view/component.js
```diff
@@ -53,7 +54,33 @@ export default Component.extend({
...
+
+  /**
+   * Note: jobId must be a string
+   * @param  {integer} jobId
+   * @return {undefined}
+   */
+  openParametersModal(jobId) {
+    const job = this.jobs.findBy('id', `${jobId}`);
+
+    this.setProperties({
```
jithine, line 80: where are the default parameters being fetched? Is behavior similar to the modal interactions happening in event view ?

adong: Yeah. parameters are being fetched from route event

PR #573, opened 2020-07-29
feat(2092): [1] Dropdown build parameters

yoshwata:
## Context

Add feature of https://github.com/screwdriver-cd/screwdriver/issues/2092

## Objective

Add feature of dropdown build parameters.

### yaml example:
```
parameters:
  resource: ["commands/publish", "templates/mock", "templates/tito"]
  from: "latest"
  to: ["test", "stable"]
    
jobs:
  main:
    requires: [~pr, ~commit]
    image: node:12
    steps:
      - show1: 'echo "Region: $(meta get parameters.resource.value)"'
      - show2: 'echo "Region: $(meta get parameters.from.value)"'
      - show3: 'echo "Region: $(meta get parameters.to.value)"'
```

## References

https://github.com/screwdriver-cd/screwdriver/issues/2092

## License

### DIFF
--- a/app/components/pipeline-parameterized-build/component.js
+++ b/app/components/pipeline-parameterized-build/component.js
@@ -88,12 +88,14 @@ export default Component.extend({
       const value = propertyVal.value || propertyVal || '';
       const description = propertyVal.description || '';
       // value to update selected parameter
-      let listSelected = { name: value[0] };
+      let listSelected = {};
 
       const selectParameters = [];
 
       if (Array.isArray(value)) {
         // initialize selected list
+        listSelected = { name: value[0] };
+        // initialize selected text
         normalizedParameterizedModel[propertyName] = value[0];
         if (this.parameterText[propertyName]) {
           // when already selected
--- a/app/components/pipeline-parameterized-build/component.js
+++ b/app/components/pipeline-parameterized-build/component.js
@@ -26,8 +26,16 @@ export default Component.extend({
    */
   buildParameters: {},
 
+  /**
+   * parameterText are expected to be an object consists of key value pairs
+   * @type {Object}
+   */
   parameterText: {},
 
+  /**
+   * parameterSelect are expected to be an object consists of key value pairs
+   * @type {Object}
+   */
   parameterSelect: {},
 
   /**
--- a/app/components/pipeline-parameterized-build/component.js
+++ b/app/components/pipeline-parameterized-build/component.js
@@ -26,6 +26,10 @@ export default Component.extend({
    */
   buildParameters: {},
 
+  parameterText: {},
+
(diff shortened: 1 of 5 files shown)


wahapo:
What kind of yaml would you feed it to show as shown in the diagram? I'd like an example for reference.

yoshwata:
> What kind of yaml would you feed it to show as shown in the diagram?

I added it.

wahapo:
I think it's more intuitive to use [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist).

yoshwata:
> Thanks! I think it's more intuitive to use datalist.

It's what I tried to implement but I could not find the way how to implement in ember.

What I tried:
https://stackoverflow.com/questions/18724155/can-a-datalist-be-used-with-ember-js-select)
https://ember-basic-dropdown.com/docs/trigger-events (The bottommost example)

These are not worked for me and it seems to be difficult so I give up.

app/components/pipeline-parameterized-build/component.js
```diff
@@ -65,21 +80,61 @@ export default Component.extend({
...
-  normalizeParameters(parameters = {}) {
+  normalizeParameters(parameters = {}, selectedValue) {
     const normalizedParameters = [];
     const normalizedParameterizedModel = {};
 
     Object.entries(parameters).forEach(([propertyName, propertyVal]) => {
       const value = propertyVal.value || propertyVal || '';
       const description = propertyVal.description || '';
+      // value to update selected parameter
+      let listSelected = { name: value[0] };
```
ibu1224, line 91: What happens in the following cases?

```
const value =  ''
let listSelected = { name: value[0] };
```

yoshwata: Nothing is happen because it is not referred for `''` case, but it should be checked. I fixed it.

adong:
LOL. I like the dropdown idea and did another implementation (tests pending). 
Please take a look at `https://github.com/screwdriver-cd/ui/compare/adong/parameter-dropdown?expand=1`

image

yoshwata:
@adong Thank you. I checked, but it seems that dropdown has no ability to receive input text. I think [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) is best. Do you have ideas to do that.

adong:
Hmm, I thought it would be a dropdown with predefined selections, if we allow dropdown + free form, `how about this`?




Code: `https://github.com/screwdriver-cd/ui/pull/new/adong/parameter-dropdown-override`

yoshwata:
https://github.com/screwdriver-cd/ui/pull/575 is better

PR #574, opened 2020-07-30
fix(1826): Clickable URLs in console output

adong:
## Context

For issue: https://github.com/screwdriver-cd/screwdriver/issues/1826, user reports that prefer to have text could be hyperlinked to be a clickable anchor.

## Objective

This PR only transform `http` and `https` text to be an anchor

Test link: https://cd.screwdriver.cd/pipelines/1/builds/422397/steps/publish-npm-and-git-tag

Before: 
image

After: 
image

## References

In order to use [`ansi_up`](https://github.com/drudru/ansi_up), we need to use the following configs, see [Does rendering URL requires escape_for_html set to false?](https://www.gitmemory.com/issue/drudru/ansi_up/63/574600782)

```
ansiUp.escape_for_html = false;
ansiUp.url_whitelist = { http: 1, https: 1 };
```

For backtracking purposes, we need to be mindful of the old implementations.

- [screwdriver-cd/ui@d5b151f](https://github.com/screwdriver-cd/ui/commit/d5b151f)
- [screwdriver-cd/ui@af449a6](https://github.com/screwdriver-cd/ui/commit/af449a6)

API Reference:

- [Ember htmlSafe](https://api.emberjs.com/ember/3.10/functions/@ember%2Ftemplate/htmlSafe)
This is telling ember that developer knows given string is html safe, and ember will render as html without using triple braces`{{{ }}}`

- [Handlebars.Utils.escapeExpression(string)](https://handlebarsjs.com/api-reference/utilities.html#handlebars-escapeexpression-string)
This will escape string to be html safe

## License


app/components/build-log/component.js
```diff
@@ -7,6 +7,39 @@ import ENV from 'screwdriver-ui/config/environment';
...
+    .join(' ');
+}
+
+/**
+ * [applyHyperLinkToLogs description]
+ * @param  {Array}  logs [description]
+ * @return {[type]}      [description]
+ */
+function applyHyperLinkToLogs(logs = []) {
+  logs.forEach(log => {
```
jithine, line 38: this is what we are trying to avoid with https://github.com/screwdriver-cd/screwdriver/issues/2068

cc @klu909

adong: UI would be slow even if we could wrap this code in a web worker. 

In that case, maybe we can try to use a similar code in the backend to extract regex, and replace the text with anchor tags. Then UI can render as is.

jithine: it might be faster to apply the regex on the log as a whole page, rather than individual lines.

DekusDenial:
you could make it so that the pattern matching happens when the line is hovered

PR #576, opened 2020-07-31
fix(1684): update ember to 3.16

wahapo:
## Context
Fix this issue https://github.com/screwdriver-cd/screwdriver/issues/1684.

We can reproduce it by entering the following yaml into [validator](https://cd.screwdriver.cd/validator).
```
jobs:
  main:
    image: node:10
    steps:
      - ls
    annotations:
      sd.cd: 1 
```

## Objective
This PR will fix it, as shown in the screenshot below.

It's because of this bug.
https://github.com/emberjs/ember.js/issues/17529

We can fix it by updating to at least [v3.13.0](https://github.com/emberjs/ember.js/releases/tag/v3.13.0). I would suggest giving it up to a version of [LTS](https://emberjs.com/releases/lts).

## References
https://github.com/screwdriver-cd/screwdriver/issues/1684
https://emberjs.com/releases/lts
https://github.com/offirgolan/ember-light-table/pull/701
https://deprecations.emberjs.com/v3.x/#toc_jquery-apis

## License


jithine:
Looks like few tests failed. 

cc @adong

package.json
```diff
@@ -69,7 +69,7 @@
     "ember-ace": "^2.0.1",
     "ember-ajax": "^5.0.0",
     "ember-bootstrap": "^2.8.1",
-    "ember-cli": "~3.10.1",
+    "ember-cli": "^3.20.0",
```
jithine, line 72: if we are planning LTD shouldn't it be 3.16 ? https://github.com/emberjs/ember.js/blob/v3.16.6/CHANGELOG.md

adong:
Nice. I like `3.16`, which is the latest LTS https://emberjs.com/releases/lts

```bash
cd screwdriver-cd/ui

# optional and important, some dependencies may not in enterprise npm registry yet
# npm config set registry https://registry.npmjs.org/

npm install -g ember-cli-update
ember-cli-update --to 3.16.0
```

Fix eslint styles:

```
node ./node_modules/eslint/bin/eslint.js --fix .
```

Thank you for taking a stab at it 👍

Refer: 
1) https://github.com/ember-cli/ember-cli-update has more details.

.travis.yml
```diff
@@ -0,0 +1,27 @@
+---
```
jithine, line 1: why do we ever need this ?

wahapo: Probably not necessary. I'll delete this file.

wahapo:
I feel like [this stub](https://github.com/screwdriver-cd/ui/blob/32d16fa559dac66f8f064a8e45d5a557a0b75a18/tests/unit/application/route-test.js#L34) is not being replaced correctly. 🤔

adong:
> I feel like [this stub](https://github.com/screwdriver-cd/ui/blob/32d16fa559dac66f8f064a8e45d5a557a0b75a18/tests/unit/application/route-test.js#L34) is not being replaced correctly. 🤔

[ember-sinon-quint](https://github.com/elwayman02/ember-sinon-qunit) has a newer version for ember 3.16: https://github.com/elwayman02/ember-sinon-qunit/releases/tag/v5.0.0

To install the newer version, we need to upgrade `package.json`:

```javascript
    "ember-sinon": "^5.0.0",
    "ember-sinon-qunit": "^5.0.0",
```

Also we need to have the following for `tests/test-helper.js`:

```javascript
import { setApplication } from '@ember/test-helpers'; 
import { start } from 'ember-qunit'; 
import Application from '../app'; 
import config from '../config/environment'; 
import setupSinon from 'ember-sinon-qunit';
 
setApplication(Application.create(config.APP)); 
 
setupSinon();
 
start(); 
```

adong:
I also noticed that 
```
  result: computed('value', {
    async get() {
```
in `app/components/pipeline-list-coverage-cell/component.js` needs to be changed to

```
  result: computed('value', {
    get: async function() {
```

I tested on [babel](https://babeljs.io/en/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDA3gKAJBwgTzMGAcwFMoAKASlRjRhgwwCdSBXRsGMYgdxgAVGIALYBLCMQB0zCCAA2AN2JkA5MooBuWvQC-AGjRaMAeiNFSALni58MAGYt8UEeEqpDJmMyhsOXXgOExSWk5RRU1TTo6D206NG0gA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Ces2016%2Ces2017%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose%2Cenv&prettier=false&targets=&version=7.11.1&externalPlugins=babel-plugin-ember-property-computed%401.0.1)

image

It seems the first way of written was never supported. 🤷‍♂️

adong:
Looks like `ember-sinon-quit` has deprecated `https://github.com/elwayman02/ember-sinon-qunit#deprecated-features` since https://github.com/elwayman02/ember-sinon-qunit/tree/aab1b20c198f104bd02e6fd4c1b610e5c7f88f30

image

wahapo:
Cool! That's exactly what I was looking for right now.

package.json
```diff
@@ -90,35 +96,37 @@
     "ember-cli-template-lint": "^1.0.0-beta.3",
     "ember-cli-uglify": "^3.0.0",
     "ember-component-css": "^0.6.9",
+    "ember-copy": "^1.0.0",
     "ember-d3": "^0.5.1",
-    "ember-data": "~3.9.3",
-    "ember-export-application-global": "^2.0.0",
+    "ember-data": "~3.16.0",
+    "ember-export-application-global": "^2.0.1",
+    "ember-fetch": "^7.0.0",
```
adong, line 103: Please remove `ember-fetch`, since we already have `ember-ajax`

adong:
@wahapo 
Another question, I found `eslint . --fix` is not fixing anything for me. It's very frustrating, and I'm curious how are you approaching the linter errors?

wahapo:
@adong 

1. I just run `npm run "lint:fix"` command.
2. I don't know the details. It may have something to do with the fact that the version of eslint is v5.16.0.
```
$ ./node_modules/eslint/bin/eslint.js -v
v5.16.0
```

adong:
Very strange. If I run the following command to run test locally.

```
CI=true npm test
```

Output:
```
1..1064
# tests 1064
# pass  1064
# skip  0
# fail  0

# ok
```

All tests passed. Not sure whats going on.

wahapo:
I'd like you to actually try running the UI and check if there are any problems. I'm going to check it now too.

app/components/workflow-tooltip/component.js
```diff
@@ -12,17 +12,16 @@ export default Component.extend({
...
-      let left = this.left ? event.layerX - 20 : event.layerX - el.outerWidth() / 2;
 
-      el.css({
-        top,
-        left
-      });
+      let left = this.left ? event.layerX - 20 : event.layerX - el.offsetWidth / 2;
+
+      el.style.top = top;
+      el.style.left = left;
```
adong, line 24: Please keep the old logics. 

```
  didUpdateAttrs() {
    this._super(...arguments);

    const event = get(this, 'tooltipData.mouseevent');
    const el = this.$();

    // setting tooltip position
    if (el && event) {
      let top = event.layerY + get(this, 'tooltipData.sizes.ICON_SIZE');
      let left = this.left ? event.layerX - 20 : event.layerX - el.outerWidth() / 2;

      el.css({
        top,
        left
      });
    }
  }
```


This resulted tooltip not position correctly, after changed back to old code, it renders correctly.
image

adong: I looked again, I think its this issue. `el.css()` add `px` for us, while handle css our self, we have to add `px` ourselves. o

In that case, we can either use the old `el.css` behavior, or update to

```javascript
      el.style.top = `${top}px`;
      el.style.left = `${left}px`;
```

Ref: https://stackoverflow.com/questions/10351291/unable-to-set-div-style-left-and-div-style-top-css-properties-via-javascript

app/components/build-log/component.js
```diff
@@ -369,18 +370,19 @@ export default Component.extend({
         set(this, 'isDownloading', true);
 
         this.getLogs(true).then(() => {
-          this.$('#downloadLink')
-            .attr({
-              download: `${buildId}-${stepName}.log`,
-              href: this.logService.buildLogBlobUrl(buildId, stepName)
-            })[0]
-            .click();
+          const el = this.element.querySelectorAll('#downloadLink');
```
adong, line 373: Minor optimization: 

Because of the `#`, there ought to be only one element with that id, so I think this is the same as

```javascript
const el = this.element.querySelector('#downloadLink');
el.setAttribute('download', `${buildId}-${stepName}.log`);
el.setAttribute('href', this.logService.buildLogBlobUrl(buildId, stepName));
el.click();
```
While `querySelectorAll('#downloadLink');` always returns an array with single element.

config/optional-features.json
```diff
@@ -1,3 +1,6 @@
 {
-  "jquery-integration": true
+  "application-template-wrapper": false,
+  "default-async-observers": true,
+  "jquery-integration": true,
+  "template-only-glimmer-components": true
```
adong, line 5: `  "template-only-glimmer-components": true`

This feature breaks `404` page,

image

and changes html output to 


image


Solution: 
> It is recommended that you enable this feature. Existing applications adopting this optional feature should add a .js file for any existing template-only components containing a basic Ember component class. This will maintain backwards compatibility for existing templates while new template-only components gain the advantages of this feature. (https://guides.emberjs.com/release/configuring-ember/optional-features/#toc_template-only-glimmer-components)



create a file `app/components/404-display/component.js`

```

export default Component.extend({});
```

wahapo:
I don't remember why I added `'script-src': ["'unsafe-eval'"]`, but now that I've tried it, it doesn't seem to be a problem, so I'll remove it. Thanks.

config/optional-features.json
adong: One contributor `ember-component-css` from said this is an issue from "ember-component-css": "^0.6.9"`, there's an alpha version to resolve it, but we are not going to incorporate now. If you are interested in, please follow-up here: https://github.com/ebryn/ember-component-css/issues/342#issuecomment-669707620

wahapo: Ok, I think that's fine for now, too.

app/components/validator-input/component.js
```diff
@@ -5,7 +5,9 @@ export default Component.extend({
   actions: {
     nameClick() {
       this.toggleProperty('isOpen');
-      this.$('#validator-ace-editor').toggle('hidden');
+      this.element
```
adong, line 8: Similar here, `this.querySelector` is preferred when querying for `#id`

app/styles/app.scss
```diff
@@ -178,3 +178,5 @@ body {
 }
 
 @import 'ember-power-select';
+
+@import 'ember-bootstrap/bootstrap';
```
adong, line 182: `@import 'ember-bootstrap/bootstrap'` is duplicated in L67 `app/styles/app.scss`

wahapo:
There may be something wrong with job-list-view. First, select list-view and refresh it to make it a graph view. After doing it a few times, the list-view remained visible.

adong:
> There may be something wrong with job-list-view. Select job-list-view and refresh it, and select job-graph-view. After doing it a few times, the list-view remained visible.
> 


1) I don't see this PR has logics changes to the list-view & graph-view 
2) I'm not able to reproduce it locally 😅

I was thinking maybe we merge this PR, and see. We can always do a bug-fix, if it's reproducible :)

wahapo:
Maybe my locally is just weird. I agree with that policy.

adong:
Merged as `feat(1684)`

@wahapo Thank you for doing the ember migration, great job! Congrats. Now we don't have to worry about ember migration for this year. Truly amazing. 
Next, we are going to pull `ember-denali` 

[`ember install @denali-design/ember@1.0.0-alpha.8`](https://www.npmjs.com/package/@denali-design/ember/v/1.0.0-alpha.8)

wahapo:
@adong I appreciate the many comments and advice. Thank you so much! I hope this commit works well🙏

app/components/validator-pipeline/styles.scss
```diff
@@ -45,4 +45,8 @@
       color: $sd-failure;
     }
   }
+
+  div .hidden {
+    display: node !important;
```
DekusDenial, line 50: none?

app/components/validator-job/styles.scss
```diff
@@ -54,4 +54,8 @@
       color: $sd-failure;
     }
   }
+
+  div .hidden {
+    display: node !important;
```
DekusDenial, line 59: none?

adong: PR https://github.com/screwdriver-cd/ui/pull/578

app/components/validator-pipeline/styles.scss
adong: PR https://github.com/screwdriver-cd/ui/pull/578

PR #579, opened 2020-08-07
fix(2152): Support passing projectKey to getCoverageInfo

tkyi:
## Context

Would be nice if we could use existing projectKey data to get coverage info.

## Objective

This PR passes projectKey to get coverage endpoint.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2152

## License

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -109,7 +109,7 @@ export default Component.extend({
   coverageInfoCompute() {
     // Set coverage query startTime to build start time since user can do coverage during user step
     const buildStartTime = this.buildSteps[0].startTime;
-    const { coverageStepEndTime } = this;
+    const { coverageStepEndTime, buildMeta } = this;
 
     if (!coverageStepEndTime) {
       this.set('coverageInfo', {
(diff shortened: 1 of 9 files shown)


app/components/build-banner/component.js
```diff
@@ -130,7 +130,8 @@ export default Component.extend({
       pipelineId: this.pipelineId,
       prNum: this.prNumber,
       jobName: this.jobName,
-      pipelineName: this.pipelineName
+      pipelineName: this.pipelineName,
+      projectKey: buildMeta.build?.coverageKey || null
```
adong, line 134: If backend takes `projectKey` as `undefined`, then you dont need to ` || null`;

app/pipeline/build/template.hbs
```diff
@@ -19,6 +19,7 @@
   isAuthenticated=session.isAuthenticated
   event=event
   prEvents=prEvents
+  prNumber=job.group
```
adong, line 22: https://github.com/screwdriver-cd/ui/blob/829ce1bad91ce38059cab9d9638348901d3cc64a/app/job/model.js#L24-L28

Took me a while to see `job.group` => `prNumber`

app/shuttle/service.js
```diff
@@ -92,24 +92,9 @@ export default Service.extend({
...
-    jobId,
-    startTime,
-    endTime,
-    pipelineId,
-    prNum,
-    jobName,
-    pipelineName,
-    scope
-  ) {
+  async fetchCoverage(data) {
```
adong, line 95: Could you add some info about what `data` parameter looks like? 

Since it has been refactored in the `app/components/pipeline-list-coverage-cell/component.js`, it's not easy to see what's expected value anymore.

```
        buildId,	
        jobId,	
        startTime,	
        endTime,	
        pipelineId,	
        prNum,	
        jobName,	
        pipelineName,	
        scope
```

PR #581, opened 2020-08-12
fix(2175): fix parameters always use first input values

adong:
## Context

Issue: https://github.com/screwdriver-cd/screwdriver/issues/2175

Current implementation keeps using the old data rather than the newer user input.

## Objective

This PR fixes the issue, and will use new user input instead.

## License

### DIFF
--- a/app/components/pipeline-start/component.js
+++ b/app/components/pipeline-start/component.js
@@ -47,14 +47,10 @@ export default Component.extend({
 
   actions: {
     startBuild(parameters) {
-      let args = this.startArgs;
-
-      if (parameters) {
-        args = parameters;
-      }
+      const args = this.startArgs;
       const startFunc = this.startBuild;
 
-      startFunc.apply(null, args);
+      startFunc.apply(null, [...args, parameters]);
     },
 
     toggleDropdown(toggleAction) {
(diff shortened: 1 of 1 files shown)


app/components/pipeline-start/component.js
```diff
@@ -50,7 +50,7 @@ export default Component.extend({
       let args = this.startArgs;
 
       if (parameters) {
-        args.push(parameters);
+        args = parameters;
```
adong, line 53: Since this code keeps pushing the new params to the end of the array, once L57 `startFunc.apply(null, args);` executes, it will always use the first element in the array, even though `parameters` get updated and saved, just never used the new data.

adong: This is kind destructive, found a better way to write it.
```
  startFunc.apply(null, [...args, parameters]);
```

app/components/pipeline-start/component.js
```diff
@@ -47,14 +47,10 @@ export default Component.extend({
...
-      let args = this.startArgs;
-
-      if (parameters) {
-        args.push(parameters);
-      }
+      const args = this.startArgs;
       const startFunc = this.startBuild;
 
-      startFunc.apply(null, args);
+      startFunc.apply(null, [...args, parameters]);
```
adong, line 53: This way, it handles both

https://github.com/screwdriver-cd/ui/blob/915fb211a69d2c72b5081258a0c9e052897615ba/app/pipeline/events/controller.js#L381

and

https://github.com/screwdriver-cd/ui/blob/915fb211a69d2c72b5081258a0c9e052897615ba/app/pipeline/events/controller.js#L546

PR #583, opened 2020-08-13
fix: add loading message when loading template in create pipeline form

adong:
## Context

Add message Loading... while loading templates during create pipeline page

## Objective

I think there are 2 ways of doing it:

1) Using [`noMatchesMessage`](https://2-x.ember-power-select.com/docs/api-reference)

```javascript
          {{#power-select
            options=templates
            selected=selectedTemplate
            onchange=(action "selectTemplate")
            searchField="name"
            searchEnabled=true
            noMatchesMessage="Loading..."
            as |template| }}
            {{template.name}}
          {{/power-select}}
```

2) using `{{else}}` block 
```
{{#power-select
  options=templates
  selected=selectedTemplate
  onchange=(action "selectTemplate")
  searchField="name"
  searchEnabled=true
  as |template| }}
  {{template.name}}
{{else}}
  <span>Loading...</span>
{{/power-select}}
````

Unfortunately, 2nd way is being deprecated. Ref: https://github.com/cibernox/ember-power-select/issues/1370

Screenshot:

## License

### DIFF
--- a/app/components/pipeline-create-form/template.hbs
+++ b/app/components/pipeline-create-form/template.hbs
@@ -51,6 +51,7 @@
             onchange=(action "selectTemplate")
             searchField="name"
             searchEnabled=true
+            noMatchesMessage="Loading..."
             as |template| }}
             {{template.name}}
           {{/power-select}}
--- a/app/components/pipeline-create-form/template.hbs
+++ b/app/components/pipeline-create-form/template.hbs
@@ -51,7 +51,7 @@
             onchange=(action "selectTemplate")
             searchField="name"
             searchEnabled=true
-            noMatchesMessage="Loading..."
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/components/pipeline-create-form/template.hbs [hunk 1]
```diff
@@ -51,6 +51,7 @@
             onchange=(action "selectTemplate")
             searchField="name"
             searchEnabled=true
+            noMatchesMessage="Loading..."
```
jithine, line 54: does this mean, if I search for something and it's not there in the list, it still sows `Loading` ?

adong: You are correct. It will show `Loading...` for non-matched items. Let me update to handle that case.

PR #588, opened 2020-09-01
fix: add DISPLAY_NAME_LENGTH to be configurable

adong:
## Context

People are frustrated not able to customize job name's length to fit their own needs, and we cannot predict the proper length for each use case, so we are opening up the configuration to users to decide. 

This is just a first step toward building customizing personal preference, which yet to come

## Objective
To enable users to have a customized length of job name in the workflow graph page, by using `localStorage`. 

## License


tkyi:
I still think focusing on displaying job description would let users have shorter job names and resolve this display issue: https://github.com/screwdriver-cd/screwdriver/issues/1493

adong:
> I still think focusing on displaying job description would let users have shorter job names and resolve this display issue: [screwdriver-cd/screwdriver#1493](https://github.com/screwdriver-cd/screwdriver/issues/1493)

Yep, cannot agree with you enough.

app/components/workflow-graph-d3/component.js
```diff
@@ -136,7 +137,14 @@ export default Component.extend({
     });
   },
   draw(data) {
-    const MAX_DISPLAY_NAME = 20;
+    let MAX_DISPLAY_NAME = ENV.APP.DISPLAY_NAME_LENGTH;
+
+    const preferredLength = parseInt(localStorage.getItem('DISPLAY_NAME_LENGTH'), 10);
```
jithine, line 142: can you keep default to current 20

adong: Default is 20, `10` means base 10 for `parseInt(numberInStringFormat, 10)`,

jithine: 🤦

PR #589, opened 2020-09-01
fix: implement dropdown to 'pipeline start from here' on workflow graph

adong:
## Context

Currently, the dropdown selection is only shown at the start of the pipeline, this PR will implement and add dropdown selection to `Start pipeline from here`

## Objective

After
image

## License

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -4,6 +4,7 @@ import { get, computed, set, setProperties } from '@ember/object';
 import { reject } from 'rsvp';
 import { isRoot } from 'screwdriver-ui/utils/graph-tools';
 import { isActiveBuild } from 'screwdriver-ui/utils/build';
+import { copy } from 'ember-copy';
 
 export default Component.extend({
   // get all downstream triggers for a pipeline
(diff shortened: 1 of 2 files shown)


app/components/pipeline-workflow/template.hbs
```diff
@@ -78,7 +78,7 @@
         {{/if}}
       {{else if (get-length tooltipData.selectedEvent.meta.parameters)}}
         {{#pipeline-parameterized-build
-          buildParameters=tooltipData.selectedEvent.meta.parameters
+          buildParameters=buildParameters
```
jithine, line 81: does this pull default params, or the event's params from `meta`?

adong: Former, default parameters. 

This PR change will get parameters from `pipelines`, not the parameters which were used.

jithine: when this parameter popup appears on job restart context it should pick values from the event for the build

app/components/pipeline-workflow/component.js
```diff
@@ -47,6 +48,30 @@ export default Component.extend({
     });
   },
 
+  buildParameters: computed('tooltipData', function preselectBuildParameters() {
+    const defaultParameters = this.getWithDefault('pipeline.parameters', {});
+    const buildParameters = copy(defaultParameters, true);
+
+    if (this.tooltipData) {
```
adong, line 55: I just updated that it will show the used value as the selected value. @jithine

PR #594, opened 2020-09-03
fix(715): Catch error message on the secrets page

wahapo:
## Context
I want fixed this issue.
https://github.com/screwdriver-cd/screwdriver/issues/715

## Objective
It displays errors, as shown in the screenshot below.

## References
The cause was that an exception was occurring [here](https://github.com/screwdriver-cd/ui/blob/44691e60419488c24fc17438dd5d987f1ee8b371/app/pipeline/secrets/route.js#L21), but we wasn't catch. The same goes for [secrets](https://github.com/screwdriver-cd/ui/blob/44691e60419488c24fc17438dd5d987f1ee8b371/app/pipeline/secrets/route.js#L16).

I'm not sure how to write a test in this case.

## License

### DIFF
--- a/app/components/pipeline-secret-settings/component.js
+++ b/app/components/pipeline-secret-settings/component.js
@@ -10,12 +10,6 @@ export default Component.extend({
   errorMessage: '',
   secretsSorting: ['name'],
   sortedSecrets: sort('secrets', 'secretsSorting'),
-  didReceiveAttrs() {
-    this._super(...arguments);
-    this.get('secrets').catch(error => {
-      this.set('errorMessage', error.errors[0].detail);
-    });
-  },
   isButtonDisabled: computed('newName', 'newValue', {
     get() {
       return !this.newName || !this.newValue;
--- a/app/components/pipeline-secret-settings/component.js
+++ b/app/components/pipeline-secret-settings/component.js
@@ -10,6 +10,12 @@ export default Component.extend({
   errorMessage: '',
   secretsSorting: ['name'],
   sortedSecrets: sort('secrets', 'secretsSorting'),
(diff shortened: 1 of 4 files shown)


adong:
> I'm not sure how to write a test in this case.

Just write a unit test let `this.set('secrets') = RSVP.reject()` and verify the `this.errorMessage` is the `errorMessage` that supposed to be.

adong:
After I take a look. Could you try add `error` action in `app/pipeline/secrets/route.js` instead? See, 
```
  actions: {
    error(error) {
      // if error.errors[0].detail is the error messages
	  this.controllerFor('pipeline.secrets').set('errorMessage', error.errors[0].detail);

      return true;
    }
 }
```

https://github.com/screwdriver-cd/ui/blob/44691e60419488c24fc17438dd5d987f1ee8b371/app/pipeline/events/show/route.js#L73-L79 as example 

Ref: https://guides.emberjs.com/v3.16.0/routing/loading-and-error-substates/#toc_the-error-event

adong:
Actually, maybe just this to `app/pipeline/secrets/route.js`?

```javascript
    return this.store
      .findAll('token', { adapterOptions: { pipelineId: pipeline.get('id') } })
      .then(tokens => ({
        tokens,
        secrets,
        pipeline
      })).catch((error) => {
        this.controllerFor('pipeline.secrets').set('errorMessage', error.errors[0].detail); // <-
      });
```

image

wahapo:
Oh, it's great, it's going to work.

adong (approved):

jithine:
👏 Thank you for fixing this.

PR #598, opened 2020-09-17
fix: prevent clicking on pr event will trigger redirect to event pages

adong:
## Context

Some user reports that click on PR-build, will trigger redirect to event page.
image

## Objective

This is a regression issue, and this PR will fix that.

## License


jithine (approved):
We have to figure out a way to share PR events as well.

PR #616, opened 2020-10-22
feat: [UX] Permalink to grid/list view

OronNadiv:
## Context
Users want to have the ability to go directory to grid/list view.

## Objective
Create a permalink to `pipeline.events.show` in "list" or "grid" view.

This permlink (`/pipelines/123/events/456?view=list`) will directly show the list view:
image

This permlink (`/pipelines/123/events/456?view=grid`) will directly show the grid view:
image

Clicking the toggle will change the URL (`view=grid` vs `view=list`) 


OronNadiv:
@adong I am aware that the tests are failing but wanted to have your input about my change to see that I'm addressing it correctly.  If my solution is right, then I'll go ahead and fix the tests.  Thanks.

jithine (commented):
grid view shouldn't have the eventId in URl.  Should be like `/pipelines/123/events?view=grid`

OronNadiv:
> grid view shouldn't have the eventId in URl. Should be like `/pipelines/123/events?view=grid`

I don't think that's the case.  `/pipelines/123/events` will be redirected to `/pipelines/123/events/456`.

[Here](https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/events/index/route.js#L10) is the code.

app/pipeline/events/controller.js
```diff
@@ -374,7 +380,11 @@ export default Controller.extend(ModelReloaderMixin, {
         });
       }
 
-      this.set('showListView', showListView);
+      const eventId = get(this.selectedEventObj, 'id');
+
+      this.transitionToRoute('pipeline.build', this.get('pipeline.id'), eventId, {
+        queryParams: { view: showListView ? 'list' : 'grid' }
```
adong, line 386: Is routing to another route with query params meant to keep in the browser history? So, people can go back and forward?

OronNadiv: I don't think it's necessary.  What is the best way to change the `view` query param without calling `transitionToRoute`?

adong: The old way, `this.set('view', toBe);` 

Now, you can just write, `this.view = toBe`.

OronNadiv: Done.  Thanks for your feedback @adong .

adong (approved):

PR #620, opened 2020-10-27
feat(2199): add pipeline settings

adong:
## Context

Allow users to define their own downtime jobs
## Objective

image

## License

### DIFF
--- a/app/components/pipeline-options/component.js
+++ b/app/components/pipeline-options/component.js
@@ -4,9 +4,11 @@ import { not, or, sort } from '@ember/object/computed';
 import { computed } from '@ember/object';
 import Component from '@ember/component';
 import ENV from 'screwdriver-ui/config/environment';
-import { parse, getCheckoutUrl } from '../../utils/git';
+import { parse, getCheckoutUrl } from 'screwdriver-ui/utils/git';
+import { statuses } from 'screwdriver-ui/utils/build';
 
 const { MINIMUM_JOBNAME_LENGTH } = ENV.APP;
+// const DEFAULT_DOWNTIME_STATUSES = ['FAILURE'];
 
 export default Component.extend({
   store: service(),
@@ -31,6 +33,9 @@ export default Component.extend({
   user: null,
   jobId: null,
   jobSorting: ['name'],
+  statuses,
+  metricsDowntimeStatuses: [],
+  metricsDowntimeJobs: [],
   minDisplayLength: MINIMUM_JOBNAME_LENGTH,
   sortedJobs: sort('jobs', 'jobSorting'),
   isInvalid: not('isValid'),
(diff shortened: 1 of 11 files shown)


app/components/pipeline-options/component.js
```diff
@@ -71,6 +74,12 @@ export default Component.extend({
     }
 
     this.set('desiredJobNameLength', desiredJobNameLength);
+
+    const metricsDowntimeJobs = this.get('pipeline.settings.metricsDowntimeJobs').map(jobId => {
```
adong, line 78: Can I have job name “main” from a pipeline, that currently, job id is 1, and that later changed to 11?

In that case, this might break. Bcause we cannot find old jobId in new jobs.

tkyi: it shouldn't be very often that a `jobId` would change for the same job name
jobs will only be added or changed if the user changes it in the yaml

either way, the UI should be able to look up the correct `jobId` in the `workflowGraph.nodes` in the `pipeline` model

jithine:
Does it make sens to show this without the graph being added first ?

adong:
> Does it make sens to show this without the graph being added first ?

Good call, Let me add a feature flag to hide it. Will have another PR to add the graphics and enable such feature.

app/components/pipeline-options/template.hbs
```diff
@@ -37,6 +40,34 @@
               </div>
             </div>
           </li>
+
+          {{#if displayDowntimeJobs}}
+            <li>
+              <div class="row">
+                <div class="col-xs-10">
+                  <h4>Downtime Jobs</h4>
+                  <p>Pick your own preferred jobs to be counted in metrics graph</p>
```
tkyi, line 49:
```suggestion
                  <p>Pick your own preferred jobs to be counted in metrics graph (default all jobs)</p>
```

config/environment.js
```diff
@@ -74,8 +75,11 @@ module.exports = environment => {
   };
 
   if (environment === 'development') {
+    delete require.cache[require.resolve('./local.js')];
     const localAppConfig = require('./local.js'); // eslint-disable-line global-require
 
+    console.log('localAppConfig', localAppConfig); // eslint-disable-line no-console
```
tkyi, line 81: console log

app/shuttle/service.js
```diff
@@ -130,5 +130,26 @@ export default Service.extend({
     const raw = true;
 
     return this.fetchFromApi(method, url, data, raw);
+  },
+
+  /**
+   * updatePipelineSettings
+   * @param  {Number} pipelineId [description]
```
tkyi, line 137: missing param descriptions

app/shuttle/service.js
```diff
@@ -130,5 +130,26 @@ export default Service.extend({
...
+   * @param  {Object} settings
+   * @param  {Array}  settings.metricsDowntimeJobs
+   * @return {Promise}
+   */
+  async updatePipelineSettings(pipelineId = 1, settings) {
```
tkyi, line 142: should pipelineId really be default `1`?

config/environment.js
adong: This is for `development` environment, and for print out the local endpoints config only.

app/shuttle/service.js
adong: Not sure, we can remove the default `pipelineId`, or set to some other number?

app/components/pipeline-options/template.hbs
```diff
@@ -168,7 +199,7 @@
           <div class="row">
             <div class="col-xs-10">
               <h4>Display Name Length</h4>
-              <p>Setup your own preferred job name length to display on the workflow graph</p>
+              <p>Setup your own preferred job name length to display on the workflow graph (range: 20 - 99)</p>
```
adong, line 202: This is for https://github.com/screwdriver-cd/screwdriver/issues/2234

app/shuttle/service.js
tkyi: yeah just remove default I guess

tkyi (approved):

app/shuttle/service.js
adong: Will update that next PR.

PR #621, opened 2020-10-30
feat: jobs endpoint

OronNadiv:
## Context

The way we presented the "jobs" could be confusing.  We want to have a separate dialog and URL for the jobs view.

## Objective

We moved the jobs view to a separate controller.  The jobs view 

image


jithine (approved):
few minor cleanups, LGTM

app/components/pipeline-graph-nav/styles.scss
```diff
@@ -30,7 +30,6 @@
   }
 
   .view-toggle {
```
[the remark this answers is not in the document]
OronNadiv: Yes, just in case we want to use it in the future so it's good to keep it there.  This way it will be more visible to other devs.

app/pipeline/events/controller.js
```diff
@@ -8,10 +8,156 @@ import ENV from 'screwdriver-ui/config/environment';
 import ModelReloaderMixin from 'screwdriver-ui/mixins/model-reloader';
 import { isPRJob } from 'screwdriver-ui/utils/build';
 
+/** */
```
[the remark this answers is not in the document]
OronNadiv: It's part of the lint validation.  I can either remove the commend and add `eslint-ignore` or have an empty comment.  I'll go with the `eslint-ignore`

jithine (approved):

app/pipeline/jobs/index/route.js
```diff
@@ -0,0 +1,36 @@
+import { inject as service } from '@ember/service';
+import Route from '@ember/routing/route';
+import ENV from 'screwdriver-ui/config/environment';
+import RSVP from 'rsvp';
+
+export default Route.extend({
```
adong, line 6: To use `routeAfterAuthentication`, this route needs to be extended from the following:

```
import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';

export default Route.extend(AuthenticatedRouteMixin, {
```

adong:
Builds button is not working any more.
image

app/components/pipeline-nav/template.hbs
```diff
@@ -5,7 +5,7 @@
   {{#if pipeline.childPipelines}}
     {{#nav.item}}{{#nav.link-to "pipeline.child-pipelines"}}Child Pipelines{{/nav.link-to}}{{/nav.item}}
   {{/if}}
-  {{#nav.item}}{{#nav.link-to "pipeline.events"}}Events{{/nav.link-to}}{{/nav.item}}
+  {{#nav.item}}{{#nav.link-to buildsLink}}Builds{{/nav.link-to}}{{/nav.item}}
```
adong, line 8: `buildsLink` I think this is the issue. 

Options
1) Revert back to old
2) If you want to keep the flexibility to pass link, it also needs to see where 
```
{{pipeline-nav
```
being used and adds rest like `pipeline/secrets/template.hbs`, `pipeline/options/template.hbs`  `pipeline/options/metrics.hbs`  

3) Or, use this `{{or buildsLink "default"}}`

```hbs
  {{#nav.item}}{{#nav.link-to (or buildsLink "pipeline.events")}}Builds{{/nav.link-to}}{{/nav.item}}
```

I think 3rd option makes sense here

adong (changes requested):

app/pipeline/events/controller.js
```diff
@@ -8,10 +8,156 @@ import ENV from 'screwdriver-ui/config/environment';
...
+        event.hasMany('builds').reload();
+      }
+    } catch (e) {
+      this.set('errorMessage', Array.isArray(e.errors) ? e.errors[0].detail : '');
+    }
+  }
+}
+
+// eslint-disable-next-line require-jsdoc
+export async function startDetachedBuild(job, options = {}) {
```
adong, line 43: I felt like this is [anti-pattern](https://deprecations.emberjs.com/v3.x/#toc_object-alias-method), refactoring so can be shared in two multiple controller is a good idea. 

Because `this` scope is tightly coupled, it
s kinda defeat the purpose... 

As far as suggestions, I dont have one, just wanted to point out here.

app/pipeline/jobs/index/route.js
OronNadiv: 👍

app/pipeline/events/controller.js
OronNadiv: I totally agree.  Thanks for pointing it out.

app/components/pipeline-nav/template.hbs
OronNadiv: Good catch.  I thought I tested it but apparently, I did not.
I fixed it using a new service.  Check out [this](https://github.com/screwdriver-cd/ui/pull/621/commits/e73baeaccf09e5cf78826a33b875c247da1d55bc) commit.

app/pipeline/jobs/index/controller.js
```diff
@@ -0,0 +1,204 @@
...
+        };
+      } else {
+        eventPayload = {
+          pipelineId,
+          startFrom,
+          causeMessage
+        };
+      }
+
+      // await this.createEvent(eventPayload, false);
```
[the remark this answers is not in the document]
OronNadiv: done.

app/pipeline/jobs/index/route.js
OronNadiv: @adong are you sure it's required?  It already exists in the 'pipeline' route.  See [here](https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/route.js#L6).

app/components/pipeline-nav/component.js
```diff
@@ -1,8 +1,13 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { htmlSafe } from '@ember/string';
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
+  pipelineService: service('pipeline'),
+  buildsLink: computed('pipelineService.buildsLink', function() {
```
ldong, line 8: Though I think using service is overkill, and it's okay.


Since dependent keys and return values are the same, you can use `alias` to shorten it.

```
import { alias } from '@ember/object/computed';

  pipelineService: service('pipeline'),
  buildsLink: alias('pipelineService.buildsLink');
```

Nevertheless, just a minor comment, I'm approving the PR.

app/components/pipeline-nav/component.js
```diff
@@ -1,8 +1,13 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { htmlSafe } from '@ember/string';
+import { inject as service } from '@ember/service';
 
 export default Component.extend({
+  pipelineService: service('pipeline'),
```
adong, line 7: Though I think using service is overkill, and it's okay.


Since dependent keys and return values are the same, you can use `alias` to shorten it.

```
import { alias } from '@ember/object/computed';

  pipelineService: service('pipeline'),
  buildsLink: alias('pipelineService.buildsLink');
```

Nevertheless, just a minor comment, I'm approving the PR.

app/pipeline/jobs/index/route.js
adong: Yes. I believe this causes the error if not extended from `AuthenticatedRouteMixin` : after session timeout, and log back in, not being redirected.

adong (approved):

OronNadiv:
@adong thanks for your diligent review, and your help with this feature.

PR #623, opened 2020-11-01
fix: workflow graphs "flicker" every few seconds.

OronNadiv:
## Context
The workflow graph flickers every few seconds.  This is due to different sorting of the nodes/edges every time we fetch it from the server.

### DIFF
--- a/app/event/serializer.js
+++ b/app/event/serializer.js
@@ -1,24 +1,30 @@
 import { assign } from '@ember/polyfills';
 import DS from 'ember-data';
+import _ from 'underscore';
+
+export const sortWorkflowGraph = workflowGraph => {
+  if (workflowGraph) {
+    if (workflowGraph.nodes) {
+      workflowGraph.nodes = _.sortBy(workflowGraph.nodes, node => {
+        return `${node.name}_____${node.id}`;
+      });
+    }
+    if (workflowGraph.edges) {
+      workflowGraph.edges = _.sortBy(workflowGraph.edges, edge => {
+        return `${edge.src}_____${edge.dest}`;
+      });
+    }
+  }
+};
 
 export default DS.RESTSerializer.extend({
   normalizeResponse(store, typeClass, payload, id, requestType) {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -138,6 +138,7 @@
     "memoizerific": "^1.11.3",
     "prettier": "^1.18.2",
     "qunit-dom": "^1.0.0",
-    "sass": "^1.23.0"
+    "sass": "^1.23.0",
+    "underscore.string": "^3.3.5"
```
jithine, line 142: this is a project which is not being updated over 5years. Not sure if it's still relevant or supported.

app/event/serializer.js
```diff
@@ -1,24 +1,33 @@
 import { assign } from '@ember/polyfills';
 import DS from 'ember-data';
+import _s from 'underscore.string';
+
+export const sortWorkflowGraph = workflowGraph => {
+  if (workflowGraph) {
+    if (workflowGraph.nodes) {
+      workflowGraph.nodes.sort((node1, node2) => {
+        return (
+          _s.naturalCmp(node1.name, node2.name) ||
```
jithine, line 10: shouldn't https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare do this ?

OronNadiv: No.
```
console.log('1'.localeCompare('10'));
-1

console.log('2'.localeCompare('10'));
1
```

package.json
OronNadiv: It's a very popular package.  It hasn't been updated for two years because it almost has no dependencies (rear security vulnerabilities).
It is a simple implementation of string manipulations.
image

adong (approved):

app/event/serializer.js
DekusDenial: fyi, enabling `numeric` option for `localeCompare` will be enough for alphanumerical sorting

PR #628, opened 2020-11-10
feat: Auto refresh jobs and events.

OronNadiv:
## Context

In both workflow and list views, the user should be able to:
1. See new events/jobs as they triggered on other tabs/computers.
1. See a job starts/stops without clicking the "refresh" button.


jithine (approved):

app/components/pipeline-list-view/component.js
```diff
@@ -208,6 +209,16 @@ export default Component.extend({
 
     return rows;
   },
+  jobsObserver: observer('jobsDetails.[]', function jobsObserverFunc({ jobsDetails }) {
+    const rows = this.getRows(jobsDetails);
+    const lastRows = get(this, 'lastRows') || [];
+    const isEqualRes = isEqual(rows.map(r => r.job), lastRows.map(r => r.job));
```
adong, line 215: I'm not sure if the response returns jobs in a consistent order, maybe we need to `sort` before comparing.

adong: I could be wrong, but I don't think the observer's function has parameters.

```
jobsObserverFunc({ jobsDetails }) {
    const rows = this.getRows(jobsDetails);
```

So this function always sets the table with new rows.

app/pipeline/jobs/index/controller.js
```diff
@@ -1,14 +1,39 @@
...
+import { get, set, computed } from '@ember/object';
 import { jwt_decode as decoder } from 'ember-cli-jwt-decode';
 
 import ENV from 'screwdriver-ui/config/environment';
 import ModelReloaderMixin from 'screwdriver-ui/mixins/model-reloader';
-import { isPRJob } from 'screwdriver-ui/utils/build';
+import { isPRJob, isActiveBuild } from 'screwdriver-ui/utils/build';
+import moment from 'moment';
 import { createEvent, startDetachedBuild, stopBuild, updateEvents } from '../../events/controller';
+import { SHOULD_RELOAD_SKIP, SHOULD_RELOAD_YES } from '../../../mixins/model-reloader';
```
adong, line 11: Ember know the relative path pain, and you could change the syntax from 

`'../../../mixins/model-reloader';`

to 

`'screwdriver-ui/mixins/model-reloader';`

app/pipeline/jobs/index/controller.js
```diff
@@ -22,11 +47,14 @@ export default Controller.extend(ModelReloaderMixin, {
     });
   },
 
-  reload() {
+  async reload() {
     try {
       this.send('refreshModel');
+      await this.refreshListViewJobs();
```
adong, line 53: I dont think this guarantees `model()` is fetched and loaded when line `this.refreshListViewJobs()` runs.

app/pipeline/jobs/index/controller.js
```diff
@@ -192,8 +220,14 @@ export default Controller.extend(ModelReloaderMixin, {
       }
 
       await this.createEvent(eventPayload);
+      await this.reload();
+      set(this, 'lastRefreshed', moment().subtract(1, 'day'));
```
adong, line 224: `moment().subtract(1, 'day')`, I don't get the logic here, why the date is set to yesterday?

adong (changes requested):
👀

app/pipeline/jobs/index/controller.js
OronNadiv: 👍

app/pipeline/jobs/index/controller.js
OronNadiv: I wanted to reset the "lastRefreshed" value to trigger another refresh.

app/components/pipeline-list-view/component.js
OronNadiv: Done. 👍

app/pipeline/jobs/index/controller.js
OronNadiv: Good point but I can't find a use-case where refreshListViewJobs does not refresh the list.

app/components/pipeline-list-view/component.js
```diff
@@ -208,6 +209,19 @@ export default Component.extend({
 
     return rows;
   },
+  jobsObserver: observer('jobsDetails.[]', function jobsObserverFunc({ jobsDetails }) {
+    const rows = this.getRows(jobsDetails);
```
adong, line 213: I have to say that I didn't think there would be parameters for observer function,  and I was wrong.

It indeed has
```
observer('value', function(sender, key, value, rev) {
```

see Ref: https://api.emberjs.com/ember/3.20/classes/Observable/methods

app/components/pipeline-list-view/component.js
```diff
@@ -208,6 +209,19 @@ export default Component.extend({
...
+    const lastRows = get(this, 'lastRows') || [];
+    const isEqualRes = isEqual(
+      rows.map(r => r.job).sort((a, b) => (a.jobName || '').localeCompare(b.jobName)),
+      lastRows.map(r => r.job).sort((a, b) => (a.jobName || '').localeCompare(b.jobName))
+    );
```
adong, line 218: 👍

app/pipeline/jobs/index/controller.js
adong: It's okay.

adong (approved):
Looks good to me 👍

jithine (approved):

PR #629, opened 2020-11-12
fix: fix workflow nodes sorting

OronNadiv:
## Context
Workflow nodes order on the first column should be the same order they appear in the `screwdriver.yaml` file.


app/event/serializer.js
```diff
@@ -5,17 +5,10 @@ const compare = (s1, s2) =>
   (s1 || '').localeCompare(s2 || '', undefined, { numeric: true, sensitivity: 'base' });
 
 export const sortWorkflowGraph = workflowGraph => {
-  if (workflowGraph) {
-    if (workflowGraph.nodes) {
```
jithine, line 9: does removing node sorting still address the previous flickering issue ?

jithine (approved):

PR #633, opened 2020-11-17
feat(2267): refactor collection page and apply lazy loading

klu909:
## Context

Collection page does not render until metrics for all pipeline are loaded.
## Objective

Move metric query to each individual pipeline card/row component so collection page render without waiting for metrics 
## References

https://github.com/screwdriver-cd/screwdriver/issues/2267

## License

### DIFF
--- a/app/components/collection-table-row/template.hbs
+++ b/app/components/collection-table-row/template.hbs
@@ -8,15 +8,21 @@
 </td>
 <td class="branch">
   {{fa-icon "code-fork"}}
-  <span title={{pipeline.branch}}>{{pipeline.branch}}</span>
+  <span title={{branch}}>{{branch}}</span>
 </td>
 <td class="status">
   {{#link-to "pipeline" pipeline.id title=pipeline.scmRepo.name}}
     {{fa-icon lastEventInfo.icon class=lastEventInfo.statusColor}}
   {{/link-to}}
-  <a href={{lastEventInfo.commitUrl}}>
-    {{lastEventInfo.sha}}
-  </a>
+  {{#if storeQueryError}}
+    An error occur while loading events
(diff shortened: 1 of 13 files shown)


app/components/collection-table-row/template.hbs
```diff
@@ -8,15 +8,21 @@
...
+  <span title={{branch}}>{{branch}}</span>
 </td>
 <td class="status">
   {{#link-to "pipeline" pipeline.id title=pipeline.scmRepo.name}}
     {{fa-icon lastEventInfo.icon class=lastEventInfo.statusColor}}
   {{/link-to}}
-  <a href={{lastEventInfo.commitUrl}}>
-    {{lastEventInfo.sha}}
-  </a>
+  {{#if queryError}}
```
adong, line 17: What `queryError` could be?

app/components/pipeline-card/component.js
```diff
@@ -11,20 +15,46 @@ export default Component.extend({
   pipelineSelected: false,
   classNames: ['pipeline-card'],
   reset: false,
-
+  queryError: false,
+  updated: and('eventsInfo', 'lastEventInfo'),
```
adong, line 19: Can we rename `updated` to `hasBothEventsAndLatestInfo` ? Latter is more intuitive.

adong (changes requested):
Please add some descriptions to this PR.

app/components/collection-table-row/template.hbs
klu909: if store query fails, for example getting 402 or 404 for whatever reason

klu909: it's just a boolean

adong (approved):

PR #644, opened 2020-12-03
fix: remove scroll

adong:
## License


jithine:
Is this the right place ? Looks like changes are in pipeline events, it should be done for pipeline jobs view ?

PR #648, opened 2020-12-15
fix(2283): enable to download `sd-setup-init` log

MysticDoll:
## Context
API returns 500 when we try to download `sd-setup-init` log.
This PR makes it enable to download `sd-setup-init` log.

## Objective
- create log file for `sd-setup-init` by `URL.createObjectURL`

## References
screwdriver-cd/screwdriver#2283

## License

### DIFF
--- a/app/components/build-log/component.js
+++ b/app/components/build-log/component.js
@@ -368,9 +368,26 @@ export default Component.extend({
     },
     download() {
       const { buildId, stepName } = this;
-      const downloadLink = `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/builds/${buildId}/steps/${stepName}/logs?type=download`;
 
-      window.open(downloadLink, '_blank');
+      if (stepName === 'sd-setup-init') {
+        const fileName = `${stepName}-log.text`;
+        const downloadLink = URL.createObjectURL(
+          new File([this.logs.map(({ m }) => m).join('\n')], fileName, {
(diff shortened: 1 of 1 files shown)


app/components/build-log/component.js
```diff
@@ -368,9 +368,26 @@ export default Component.extend({
...
+            type: 'text/plain'
+          })
+        );
+
+        const e = document.createElement('a');
+
+        e.setAttribute('href', downloadLink);
+        e.setAttribute('download', fileName);
+        e.setAttribute('target', '_blank');
+        e.click();
```
MysticDoll, line 385: We cannot open window for blob URL.
Therefore we should create Element for blob link.

jithine (commented):
Perhaps we should have moved the implementation of `sd-setup-init` logs to API, then it would have behaved like any other steps logs.

MysticDoll:
> Perhaps we should have moved the implementation of `sd-setup-init` logs to API, then it would have behaved like any other steps logs.

I think so too. I will close this PR.

But it is the pseudo log for `sd-setup-init` and there is no actual log for it.
I don't know which is better disable the download link for it or handle the log for it in API. 😢

jithine:
@MysticDoll Doing in API means will have to fetch build model to do what UI is doing here. https://github.com/screwdriver-cd/ui/blob/38e2ce7c306e3b855ba1627d682037a11a06e460/app/components/build-log/component.js#L74-L152

Maybe UI should handle it.

PR #651, opened 2020-12-18
fix: lazy load coverage when row/cell is in view-port

adong:
## Context

Current implementation of getting coverages will trigger a second API call, with this PR, we will load on-demand and will only load coverages when in viewport.

## Objective

Sample coverage pipeline: https://cd.screwdriver.cd/pipelines/67/jobs
## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #655, opened 2021-01-11
fix: increase limit from 10 to 200 to load in job list view

adong:
## Context

Since https://github.com/screwdriver-cd/ui/pull/651, we can increase the number of jobs displayed in the first load of the job-list-view, without triggering coverages that are not in the viewport.

## Objective

It's a trade off between loading/pre-loading a number of jobs in the job-list-view, or let users scroll multiple times to the desired number of jobs. So, this PR increases from 20 to 200.

## License

### DIFF
--- a/app/pipeline/jobs/index/controller.js
+++ b/app/pipeline/jobs/index/controller.js
@@ -18,9 +18,13 @@ export default Controller.extend(ModelReloaderMixin, {
   lastRefreshed: moment(),
   shouldReload(model) {
     const job = model.jobs.find(j => {
-      return j.builds.find(b => {
-        return isActiveBuild(b.get('status'), b.get('endTime'));
-      });
+      if (j.hasMany('builds').value() !== null) {
+        return j.builds.find(b => {
+          return isActiveBuild(b.get('status'), b.get('endTime'));
+        });
+      }
+
+      return null;
     });
 
     let res;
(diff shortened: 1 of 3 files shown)


jithine (approved):

app/pipeline/jobs/index/controller.js
```diff
@@ -18,9 +18,13 @@ export default Controller.extend(ModelReloaderMixin, {
   lastRefreshed: moment(),
   shouldReload(model) {
     const job = model.jobs.find(j => {
-      return j.builds.find(b => {
-        return isActiveBuild(b.get('status'), b.get('endTime'));
-      });
+      if (j.hasMany('builds').value() !== null) {
```
adong, line 21: It turns out many build API calls were invoked here by referencing `j.builds`, so we can add a check to see if we have builds loaded before reloading the nested models. 

Once that is checked, the bottleneck I see from making from `${ENV.APP.SDAPI_HOSTNAME}/${ENV.APP.SDAPI_NAMESPACE}/builds/statuses` were gone, and the loading situation is dramatically reduced. 
https://github.com/screwdriver-cd/ui/blob/395a0479ac47cf8f3d95861fc3a43e9ebf2f57bc/app/build-history/adapter.js#L9-L11


Reference: 
1) how to check if the model is loaded. https://github.com/emberjs/data/issues/5629#issuecomment-420565784

DekusDenial (approved):

jithine (approved):

PR #657, opened 2021-01-13
feat: ignore spaces when searching.

jithine:
## Context

Spaces in search text doesn't return any results.

## License


adong (approved):
LGTM

PR #664, opened 2021-02-04
fix(2327): Restart dialog in workflow graph is confusing

adong:
## Context

Sometimes words can be confusing to our users, we are adding `Restart` if the build had run before, `Start` otherwise, see issue https://github.com/screwdriver-cd/screwdriver/issues/2327 for more context.

## Objective

Before:
image

After:
image

image

If a job's status is undefined, then we will show it as `Start`, otherwise, will show it as `Restart`.

image

## References

Issue: https://github.com/screwdriver-cd/screwdriver/issues/2327
## License


adong:
Need to add a test case

vnugopal (approved):

PR #665, opened 2021-02-06
fix: add hide PR button under pipeline options

adong:
## Context

Some users prefer only to see jobs that are triggered by `~commit`, and not from `~pr`

So, we are implementing a preference and leave the choices to users.
## Objective

See screenshot:

TEST PENDING, DO NOT MERGE
## License

### DIFF
--- a/app/pipeline/events/route.js
+++ b/app/pipeline/events/route.js
@@ -13,10 +13,17 @@ export default Route.extend({
   },
   setupController(controller, model) {
     this._super(controller, model);
-    controller.set('activeTab', 'events');
+    const { pipelinePreference } = model;
+
+    controller.setProperties({
+      activeTab: 'events',
+      showPRJobs: pipelinePreference.showPRJobs
+    });
+
     this.get('pipelineService').setBuildsLink('pipeline.events');
   },
   model() {
+    const pipelineId = this.get('pipeline.id');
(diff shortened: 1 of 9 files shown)


tkyi:
Cool
Wonder if we should move the show/hide triggers toggle to the same section 🤔

adong:
Rebase with latest master

app/pipeline/events/route.js
```diff
@@ -13,22 +13,32 @@ export default Route.extend({
...
       jobs: this.get('pipeline.jobs'),
       events: this.store.query('event', {
-        pipelineId: this.get('pipeline.id'),
+        pipelineId,
         page: 1,
         count: ENV.APP.NUM_EVENTS_LISTED
       }),
-      triggers: this.triggerService.getDownstreamTriggers(this.get('pipeline.id'))
+      triggers: this.triggerService.getDownstreamTriggers(this.get('pipeline.id')),
+      pipelinePreference: this.store.queryRecord('preference/pipeline', {
```
jithine, line 39: is this saved in DB  or is still browser local storage ?

adong: Stored in `localStorage`, and it's configured for each individual user

jithine: didn't we add an API for storing user preferences ? please change this to use that. Will be a good time to change the job width preference as well.

PR #666, opened 2021-02-10
fix: hide options to disable PR jobs

adong:
## Context

It seems we overlooked the use case of disabling PR jobs, which does not make much sense.

## Objective

Before
image

After:
image

## License


adong:
Rebased with latest master

tkyi (approved):

PR #661, opened 2021-01-25
fix: only reload model if has internet access

adong:
## Context

Every once in a while, users are offline from the internet or disconnected from VPN. Our users should be notified if they are offline, rather than being redirected to 404 page.

## Objective

This PR adds the warning only if it was an internet issue, and show screenshot like following:

image

## References

https://github.com/screwdriver-cd/screwdriver/issues/2325

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #668, opened 2021-02-12
fix: runtime error which does not have default values

adong:
## Context

Follow up of https://github.com/screwdriver-cd/ui/pull/661

## Objective

Show empty job list rather than not show jobs.
image

## License


jithine (approved):

jithine (commented):
is it possible to make login an actual link to login?

adong:
> is it possible to make login text in the banner an actual link to login dialogue?

Yes, we can. How about?
image

PR #674, opened 2021-03-03
fix(2350): template/command trusted logic fix

klu909:
## Context

ui should only render template/command as trusted only if latest template/command was trusted
## Objective

ui should only render template/command as trusted only if latest template/command was trusted
## References

https://github.com/screwdriver-cd/screwdriver/issues/2350
## License


jithine (approved):

jithine (commented):
So with this change,  if in command/template detail page, and I am viewing an older version of command/template, it won't show trusted  anymore ?

PR #677, opened 2021-03-12
fix: implement branch removal to recursion

adong:
## Context

It works fine for pipeline like this:

image

image

But, runs into issues if the pipeline is customizied. 

Original: 
image

Before(which has error): 
image

The implementation of this is not recursive, and only removes one level.

## Objective

1) Fix above error to be:

After(fixed):
image

2) This PR also fix redirect to 404 for non-existing pipelines

## License

### DIFF
--- a/app/pipeline/events/route.js
+++ b/app/pipeline/events/route.js
@@ -13,7 +13,7 @@ export default Route.extend({
   beforeModel() {
     this.set('pipeline', this.modelFor('pipeline').pipeline);
   },
-  setupController(controller, model) {
+  setupController(controller, model = {}) {
     this._super(controller, model);
     const pipelinePreference = getWithDefault(model, 'pipelinePreference', {});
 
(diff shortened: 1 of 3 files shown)


tkyi (approved):
Some test failures but looks good to me

app/pipeline/route.js
```diff
@@ -11,7 +11,11 @@ export default Route.extend(AuthenticatedRouteMixin, {
     const collections = this.store.findAll('collection').catch(() => []);
 
     return RSVP.hash({
-      pipeline: this.store.findRecord('pipeline', params.pipeline_id).catch(() => []),
+      pipeline: this.store.findRecord('pipeline', params.pipeline_id).catch(() => {
+        this.transitionTo('/404');
```
jithine, line 15: 404 page should be shown if and only if API response itself is 404, else it could be network error or log off scenario

app/pipeline/events/route.js
```diff
@@ -13,7 +13,7 @@ export default Route.extend({
   beforeModel() {
     this.set('pipeline', this.modelFor('pipeline').pipeline);
   },
-  setupController(controller, model) {
+  setupController(controller, model = {}) {
```
jithine, line 16: 👏 just saw this in local.

PR #685, opened 2021-03-27
fix(2395): Make collection list available in copyTo dropdown on collection view component

kprasadpvv:
## Context

- Available collection list is not displaying under copy to option in default pipeline collection dashboard.

## Objective

- As `model.collections` is not available I'm getting the collection list from store and using it directly in the component instead of passing it from parent component.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2395

##Screenshots:
- Before Fix:

NoCollectionListUnderCopyToOption

- After Fix:

Screenshot 2021-03-27 at 3 27 45 PM

## License


app/components/collection-view/component.js
```diff
@@ -40,6 +40,23 @@ export default Component.extend({
...
   }),
+  collections: computed('store', {
+    get() {
+      if (
+        !get(this, 'session.isAuthenticated') ||
+        get(this, 'session.data.authenticated.isGuest')
+      ) {
+        return [];
+      }
+      const collections = this.store.peekAll('collection');
```
adong, line 51: Though this uses `store.peek`, but it listens to `store` changes. Could be expensive on every store changes, is there a better way 🤔?

adong (changes requested):

app/components/collection-view/component.js
kprasadpvv: Yeah, getting `collections` on component initiation is enough no need to inject `store` as dependency. So, removed it.

Thank you,
kp.

app/components/collection-view/component.js
```diff
@@ -147,6 +164,7 @@ export default Component.extend({
             this.setProperties({
               removePipelineError: null,
               selectedPipelines: [],
+              reset: true,
```
kprasadpvv, line 167: To reset `pipeline selected` checkbox after successfully copy or remove operation along with reseting selectedArray.

adong (approved):

PR #686, opened 2021-03-30
fix(2227): restrict user interaction for builds that in CREATED status

adong:
## Context

See issue https://github.com/screwdriver-cd/screwdriver/issues/2227 on how to reproduce it.

image

## Objective

<details><summary>
<i>Original Thought</i>
</summary>
<p>

I have put in `RUNNING` state, because to me `CREATED` seems to be running, maybe we should change state from _CREATED_ to _QUEUED_ 🤔? 

image

image
</p>
</details>

<details><summary>New Apporach</summary>
<p>

* make users not able to go to build details page in workflow graph
* ~make users not able to go to build details page in job list view~
* do not show builds in CREATED status in job list view

</p>
</details>

## License


jithine (changes requested):
`Created` is not running, builds in `created` status should be hidden in UI in a sens that workflow graph should not allow navigation to that build detail page

tkyi:
Will status still say CREATED on event banner?

jithine:
> Will status still say CREATED on event banner?

maybe we can map `CREATED` event status to Waiting/Running ? @adong @tkyi

adong:
> Will status still say CREATED on event banner?

Yes, that remains. Users just not be able to click to build details page

adong:
> > Will status still say CREATED on event banner?
> 
> maybe we can map `CREATED` event status to Waiting/Running ? @adong @tkyi

I think I'd rather show as-is. So we can see the payload value `CREATED` on the UI, it's consistent, and would be easier to debug.

jithine:
> I think I'd rather show as-is. So we can see the payload value CREATED on the UI, it's consistent, and would be easier to debug.

@adong  event does not have a status. UI is parsing build status to display overall status. I think logic should be tweaked to give prefrence to `RUNNING` status over `CREATED`.
ie if there are builds in both `RUNNING` & `CREATED` status, it should say `RUNNING` as event status. 
now if there is only one build is `CREATED` status, we are talking about the overall status of event, which doesn't have any status of it's own. I think it's better to show event as `RUNNING ` even in this case because technically event is still active.

adong:
Rebased with latest master

jithine (approved):

PR #687, opened 2021-03-30
fix: make sure the height of the steps-list and the console is right.

adong:
## Objective

Commit on behalf of Gilad (gbendor): This will eliminate the annoying "double scrollbars" - one for the main window, and the other for the console.

## License


tkyi:
Is there a better way to handle this in CSS? Seems a little hardcoded?

PR #688, opened 2021-04-05
fix: hide unsuccessful profile image

adong:
## Context

We will hide the Github avatar url if the image loads unsuccessfully. 

~One suspect is that Chrome had removed `Referrers` header and causes GHE to render something like this, and now Firefox follows suit, and is removing `Referrers` header as well.~

Before:
image

## Objective

This `onerror="this.style.visibility='hidden';"` not working, because GHE returns http status with `200` for 0 bytes

image

Now it hides the avatar for [GHE](https://github.com/enterprise).

After:
image

## References

~[Firefox 87 trims HTTP Referrers by default to protect user privacy - Mozilla Security Blog](https://blog.mozilla.org/security/2021/03/22/firefox-87-trims-http-referrers-by-default-to-protect-user-privacy/)~
See details in: 
- [Avatar images are broken on GHE Server running in private mode](https://github.com/Reviewable/Reviewable/issues/770)
- [Cannot load avatars from GitHub Enterprise URLs](https://github.com/microsoft/vscode-pull-request-github/issues/416)
- [Preventing load of avatars on GitHub Enterprise by StanleyGoldman ](https://github.com/microsoft/vscode-pull-request-github/pull/1148) summarized accurately:

> After some research about GitHub Enterprise, I've learned that it will never return an avatar over the avatar url. The authentication needed for this is a cookie created by the GitHub Enterprise instance, we only have an OAuth Token.

## License


jithine (approved):

tkyi (approved):

app/components/user-link/component.js
```diff
@@ -1,5 +1,11 @@
 import Component from '@ember/component';
+import { computed } from '@ember/object';
+import ENV from 'screwdriver-ui/config/environment';
 
 export default Component.extend({
-  classNameBindings: ['large']
+  classNameBindings: ['large'],
+
+  isOpenSourceInstance: computed(function() {
+    return ENV.SDAPI_HOSTNAME === 'https://api.screwdriver.cd';
```
tkyi, line 9: maybe use a constant?

jithine: we should just have a flag, say ENV.SHOW_AVATARS : true/false Let's not hardcode host names

adong: Both GHE and Github versions are `production` environment. Current implementation auto detects.

If we make it like a feature flag, would this flag manually set by Cluster Admins?

jithine: yes, lets cluster admins deal with this. 

https://github.com/screwdriver-cd/ui/blob/57c043d4f3c923d89ec4de0656969563e5066cca/app/instance-initializers/supplementary-config.js#L7
https://github.com/screwdriver-cd/ui/blob/57c043d4f3c923d89ec4de0656969563e5066cca/nginx.conf#L93

adong: OK. In this case, we also need to update the Guide: https://github.com/screwdriver-cd/guide
to reflect on https://docs.screwdriver.cd/

app/instance-initializers/supplementary-config.js
```diff
@@ -29,6 +29,9 @@ export function initialize() {
   if (window.SUPPLEMENTARY_CONFIG && window.SUPPLEMENTARY_CONFIG.RELEASE_VERSION) {
     ENV.APP.RELEASE_VERSION = window.SUPPLEMENTARY_CONFIG.RELEASE_VERSION;
   }
+  if (window.SUPPLEMENTARY_CONFIG && window.SUPPLEMENTARY_CONFIG.SHOW_AVATAR) {
+    ENV.APP.SHOW_AVATAR = window.SUPPLEMENTARY_CONFIG.SHOW_AVATAR;
```
jithine, line 33: need to add to https://github.com/screwdriver-cd/ui/blob/master/nginx.conf#L93 for this to work.

jithine (approved):

PR #690, opened 2021-04-07
fix: handle remote case for undefined value for showPRJobs

adong:
## Context

We treat undefined as false value for `showPRJobs`, thus the default of workflow graph now hide `~pr` jobs

## Objective

This PR fixes just that
## License

### DIFF
--- a/app/pipeline/pulls/route.js
+++ b/app/pipeline/pulls/route.js
@@ -13,7 +13,13 @@ export default EventsRoute.extend({
     this.render('pipeline.events');
   },
   model() {
-    this.controllerFor('pipeline.events').set('pipeline', this.pipeline);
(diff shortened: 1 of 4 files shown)


tkyi (approved):

jithine (commented):
Also can you check the behavior when chainPR flag is ON ?

app/pipeline/pulls/route.js
```diff
@@ -13,7 +13,13 @@ export default EventsRoute.extend({
     this.render('pipeline.events');
   },
   model() {
-    this.controllerFor('pipeline.events').set('pipeline', this.pipeline);
+    const pipelineEventsController = this.controllerFor('pipeline.events');
```
adong, line 16: I think handles the ChainPR case.

jithine (approved):

PR #702, opened 2021-04-22
feat(2410): Add stop frozen build option

pritamstyz4ever:
## Context

Screwdriver needs a way to allow users to abort a build that has entered frozen state

## Objective

This PR adds an option to abort a build that is frozen

## References

https://github.com/screwdriver-cd/screwdriver/issues/2410

## License

### DIFF
--- a/app/components/workflow-tooltip/template.hbs
+++ b/app/components/workflow-tooltip/template.hbs
@@ -22,6 +22,9 @@
     {{#if tooltipData.displayStop}}
       <a {{action stopBuild tooltipData.selectedEvent tooltipData.job}}>Stop build</a>
     {{/if}}
+    {{#if (eq tooltipData.job.status "FROZEN")}}
+      <a {{action stopBuild tooltipData.selectedEvent tooltipData.job}}>Stop frozen build</a>
+    {{/if}}
   {{/if}}
   {{yield}}
 </div>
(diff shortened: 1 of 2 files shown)


app/components/workflow-tooltip/template.hbs
```diff
@@ -26,6 +26,9 @@
     {{#if tooltipData.displayStop}}
       <a {{action stopBuild tooltipData.selectedEvent tooltipData.job}}>Stop build</a>
     {{/if}}
+    {{#if (eq tooltipData.job.status "FROZEN")}}
```
adong, line 29: Can this be moved inside like
```
    {{#if tooltipData.displayStop}}
        <a {{action stopBuild tooltipData.selectedEvent tooltipData.job}}>Stop (if (eq tooltipData.job.status "FROZEN") "frozen" "") build</a>
    {{/if}}
```

adong: Actually, I'm not sure about this, I tried on `https://cd.screwdriver.cd/pipelines/5870`.

pritamstyz4ever: But this is only for frozen builds, which don't display stop button and we have the check for status too, wouldn't that be enough?

adong (approved):

PR #704, opened 2021-04-22
fix: add 'unsafe-inline' to script-src

adong:
## Context
add 'unsafe-inline' to script-src

## License


OronNadiv (approved):

jithine (changes requested):
Can we use https://content-security-policy.com/hash/ rather than allowing all inline scripts.

PR #712, opened 2021-05-05
fix: adjust customized label length, and show as tooltip when overflows

adong:
## License


adong:
1) Add margin between 2 buttons
image
2) Use `word-break: break-all;` to display the label instead of 
```
      text-overflow: ellipsis;
      width: 130px;
      white-space: nowrap;
      overflow: hidden;
```

DekusDenial (approved):

PR #715, opened 2021-05-21
fix: set max length for display name length

adong:
## Context

Add guard to make sure input value are in range (20 - 99)

## Objective

Now, the UI will not block users from the normal workflow, even if the data was legacy (out of range previously).
image

## License

### DIFF
--- a/app/shuttle/service.js
+++ b/app/shuttle/service.js
@@ -196,7 +196,13 @@ export default Service.extend({
     const method = 'get';
     const url = `/users/settings`;
 
-    return this.fetchFromApi(method, url);
+    try {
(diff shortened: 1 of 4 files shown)


app/shuttle/service.js
```diff
@@ -196,7 +196,13 @@ export default Service.extend({
     const method = 'get';
     const url = `/users/settings`;
 
-    return this.fetchFromApi(method, url);
+    try {
+      const userSettings = await this.fetchFromApi(method, url);
```
adong, line 200: In the catch statement, it will suppress error like: image

tkyi (approved):

PR #725, opened 2021-06-11
fix(2393): Faster list view display

ibu1224:
## Context
Fix the following issue
https://github.com/screwdriver-cd/screwdriver/issues/2393
I'll fix it in [comment 2](https://github.com/screwdriver-cd/screwdriver/issues/2393#issuecomment-809915993).

## Objective

The list view display crashes when there are many target builds.
Therefore, we will make the API fetch for list view display async.
Modify the query to specify a single `jobId` at a time

## References

- issue
  -  https://github.com/screwdriver-cd/screwdriver/issues/2393

## License


tkyi (approved):

app/pipeline/jobs/index/controller.js
```diff
@@ -93,33 +93,38 @@ export default Controller.extend(ModelReloaderMixin, {
...
+
+          if (job) {
+            nextJobDetail.jobName = job.name;
+            nextJobDetail.jobPipelineId = job.pipelineId;
+            nextJobDetail.annotations = job.annotations;
+            // PR-specific
+            nextJobDetail.prParentJobId = job.prParentJobId || null;
+            nextJobDetail.prNum = job.group || null;
+          }
+          nextJobsDetails.push(nextJobDetail);
```
yoshwata, line 123: Do you need to put a catch for this block?

ibu1224: Probably not necessary.
If you catch an error here, you can control the error message.
If an error occurs, you should see some kind of message in the ui developer tool, so you don't need to control the message.

PR #732, opened 2021-07-09
WIP: Cleanup warnings

ombharatiya:
## Context

UI tests have lots of ember & sinon.js warnings on deprecated function usage

## Objective

Cleanup Deprecated warnings

## References

https://github.com/screwdriver-cd/screwdriver/issues/2493

## License

### DIFF
--- a/app/pipeline/events/controller.js
+++ b/app/pipeline/events/controller.js
@@ -339,6 +339,9 @@ export default Controller.extend(ModelReloaderMixin, {
       }
 
       return [];
+    },
+    set(key, value) {
+      return value;
     }
   }),
   events: computed('pipelineEvents', 'prEvents', 'currentEventType', {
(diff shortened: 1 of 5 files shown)


app/pipeline/events/controller.js
```diff
@@ -390,6 +393,11 @@ export default Controller.extend(ModelReloaderMixin, {
   selectedEvent: computed('selected', 'mostRecent', {
```
adong, line 393: You could just update to 

```
selectedEvent: computed('selected', 'mostRecent', function selectedEvent() {
	// return 
    // logics
});
```

for getters only, `set` is only necessary if we actually _used_ `set`. 

This way is much easier to refactor

ombharatiya: We're using `set` in test files and that's why it's useful.

adong: Just want to point out that: `this.set('selectedEvent', someVar)` ≠ [`this.selectedEvent = someVar`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set)

Ref: https://stackoverflow.com/a/55937966

tkyi:
Closing as this PR is 3 months old
Feel free to reopen if you're still actively working on it

PR #737, opened 2021-07-29
fix(2360): fix template/command breadcrumbs

itleigns:
## Context
[willTransition](https://github.com/sonic-screwdriver-cd/ui/blob/723a7e8e89e40e8f36d69e24a1babf49a0fa3944/app/templates/route.js#L10) always throws an error because the value of `transition.params` is undefined. So template/command breadcrumbs do not work as expected.

## Objective
This PR fixes `willTransition`. We set `transition.to.params` as `newParams` which has the information of the destination url. As a result, The breadcrumbs work correctly when we click the button pointed to in the image below. 

Template

However, if we specify the namespace from the following button, the breadcrumb will not appear.

Template-2

## References
https://github.com/screwdriver-cd/screwdriver/issues/2360
https://api.emberjs.com/ember/3.27/classes/Transition/properties/to?anchor=to

## License

### DIFF
--- a/app/commands/route.js
+++ b/app/commands/route.js
@@ -8,7 +8,7 @@ export default Route.extend(AuthenticatedRouteMixin, {
   },
   actions: {
     willTransition(transition) {
-      let newParams = transition.params[transition.targetName];
+      let newParams = transition.to.params;
 
       this.controller.set('routeParams', newParams);
     }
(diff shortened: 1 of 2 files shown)


ibu1224 (approved):

adong (approved):
Nice, looks good to me!

tkyi (approved):
Thank you!

itleigns:
Thank you for the reviews!
I restart the functest and it passed.

PR #733, opened 2021-07-12
feat(2492): Update to eslint-config-screwdriver v5

tkyi:
## Objective

This PR updates eslint-config-screwdriver package to v5.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2492

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #740, opened 2021-08-11
feat: upgrade eslint-config-screwdriver to v5

adong:
## Objective
Successor PR of https://github.com/screwdriver-cd/ui/pull/733 to upgrade eslint-config-screwdriver to v5

Since screwdriver eslint config uses airbnb 
https://github.com/screwdriver-cd/eslint-config-screwdriver/blob/a997635acfec8c74222e0ec51b228b2835b261f5/index.js#L5

The [complexity](https://eslint.org/docs/rules/complexity) of airbnb base is set to be **11** at that particular version, [see discussion](https://github.com/airbnb/javascript/issues/2400), and later fix to another arbitrary number [20](https://github.com/airbnb/javascript/commit/dc3af3a90ebf56a5e9df2d4c7b40d6518200bb87#diff-6282a40af9da1a9827fe5bba99329941e9292706b9d3f6a5ffdbaffa9c11e2b7). So, maybe we need to override with our own number to fit our needs.

## License


.eslintrc.js
```diff
@@ -55,7 +55,13 @@ module.exports = {
     'ember/no-new-mixins': 'off',
     'ember/no-observers': 'off',
     'ember/no-side-effects': 'off',
-    'prettier/prettier': 'error'
+    'prettier/prettier': [
+      'error',
+      {},
+      {
+        usePrettierrc: true
```
adong, line 62: This is to use local `.prettierrc` to override space indention from 4 to 2.

See https://github.com/screwdriver-cd/eslint-config-screwdriver/blob/a997635acfec8c74222e0ec51b228b2835b261f5/index.js#L66

.prettierrc
```diff
@@ -1,6 +1,7 @@
 {
   "singleQuote": true,
+  "printWidth": 120,
+  "tabWidth": 2,
```
adong, line 4: Override tabWidth to 2 instead of https://github.com/screwdriver-cd/eslint-config-screwdriver/blob/a997635acfec8c74222e0ec51b228b2835b261f5/index.js#L66

package.json
```diff
@@ -75,7 +75,7 @@
     "d3": "^5.12.0",
     "ember-ace": "^2.0.1",
     "ember-ajax": "^5.0.0",
-    "ember-auto-import": "^1.7.0",
+    "ember-auto-import": "^2.1.0",
```
adong, line 78: Upgrade ember-auto-import to 2.x and webpack to 5.x to support split vendor.js and our code.js in source map, again. See: https://github.com/ef4/ember-auto-import/pull/315

adong:
But `complexity` is a warning, so we can leave it for now. All errors are resolved.

app/application/adapter.js
```diff
@@ -47,7 +51,7 @@ export default DS.RESTAdapter.extend(DataAdapterMixin, {
    */
   decoratePayload(key, payload) {
     if (Array.isArray(payload[key])) {
-      payload[key].map(o => this.insertLink(key, o));
+      payload[key].map((o) => this.insertLink(key, o));
```
tkyi, line 54: is there a way to turn off this rule so it's more consistent with other files?
```suggestion
      payload[key].map(o => this.insertLink(key, o));
```

klu909: set 'arrow-parens' rule to 'as-needed' https://eslint.org/docs/rules/arrow-parens#as-needed

tkyi: ^yes that one!

adong: It actually was set in the prettier: `https://prettier.io/docs/en/options.html#arrow-function-parentheses`, let me add to make sure it is turned off.

tkyi (approved):

klu909 (approved):

PR #766, opened 2022-01-10
fix: should catch server error when fetching build log

klu909:
## Context

ui calls api repeatedly on status code 500
## Objective

catch status code 500 when fetching log
## License

### DIFF
--- a/app/build-logs/service.js
+++ b/app/build-logs/service.js
@@ -57,7 +57,7 @@ export default Service.extend({
               started && jqXHR.getResponseHeader('x-more-data') === 'false';
           })
           .catch(error => {
-            if (error.jqXHR && [403, 404].includes(error.jqXHR.status)) {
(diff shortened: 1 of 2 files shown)


tkyi (approved):

adong (approved):

app/build-logs/service.js
```diff
@@ -57,7 +57,7 @@ export default Service.extend({
               started && jqXHR.getResponseHeader('x-more-data') === 'false';
           })
           .catch(error => {
-            if (error.jqXHR && [403, 404].includes(error.jqXHR.status)) {
+            if (error.jqXHR && [403, 404, 500].includes(error.jqXHR.status)) {
               done = true;
```
adong, line 61: I think we can add another flag say, 'erroredOut = true'

adong: and check `if (erroredOut == true) ` in the `finally` block to add a customized line, like 'No data from the server, please check later' to notify the end-users.

PR #767, opened 2022-01-13
fix: Hide stop button for unknown event

yoshwata:
## Context

When the event status becomes UNKNOWN for some reason, for example, if there is no build at the trigger destination, a stop button is placed for the target event.

The Stop button is confusing to the user because no build is actually running.

## Objective

Hides the STOP button for events in the UNKNOWN state.

Before:

スクリーンショット 2022-01-13 13 23 02

After:
スクリーンショット 2022-01-13 13 22 05

## License

### DIFF
--- a/tests/integration/components/pipeline-event-row/component-test.js
+++ b/tests/integration/components/pipeline-event-row/component-test.js
@@ -51,7 +51,8 @@ const event = {
     { jobId: 1, id: 4, status: 'SUCCESS' },
     { jobId: 2, id: 5, status: 'SUCCESS' },
     { jobId: 3, id: 6, status: 'FAILURE' }
-  ]
+  ],
+  isRunning: false
 };
 
 module('Integration | Component | pipeline event row', function (hooks) {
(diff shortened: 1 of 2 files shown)


tests/integration/components/pipeline-event-row/component-test.js
```diff
@@ -100,6 +101,89 @@ module('Integration | Component | pipeline event row', function (hooks) {
...
+    this.actions.eventClick = () => {
+      assert.ok(true);
+    };
+
+    this.set('stopPRBuilds', Function.prototype);
+    this.set('stopEvent', Function.prototype);
+
+    let eventMock = EmberObject.create(copy(event, true));
+
+    eventMock.isRunning = false;
```
y-oksaku, line 156: I think `isRunning` has to be `true` here, and need to check a stop button is hidden in [the test above](https://github.com/screwdriver-cd/ui/blob/master/tests/integration/components/pipeline-event-row/component-test.js#L90).

yoshwata: light, I've fixed.

y-oksaku (approved):

itleigns (approved):

wahapo (approved):

tkyi (approved):

PR #773, opened 2022-03-22
fix(2676): Update validator with provider field

tkyi:
## Context

Provider field is not showing up in validator yet.

## Objective

This PR adds provider field to validator.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/2676

## License


adong (dismissed):

app/components/validator-job/template.hbs
```diff
@@ -154,4 +154,16 @@
     </ul>
   </span>
 </div>
+<div class="provider" title="This is the cloud provider you have added to the job for AWS Native builds.">
```
jithine, line 157: perhaps it should say, this is the ...AWS cloud provider configuration...
Also shouldn't this be optional ? only if `job.provider` exist

tkyi: I was just following what the other ones look like
A lot of them are also optional but show up by default on the right

I can change wording, but I think it's a given that it's configuration, right?

jithine (approved):

app/components/validator-job/template.hbs
jithine: so I think eventually when we support multiple providers, it might make sense to just call it provider 🤔

PR #785, opened 2022-05-05
fix(2707): build not found redirects to intended pipeline page with an error message

banusha2:
## Context

 As per the requirement user must redirects to intended pipeline page with an error message "Build does not exist". This will let user know that build doesn't exist.

## Objective
User must redirects to the intended pipeline page with an error message"Build does not exist".
screenshot

## References
Related to https://github.com/screwdriver-cd/screwdriver/issues/2707

## License

### DIFF
--- a/app/404/route.js
+++ b/app/404/route.js
@@ -13,7 +13,7 @@ export default Route.extend({
       const pipelineId = getWithDefault(
         this.router.currentRoute.parent.parent.params,
         'pipeline_id',
-        '/'
+        ''
       );
 
       later(
--- a/app/404/route.js
+++ b/app/404/route.js
@@ -1,28 +1,3 @@
 import Route from '@ember/routing/route';
-import { inject as service } from '@ember/service';
-import { later } from '@ember/runloop';
-import { getWithDefault } from '@ember/object';
 
-export default Route.extend({
-  router: service(),
-
-  beforeModel() {
-    const { name } = this.router.currentRoute;
-
-    if (name === 'pipeline.build.index') {
-      const pipelineId = getWithDefault(
-        this.router.currentRoute,
-        'parent.parent.params.pipeline_id',
-        ''
-      );
-
-      later(
-        this,
-        function redirection() {
-          this.transitionTo('pipeline', pipelineId);
-        },
-        3000
-      );
-    }
-  }
-});
+export default Route.extend({});
--- a/app/404/route.js
+++ b/app/404/route.js
@@ -1,3 +1,22 @@
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { later } from '@ember/runloop';
 
-export default Route.extend({});
+export default Route.extend({
+  router: service(),
+
+  beforeModel() {
+    const { name } = this.router.currentRoute;
+    const { pipelineId } = this.router.currentRoute.parent.parent.params;
+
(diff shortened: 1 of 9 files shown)


app/pipeline/route.js
```diff
@@ -29,6 +30,9 @@ export default Route.extend(AuthenticatedRouteMixin, {
         error.errors[0].status === 404
       ) {
         this.transitionTo('/404');
+        later(function () {
+          window.location.href = 'http://cd.screwdriver.cd/';
```
adong, line 34: I get what you are trying to do, but this file is `app/pipeline/route.js` not the build route file. 
1) Good for using `later` instead of `setTimeout`
2) Do not hard code the redirect url
3) Do not use `window.location.href`. Two reasons: 1) this is a browser hard fresh 2) SPA will lose context and reload all web resources again, which is unnecessary.

adong (changes requested):

app/404/route.js
```diff
@@ -1,3 +1,22 @@
...
+import { inject as service } from '@ember/service';
+import { later } from '@ember/runloop';
 
-export default Route.extend({});
+export default Route.extend({
+  router: service(),
+
+  beforeModel() {
+    const { name } = this.router.currentRoute;
+    const { pipelineId } = this.router.currentRoute.parent.parent.params;
```
adong, line 10: When `this.router.currentRoute.parent` is undefined, it will crash the UI. 
Please use ember getWithDefault to avoid UI crash.

adong (changes requested):
Please add some context in the UI PR description, and revert `app/pipeline/route.js` changes

app/404/route.js
```diff
@@ -1,3 +1,28 @@
...
 
-export default Route.extend({});
+export default Route.extend({
+  router: service(),
+
+  beforeModel() {
+    const { name } = this.router.currentRoute;
+
+    if (name === 'pipeline.build.index') {
+      const pipelineId = getWithDefault(
```
adong, line 13: I would suggest change this

```
const pipelineId = getWithDefault(
  this.router.currentRoute.parent.parent.params,
  'pipeline_id',
  '/'
);
```


To
```
const pipelineId = getWithDefault(
  this.router.currentRoute,
  'parent.parent.params.pipeline_id',
  ''
)
```

tests/unit/pipeline/build/route-test.js
```diff
@@ -126,4 +128,24 @@ module('Unit | Route | pipeline/build', function (hooks) {
...
+      const errorRroute = this.owner.lookup('route:404');
+      const pipelineindexRoute = this.owner.lookup(
+        'route:pipeline/events/show'
+      );
+      const buildId = 34565845;
+      const pipelineId = 9373;
+
+      await visit(`/pipelines/${pipelineId}/builds/${buildId}`);
+      clock.tick(100);
+      assert.ok(errorRroute);
```
adong, line 145: Hmm, I think 

```
assert.ok(errorRroute);
assert.ok(pipelineindexRoute);
```

These are checking the route exist or not, in this case, it always will be true, and it does not check the the route changes after hitting errors. 
You can either 
1) check the route url changes
2) or check the displayed text

adong (dismissed):

adong:
Looks good to me, please add the test back, and rebase with latest master.

adong (approved):

PR #787, opened 2022-05-17
feat: add sort by failed builds and Last Run Job column to dashboard collection

banusha2:
## Context
Feature addition to dashboard collection to enhance the usability

## Objective
Add sorting criteria based on ascending or descending order for the number of failed builds (number of red bars)
Create a column after Duration, called "Last Run Job", and sorting criteria based on ascending or descending orders.

## References
Refer to: https://github.com/screwdriver-cd/screwdriver/issues/2713
## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #790, opened 2022-06-01
feat: Add Alias column for the pipeline name

banusha2:
## Context
 add a column alias for the renaming of the pipeline by the admin in preferred names which fits their pipeline.
            
## Objective
 To add an alias column to customize pipelines name
 Re-name can be shown under the options tab for pipeline level preference

## References
https://github.com/screwdriver-cd/screwdriver/issues/2710

## License


adong:
I checked out the branch before, can you attach screenshots for others?
Though it should ideally be something more than a text field, like a confirmation button, for now, that should be fine. 


Please add tests.

app/components/collection-table-row/component.js
```diff
@@ -20,6 +20,12 @@ export default Component.extend({
   hasBothEventsAndLatestEventInfo: and('eventsInfo', 'lastEventInfo'),
   showCheckbox: and('isOrganizing', 'isAuthenticated'),
 
+  aliasName: computed('pipeline', function get() {
+    const aliasName = this.pipeline.settings?.aliasName;
```
adong, line 24: Kudos for using `?.`!

app/components/collection-table-row/component.js
```diff
@@ -20,6 +20,12 @@ export default Component.extend({
...
+    const { name } = this.pipeline.scmRepo;
+
+    return aliasName || name;
```
adong, line 27: Hmm, please use `getWithDefault('aliasName', name);` over `||`

adong: For your reference, [getWithDefault](https://api.emberjs.com/ember/3.15/functions/@ember%2Fobject/getWithDefault)

app/components/collection-table-row/template.hbs
```diff
@@ -3,6 +3,9 @@
     <input type="checkbox" value={{pipelineSelected}} onchange={{action "togglePipeline"}}>
   {{/if}}
 </td>
+<td class="alias">
```
adong, line 6: Hmm, its better to check if `alias` column is needed, if this collection dashboard does not have any pipeline that has alias feature, then we don't need to show the alias column, at least the current way is showing pipeline `name` again if `aliasName` is not set. I consider it is too much duplicate info

app/components/collection-view/template.hbs
```diff
@@ -149,6 +149,7 @@
           <thead>
             <tr>
               <th class="collection-pipeline__choose"></th>
+              <th class="alias" rowspan="1">Alias</th>
```
adong, line 152: Hmm, its better to check if alias column is needed, if this collection dashboard does not have any pipeline that has alias feature, then we don't need to show the alias column, at least the current way is showing pipeline name again if aliasName is not set. I consider it is too much duplicate info

app/components/pipeline-options/component.js
```diff
@@ -278,6 +296,13 @@ export default Component.extend({
 
       debounce(this, this.updateJobNameLength, displayJobNameLength, 1000);
     },
+    async updatePipelineAlias(inputPipelinAlias) {
+      let aliasName = inputPipelinAlias;
```
adong, line 300: Typo

app/pipeline/events/route.js
```diff
@@ -44,7 +44,8 @@ export default Route.extend({
         this.pipeline,
         'settings.filterEventsForNoBuilds',
         false
-      )
+      ),
+      aliasName: getWithDefault(this.pipeline, 'settings.aliasName', 'appId')
```
adong, line 48: What is `appId`? Was it put here accidentally for debugging? Please avoid magic string

adong (changes requested):
Please make changes according to PR comments

app/components/collection-table-row/component.js
banusha2: `getWithDefault` because for some pipeline aliasName field doesn't exists. and I think getWithDefault checks if field is falsy if yes then return the alternate. but I have fixed the logic from `| |` to ternary operator. I used this before Idk what overwritten that with `| |`

adong: The ternary operator is worse than `||` in my opinion.  And I'm not talking about that, I genuinely think there's a better way of writing it. 

Let me give you one example. 

From
```javascript
    const aliasName = this.pipeline.settings?.aliasName;
    const { name } = this.pipeline.scmRepo;

    return aliasName || name;
```
To
```javascript
    const { name } = this.pipeline.scmRepo;

	return getWithDefault(this.pipeline.settings, 'aliasName', name);
```

banusha2: I tried exactly same what you are suggesting and it gave error below. And I think that's because `aliasName` field does not exists for all the pipeline. 

Screen Shot 2022-07-21 at 11 36 44 AM

adong: I just merged the latest master to this branch, I believe this issue would go away. 

`app/pipeline/model.js` already had default value for `pipeline.settings`. 
```
  settings: DS.attr({
    defaultValue() {
      return {
        groupedEvents: true,
        showEventTriggers: false
      };
    }
  }),
```

Particularly this change on `app/collection/model.js` by PR: https://github.com/screwdriver-cd/ui/pull/787, should fix it by establishing the relationship between collection to pipeline

```diff
export default DS.Model.extend({
  name: DS.attr('string'),
  description: DS.attr('string'),
  type: DS.attr('string'),
  pipelineIds: DS.attr(),
--  pipelines: DS.attr(),
++  pipelines: DS.hasMany('pipeline', { sync: true })
});
```


Test link: https://cd.screwdriver.cd/dashboards/4501

banusha2: yeah that helped. Thanks

app/components/collection-view/template.hbs
banusha2: We can hide it for collection where not a single pipeline have a aliasname if admin add's aliasName to single pipeline. Then we will again have duplicate data. 
How about keeping it empty string by default instead of pipeline name?

banusha2:
### added ellipsis and tooltip when hovering over aliasName
Screen Shot 2022-07-25 at 1 46 58 PM

### hide aliasName coloumn if not a single pipeline in collection has a aliasName. 
Screen Shot 2022-07-25 at 1 50 43 PM

app/components/collection-view/component.js
```diff
@@ -114,6 +114,22 @@ export default Component.extend({
     }
   }),
 
+  isAliasName: computed('collection.pipelines', {
```
adong, line 117: Cool, can we change from `isAliasName` to `hasAliasName` for this boolean?

app/components/collection-view/component.js
```diff
@@ -114,6 +114,22 @@ export default Component.extend({
...
+    get() {
+      let isAliasName;
+
+      this.sortedPipelines.forEach(element => {
```
adong, line 121: I think I know what this logics trying to do, 

1) The boolean value here will be overridden and totally depends on the last element of the array. If the last element array does not have `aliasName` even though the previous one has, then this boolean becomes `false`. Which is undesired. 

2) Instead of `forEach` use `[some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some]` for array. I think that ought to cover that case.

adong (changes requested):
Add more PR comments

app/components/collection-view/component.js
adong: 3) Instead of using `sortedPipelines`, can we use `collectionPipelines`? Because the `sortedPipelines` like the name suggests, should update when sorting criteria changes `sortedPipelines: sort('collectionPipelines', 'sortBy'),
`, while `collectionPipelines` should remain unchanged. So using `collectionPipelines` will have less property change, and is more efficient.

banusha2: yeah, changed to `some`

app/components/pipeline-options/component.js
```diff
@@ -160,6 +164,20 @@ export default Component.extend({
...
   },
+  async updatePipelineAlias(aliasName) {
+    const pipeline = this.get('pipeline');
+
+    try {
+      await this.shuttle.updatePipelineSettings(pipeline.id, {
+        aliasName
+      });
+    } finally {
+      pipeline.set('settings.aliasName', aliasName);
```
adong, line 175: Hmm, I dont think `pipeline.set('settings.aliasName', aliasName);`, is not needed, otherwise we never know if it was set thru api or this line.

app/components/collection-view/component.js
```diff
@@ -114,6 +114,18 @@ export default Component.extend({
     }
   }),
 
+  hasAliasName: computed('collection.pipelines', {
+    get() {
+      let isAliasName = this.collectionPipelines
```
adong, line 119: Please change the name `isAliasName` to `hasAliasName`, it would make easier to read.

app/components/pipeline-options/template.hbs
```diff
@@ -241,6 +241,22 @@
             </div>
           </div>
         </li>
+        <li>
+          <div class="row">
+            <div class="col-xs-10">
+              <h4>Re-name pipeline</h4>
+              <p>Setup your own preferred pipeline name to display in the pipeline list.</p>
```
adong, line 248: "...in the dashboard list view."

cc @tkyi see there's need to update the wordings

tests/integration/components/pipeline-options/component-test.js
```diff
@@ -504,6 +504,60 @@ module('Integration | Component | pipeline options', function (hooks) {
...
+
+    await render(hbs`{{pipeline-options pipeline=mockPipeline }}`);
+
+    assert
+      .dom('section.preference li:nth-of-type(4) h4')
+      .hasText('Re-name pipeline');
+    assert
+      .dom('section.preference li:nth-of-type(4) p')
+      .hasText(
+        'Setup your own preferred pipeline name to display in the pipeline list.'
```
adong, line 552: "...in the dashboard list view."

tests/integration/components/pipeline-options/component-test.js
```diff
@@ -504,6 +504,60 @@ module('Integration | Component | pipeline options', function (hooks) {
...
+      );
+    assert.dom('section.preference li:nth-of-type(4) input').hasNoText();
+
+    await fillIn('section.preference li:nth-of-type(4) input', 'test-pr');
+
+    assert.dom('section.preference li:nth-of-type(4) input').hasText('test-pr');
```
adong, line 558: Change from `hasText` to `hasValue` should work

adong (changes requested):
More PR comments

adong (approved):
LGTM

PR #808, opened 2022-08-08
fix(2728): gray out the Restart button for jobs that are disabled on the build details page

yk634:
## Context

As per the comment in https://github.com/screwdriver-cd/models/pull/555#pullrequestreview-1047094026, it would be preferable to have a UI where disabled jobs cannot be Restarted from the Build Details page.

## Objective

Fix the Restart button to be grayed out on the build details page of a disable job.
- While waiting for `model.pipeline.jobs` to fetch, isButtonDisabledLoaded is false and the Restart button is grayed out regardless of its state.
  - The time to wait for this fetch does not occur when transitioning from tooltip, but occurs slightly when the build detail page is accessed directly.

## License

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -72,6 +72,23 @@ export default Component.extend({
     }
   }),
 
+  isButtonDisabledLoaded: false,
+  isButtonDisabled: computed('buildStatus', 'jobDisabled', {
+    get() {
+      if (this.buildAction === 'Restart') {
+        return this.jobDisabled.then(jobDisabled => {
+          this.set('isButtonDisabledLoaded', true);
+
+          return jobDisabled;
+        });
+      }
+
+      this.set('isButtonDisabledLoaded', true);
+
+      return false;
+    }
+  }),
+
   overrideCoverageInfo() {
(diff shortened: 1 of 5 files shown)


jithine (approved):
This is great. Can we extend this to `Start` button as well ? Can UI determine, if there are no jobs to start then grey it out ? (It's scenario 1 here https://github.com/screwdriver-cd/screwdriver/issues/2749#issue-1330351159)

tkyi (approved):

y-oksaku (approved):

kumada626 (approved):

yk634:
@jithine 
> This is great. Can we extend this to `Start` button as well ? Can UI determine, if there are no jobs to start then grey it out ? (It's scenario 1 here [screwdriver-cd/screwdriver#2749 (comment)](https://github.com/screwdriver-cd/screwdriver/issues/2749#issue-1330351159))

I certainly think it's a good idea to have the Start button grayed out as well.

However, to gray out the `Start` button would require a different logic than this PR and would be difficult to extend immediately.

ibu1224 (approved):

PR #810, opened 2022-08-09
fix: Pipeline collection table show non-intuitive cursor when hover over headers

banusha2:
## Context

Pipeline collection table show non-intuitive cursor when hover over headers
## Objective
When mouse hover over Last Run Job and Build History table header, mouse cursor should be a hand to indicate sorting, while other headers should be the default instead of text
## References
https://github.com/screwdriver-cd/screwdriver/issues/2746

## License


tkyi (approved):
Try to include screenshots in future PRs

PR #822, opened 2022-09-06
fix: Move job name length user settings

banusha2:
## Context

To display JobName length and save/reset in User Settings.

## Objective
Screen Shot 2022-09-06 at 1 18 33 PM

## License


app/components/user-setting/component.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  },
+  actions: {
+    async updateUserSettings() {
+      this.updateUserSettings();
+    },
+    async resetUserSettings() {
+      this.set('isSaving', true);
+      try {
+        await this.shuttle.deleteUserSettings();
```
adong, line 55: Can you implement it in the [`deleteRecord`](https://api.emberjs.com/ember-data/3.16/classes/Store/methods/deleteRecord?anchor=deleteRecord)
See 
https://github.com/screwdriver-cd/ui/blob/master/app/preference/pipeline/adapter.js#L21

as refererence

adong: This way, the records in the ember data store will be in sync.

app/components/user-setting/style.scss
```diff
@@ -0,0 +1,184 @@
...
+  background: transparent;
+
+  &:hover {
+    color: $sd-link;
+    @include scale(1);
+  }
+  &:disabled {
+    background: $sd-text-med;
+  }
+  .saving-loading {
```
adong, line 168: Can we handle the show & hide logic in the `template.hbs`? It is not easy to debug if logics are handled in CSS.

app/user-settings/preferences/template.hbs
```diff
@@ -1,23 +1 @@
...
-              value={{desiredJobNameLength}}
-              onchange={{action "updateJobNameLength" value="target.value"}}>
-          </div>
-        </div>
-      </li>
-    </ul>
-  </section>
-</div>
\ No newline at end of file
+{{user-setting}}
```
adong, line 1: I have some reservations of converting the code into a `component`, because component will be re-render each time we enter this route (which will  make API to fetch the user preference everytime), while controller's `template.hbs` will only render once (since the controller is a singleton)

app/components/user-setting/component.js
adong: And it's easier to add tests too.

- https://github.com/screwdriver-cd/ui/blob/master/tests/unit/preference/user/serializer-test.js#L57
- https://github.com/screwdriver-cd/ui/blob/master/tests/unit/shuttle/service-test.js

app/user-settings/preferences/template.hbs
banusha2: I thought that in future if we want to add few more sections in user preferences it will be easy to do so. And adding css to route level `template.hbs`, can break other css with same classnames. I am not sure on this. what do you think?

app/components/user-setting/style.scss
banusha2: ok will do it.

app/user-settings/preferences/template.hbs
adong: Conceptionally, if some code would be re-used, then we make it into a component. 
For example, the access-token component is being used in multiple places. One here, and another one used in `pipeline -> options -> access`

adong: Could this component be re-named to say, `user-preference` instead of `user-setting`?

banusha2: Okay, so should i move the code in component to `preference/template.hbs`

adong: Actually, I thought about and tested out, I still prefer to see it in the `app/user-settings/preferences/controller.js` controller, see reasons in https://github.com/screwdriver-cd/ui/pull/822/files#r966186063

app/components/user-preference/component.js
```diff
@@ -0,0 +1,70 @@
...
+
+    if (userSetting) {
+      desiredJobNameLength = userSetting.displayJobNameLength;
+    }
+
+    this.setProperties({ desiredJobNameLength });
+  },
+
+  async updateUserSettings() {
+    this.set('isSaving', true);
```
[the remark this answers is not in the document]
banusha2: oh got it, shifting code to controller's `template.hbs`

app/shuttle/service.js
```diff
@@ -325,6 +325,20 @@ export default Service.extend({
...
 
+  /**
+   * deleteUserSettings
+   * @return {Promise}
+   */
+  async deleteUserSettings() {
+    const method = 'delete';
+    const url = `/users/settings`;
+    const data = {
+      settings: {}
```
adong, line 336: I also felt the payload `data = {settings: {}}` code is not necessary according to the backend, which does not take the payload.

https://github.com/screwdriver-cd/screwdriver/pull/2762/files#diff-ff0d529733cb812e0466575567bf1fee2f488a9dfc27fafdba520ffacb3a9a7aR26

adong: You can remove data, and just call `this.fetchFromApi(method, url)`, I think the `fetchFromApi(method, url)` will add a default value for `data`, which is `{}`

app/components/user-preference/component.js
```diff
@@ -0,0 +1,70 @@
...
+
+    let desiredJobNameLength = MINIMUM_JOBNAME_LENGTH;
+
+    const userSetting = await this.userSettings.getUserPreference();
+
+    if (userSetting) {
+      desiredJobNameLength = userSetting.displayJobNameLength;
+    }
+
+    this.setProperties({ desiredJobNameLength });
```
adong, line 29: Need to store `userSettings` here in `init` function, so in `updateUserSettings` and `resetUserSettings`, you can just use `this.userSettings` without call `const userSetting = await this.userSettings.getUserPreference();`

app/components/user-preference/component.js
```diff
@@ -0,0 +1,70 @@
...
+  },
+  actions: {
+    async updateUserSettings() {
+      this.updateUserSettings();
+    },
+    async resetUserSettings() {
+      this.set('isSaving', true);
+
+      try {
+        await this.store.deleteRecord('preference/user');
```
adong, line 56: Need to change to await `this.store.deleteRecord(this.userSetting)`, assuming this property is saved in `init` https://github.com/screwdriver-cd/ui/pull/822/files#r966309833

app/components/user-preference/style.scss
```diff
@@ -0,0 +1,184 @@
+& {
```
adong, line 1: Please refactor this back to a controller.

adong (changes requested):
See comments.

app/user-settings/preferences/controller.js
```diff
@@ -1,50 +1,71 @@
...
   maxDisplayLength: MAXIMUM_JOBNAME_LENGTH,
+  isDisabled: or('isSaving', 'isInvalid'),
 
   async init() {
     this._super(...arguments);
+    let desiredJobNameLength = MINIMUM_JOBNAME_LENGTH;
 
-    const desiredJobNameLength =
-      await this.userSettings.getDisplayJobNameLength();
+    const userSetting = await this.userSettings.getUserPreference();
```
adong, line 21: Can we rename `userSetting` to `userSettings` just for the consistency's sake, since it is `this.userSettings`?

app/user-settings/preferences/controller.js
```diff
@@ -1,50 +1,71 @@
...
+      this.setProperties({
+        isSaving: false,
+        successMessage: 'User settings updated successfully!'
+      });
+    } catch (error) {
+      this.setProperties({
+        isSaving: false,
+        errorMessage: error
+      });
+    }
```
adong, line 46: Can you refactor 
```
this.setProperties({
        isSaving: false,
```

To `finally` block with `this.set('isSaving', false);`?

app/user-settings/preferences/controller.js
```diff
@@ -1,50 +1,71 @@
...
-        displayJobNameLength = MINIMUM_JOBNAME_LENGTH;
+        this.setProperties({
+          isSaving: false,
+          successMessage: 'User settings reset successfully!'
+        });
+      } catch (error) {
+        this.setProperties({
+          isSaving: false,
+          errorMessage: error
+        });
```
adong, line 67: same here, add a finally block and for `this.set('isSaving', false);`

app/user-settings/preferences/controller.js
adong: Actually, it should not be called `userSetting` nor `userSettings`, rather than `userPreference`

adong:
This PR needs to wait until https://github.com/screwdriver-cd/screwdriver/pull/2762/files merged

app/styles/user-settings.scss
```diff
@@ -0,0 +1,184 @@
...
+  &:focus {
+    outline: none;
+  }
+  &:hover {
+    background: $sd-link;
+  }
+  &:disabled {
+    background: $sd-text-med;
+  }
+  .saving-loading {
```
adong, line 133: Deja vu, I thought I commented on this before. 🤔 

Anyway, please handle the show/hide logics in `template.hbs` rather than css, css is not easy to debug.

app/styles/user-settings.scss
```diff
@@ -0,0 +1,184 @@
...
+    display: none;
+  }
+  &.saving {
+    .saving-loading {
+      display: block;
+    }
+    .button-label {
+      display: none;
```
adong, line 176: Similar here

app/user-settings/preferences/template.hbs
```diff
@@ -11,13 +13,29 @@
...
         </div>
       </li>
+      <li>
+        <div class="row">
+          <div class="col-xs-12 right">
+            <button {{action "resetUserSettings"}} disabled={{isDisabled}} class="reset-button{{if isSaving " saving"}}">
+              {{fa-icon "refresh"}} Reset
+            </button>
+            <button {{action "updateUserSettings"}} disabled={{isDisabled}} class="blue-button{{if isSaving " saving"}}">
+              <div class="saving-loading">
```
adong, line 30: `if/else` in js over `display: none;` in css.

app/styles/user-settings.scss
banusha2: Fixed

app/styles/user-settings.scss
banusha2: fixed

app/user-settings/preferences/template.hbs
banusha2: Got it. I have fixed it now

app/user-settings/preferences/controller.js
```diff
@@ -1,50 +1,62 @@
...
 
-    this.setProperties({ desiredJobNameLength });
-  },
+    if (userPreferences) {
+      desiredJobNameLength = userPreferences.displayJobNameLength;
+    }
 
-  async updateJobNameLength(displayJobNameLength) {
-    this.setProperties({ displayJobNameLength });
+    this.setProperties({ desiredJobNameLength, userPreferences });
```
adong, line 27: Please add initialization of `successMessage, errorMessage, isSaving, isValid` properties in the init() function

adong (changes requested):
Please add tests, otherwise, looks good to me.

app/user-settings/preferences/controller.js
```diff
@@ -1,50 +1,62 @@
...
-      if (parseInt(displayJobNameLength, 10) < MINIMUM_JOBNAME_LENGTH) {
-        displayJobNameLength = MINIMUM_JOBNAME_LENGTH;
+    async updateUserSettings() {
+      this.updateUserSettings();
+    },
+    async resetUserSettings() {
+      this.set('isSaving', true);
+
+      try {
+        await this.store.deleteRecord(this.userPreferences);
```
adong, line 52: I dont think `deleteRecord` returns a promise, so no await is needed for this one. Just needs to await on the `save()`
See https://github.com/emberjs/data/blob/v4.6.1/packages/model/addon/-private/model.js#L671

adong (approved):
LGTM

PR #823, opened 2022-09-14
feat: Add timestamp options in user preferences

banusha2:
## Context
When user looks at events, getting several events happened "2 months ago." When looking  at the list view, confused with timestamp such as "August 18, 2020 4:52 PM."

## Objective
To add Timestamp options under User Preferences -> 1) UTC Timestamp, 2) Local Time Zone timestamp.

image

Screenshot: 
UTC 
image

Local Time
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/2218

## License


app/user-settings/preferences/controller.js
```diff
@@ -52,6 +81,9 @@ export default Controller.extend({
       this.set('isSaving', true);
 
       try {
+        await this.store.deleteRecord(this.userPreferences);
+        await this.userPreferences.save();
```
adong, line 85: I was going to comment, but realized this file might be out dated, because `unload` is missing.

app/user-settings/preferences/controller.js
```diff
@@ -21,10 +28,32 @@ export default Controller.extend({
     this._super(...arguments);
     let desiredJobNameLength = MINIMUM_JOBNAME_LENGTH;
 
+    let selectedTimestampFormat = this.timestampOptions.find(
```
adong, line 31: If we are hard coding to be `LOCAL_TIMEZONE`, then this could be written as `this.timestampOptions[1]`, or `this.get('timestampOptions', 'lastObject');` latter seems to be more elegant, but they are the same thing essentially.

app/user-settings/service.js
```diff
@@ -68,6 +69,22 @@ export default Service.extend({
     return desiredJobNameLength;
   },
 
+  async getTimestampFormat() {
+    let desiredTimestampFormat = 'LOCAL_TIMEZONE';
```
adong, line 73: Hmm, this actually is the part that makes what the default value is for timestamp format. Can we not hard code the `LOCAL_TIMEZONE` string?  There has to be a better way. I need to think about it.

app/user-settings/preferences/controller.js
```diff
@@ -16,23 +16,42 @@ export default Controller.extend({
   isDisabled: bool('isSaving'),
   successMessage: '',
   errorMessage: '',
+  selectedTimestampFormat: {},
+  timestampOptions: [
```
adong, line 20: I would make this into model.js, see other comments

app/preference/user/model.js
```diff
@@ -3,5 +3,7 @@ import Model, { attr, hasMany } from '@ember-data/model';
 export default class PreferenceUserModel extends Model {
   @attr('number') displayJobNameLength;
 
+  @attr('string') timestampFormat;
```
adong, line 6: This should have a default value, say

```
export const TIMESTAMP_OPTIONS = ['option1', 'option2'];
export const TIMESTAMP_DEFAULT_OPTION = 0;
...
  @attr('string', { defaultValue:  TIMESTAMP_OPTIONS[TIMESTAMP_DEFAULT_OPTION] }) timestampFormat;
```

So, other classes can import it, like `app/user-settings/service.js` can re-use it without dictating the default timezone value, and we have a central place to adjust what the default timezone value is.

app/user-settings/preferences/controller.js
adong: Actually, I don't have a strong feeling about list these `timestampOptions` here, I just think we should keep `app/user-settings/service.js` away from hard-coding default values.

app/user-settings/preferences/controller.js
```diff
@@ -16,23 +17,39 @@ export default Controller.extend({
...
   errorMessage: '',
+  selectedTimestampFormat: {},
+  timestampOptions: TIMESTAMP_OPTIONS,
 
   async init() {
     this._super(...arguments);
+
     let desiredJobNameLength = MINIMUM_JOBNAME_LENGTH;
 
+    let selectedTimestampFormat = this.get('timestampOptions', 'lastObject');
```
adong, line 28: Hmm, I get what this trying to do, but the syntax is not correct.
image


Besides, since `TIMESTAMP_DEFAULT_OPTION` is `0`, and `lastObject` usually isn't `0`. So this can be replaced with `firstObject`. 
```
this.get(`timestampOptions.firstObject`);
```

If really wants to make it modular, it ought to be

```
this.get(`timestampOptions.${TIMESTAMP_DEFAULT_OPTION}`)
```

I prefer this one ☝️

image

app/utils/timestamp-format.js
```diff
@@ -0,0 +1,6 @@
+export const TIMESTAMP_OPTIONS = [
+  { key: 'LOCAL_TIMEZONE', name: 'Local Timezone' },
+  { key: 'UTC', name: 'UTC' }
```
adong, line 3: Why is the variable inside the object called `key`, is it used in other 3rd party conventions? If it's not required, please change it to `value`, because `key` was confusing.

app/preference/user/model.js
```diff
@@ -1,7 +1,16 @@
...
+import {
+  TIMESTAMP_DEFAULT_OPTION,
+  TIMESTAMP_OPTIONS
+} from '../../utils/timestamp-format';
 
 export default class PreferenceUserModel extends Model {
   @attr('number') displayJobNameLength;
 
+  @attr('string', {
+    defaultValue: TIMESTAMP_OPTIONS[TIMESTAMP_DEFAULT_OPTION].key
```
adong, line 11: This is the part that confused me, assign key to defaut`Value`. I believe `value` is more appropriate and easy to read.
```diff
--defaultValue: TIMESTAMP_OPTIONS[TIMESTAMP_DEFAULT_OPTION].key
++defaultValue: TIMESTAMP_OPTIONS[TIMESTAMP_DEFAULT_OPTION].value
```

adong:
For this particular PR, I only see pipeline list view show different timestamp formats, per https://github.com/screwdriver-cd/screwdriver/issues/2218 request, I think the user wants to see the event list view to have utc/local time zone

image

app/user-settings/preferences/controller.js
adong: Somehow the test case didn't catch it. I guess either the test was either:

1) `selectedTimestampFormat` got overridden by `userPreference`'s timestamp
2) there's no test case was covering the `userPreference` is falsy case.


_Please test more thoroughly._

adong (changes requested):

app/utils/timestamp-format.js
banusha2: Changed

app/user-settings/preferences/controller.js
banusha2: sure adding more tests

app/components/pipeline-event-row/component.js
```diff
@@ -78,6 +82,25 @@ export default Component.extend({
       return this.get('numberOfParameters') < MAX_NUM_OF_PARAMETERS_ALLOWED;
     }
   }),
+  startDate: computed('event.createTime', function () {
+    let startDate;
+
+    const userPreferences = this.store.peekAll('preference/user');
```
adong: The `userPreference` would be empty if `preference/user` does not load before viewing this page, so, better user 
```
const userPreference = await this.getUserPreference();
```

which handles that gracefully in the userSetting service.

app/components/pipeline-graph-nav/component.js
```diff
@@ -38,5 +41,24 @@ export default Component.extend({
     get() {
       return statusIcon(this.get('selectedEventObj.status'));
     }
+  }),
+  startDate: computed('selectedEventObj.createTime', function get() {
+    let startDate;
+
+    const userPreferences = this.store.peekAll('preference/user');
```
adong, line 48: Same thing here.

```
const userPreference = await this.getUserPreference();
```



Another comment is, these two code blocks are very similar, wonder if there's a way to DRY it.

adong:
Can you add tests for
1) event-list page with 
	* Local timezone and
	* UTC
2) pipeline-graph-nav
	* Local timezone and
	* UTC
3) job list view
	* Local timezone and
	* UTC

app/components/pipeline-event-row/component.js
```diff
@@ -78,6 +82,30 @@ export default Component.extend({
       return this.get('numberOfParameters') < MAX_NUM_OF_PARAMETERS_ALLOWED;
     }
   }),
+  startDate: computed('event.createTime', {
+    get() {
+      let startDate;
+
+      const userPreferences = this.store.peekAll('preference/user');
+      const timestampPreference = get(
```
adong, line 90:
```
const userPreference = await this.getUserPreference();
```

app/components/pipeline-event-row/component.js
```diff
@@ -78,6 +86,30 @@ export default Component.extend({
       return this.get('numberOfParameters') < MAX_NUM_OF_PARAMETERS_ALLOWED;
     }
   }),
+  startDate: computed('event.createTime', {
+    get() {
+      return ObjectPromiseProxy.create({
+        promise: this.userSettings.getUserPreference().then(userPreference => {
+          let startDate = 'n/a';
+
+          const timestampPreference = get(userPreference, 'timestampFormat');
```
jithine, line 95: this code block is a good candidate for a helper function. perhaps here https://github.com/screwdriver-cd/ui/blob/master/app/utils/time-range.js
 I see this chunk of code repeated frequently

banusha2: Yes,  done

app/utils/timestamp-format.js
```diff
@@ -0,0 +1,40 @@
...
+  { value: 'UTC', name: 'UTC' }
+];
+
+export const TIMESTAMP_DEFAULT_OPTION = 0;
+
+/**
+ * Return custom locale string for date
+ * @export
+ * @param {string} timestampPreference timestamp preferred e.g UTc
+ * @param {string} startDate input date
```
tkyi, line 19:
```suggestion
 * @param {Object} userSettings  User settings
 * @param {String} startDate         Input date
```

app/user-settings/service.js
```diff
@@ -68,6 +69,22 @@ export default Service.extend({
     return desiredJobNameLength;
   },
 
+  async getTimestampFormat() {
+    let desiredTimestampFormat;
```
tkyi, line 73: should you set a default?

banusha2: Default is set in model. in our one discussion Alan said that its not good to have hard coded data in services.

app/components/pipeline-graph-nav/template.hbs
```diff
@@ -78,7 +78,7 @@
         </div>
         <div class="col">
           <span class="title">Start Date</span><br>
-          <span>{{selectedEventObj.createTimeExact}}</span>
+          <span>{{startDate.content}}</span>
```
tkyi, line 81: just wondering, what is `content`? why can't we just use `startDate` directly?

banusha2: using `async` `await` in get function was not working as expected, page gets rendered first and data is computed after that. so we used `ObjectPromiseProxy` and it returns an object and our computed data is in `content` hence `startDate.content`

PR #824, opened 2022-09-16
fix: Collection pipeline name

banusha2:
## Context

To display full name instead of Truncating pipeline name.

Screen Shot 2022-09-16 at 1 26 32 PM

## Objective
Screen Shot 2022-09-16 at 1 18 07 PM

## References

## License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN set -x \
    && apk add --no-cache --update ca-certificates \
    && apk add --virtual .build-dependencies wget \
    && apk add nginx-mod-http-lua \
-   && wget -q -O - https://github.com/screwdriver-cd/ui/releases/latest \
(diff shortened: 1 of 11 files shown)


jithine:
in the objective screenshot, looks like alias name is still truncated.

banusha2:
> License

We just removed truncating for pipeline name. For aliasNames there was a requirement to truncate and it has tooltip so upon hover ,user can see full aliasName. Do we still want to remove truncating? I feel like for lengthy names it will not look good.

jithine:
> We just removed truncating for pipeline name. For aliasNames there was a requirement to truncate and it has tooltip so upon hover ,user can see full aliasName. Do we still want to remove truncating? I feel like for lengthy names it will not look good.
let alias name also wrap  instead of truncating.  Might be a Trevor or @adong  question

adong (approved):

PR #828, opened 2022-09-21
fix: Remove pipeline name as default aliasName

banusha2:
## Context

Once any pipeline's aliasName is set in a collection, that collection shows pipeline's aliasName for all pipeline, which isn't ideal.

## Objective

Display pipeline's aliasName ,if an aliasName is set .
Screen Shot 2022-09-21 at 11 29 21 AM

## References

## License

### DIFF
--- a/tests/integration/components/collection-view/component-test.js
+++ b/tests/integration/components/collection-view/component-test.js
@@ -153,12 +153,127 @@ const mockDefaultPipelines = copy([
   })
 ]);
 
+const mockDefaultPipelinesWithoutAlias = copy([
+  EmberObject.create({
+    id: 1,
+    scmUri: 'github.com:12345678:master',
+    createTime: '2017-01-05T00:55:46.775Z',
+    admins: {
+      username: true
+    },
+    workflow: ['main'],
+    scmRepo: {
+      name: 'screwdriver-cd/screwdriver',
+      branch: 'master',
+      url: 'https://github.com/screwdriver-cd/screwdriver/tree/master'
+    },
+    scmContext: 'github:github.com',
+    annotations: {},
+    lastEventId: 12,
+    lastBuilds: [
+      {
+        id: 123,
+        status: 'SUCCESS',
+        // Most recent build
+        createTime: '2017-09-05T04:02:20.890Z'
+      }
+    ],
+    metrics: EmberPromise.resolve(mockMetrics),
+    settings: {}
+  }),
+  EmberObject.create({
+    id: 2,
+    scmUri: 'github.com:87654321:master',
+    createTime: '2017-01-05T00:55:46.775Z',
+    admins: {
+      username: true
+    },
+    workflow: ['main', 'publish'],
+    scmRepo: {
+      name: 'screwdriver-cd/ui',
+      branch: 'master',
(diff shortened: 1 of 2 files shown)


jithine (dismissed):

adong (changes requested):
Can you add a test? I’m actually a little surprised that the test case didn’t catch alias/name changes.

banusha2:
> Can you add a test? I’m actually a little surprised that the test case didn’t catch alias/name changes.

I fixed the test cases in that PR as I remember.

adong:
> > Can you add a test? I’m actually a little surprised that the test case didn’t catch alias/name changes.
> 
> I fixed the test cases in that PR as I remember.

Hmm, I'm saying there should 2 tests that

1) has only pipeline name without alias
2) has both alias and name

So, later on, if either one behavior changes, we can catch it in the tests

tests/integration/components/collection-view/component-test.js
```diff
@@ -590,6 +716,53 @@ module('Integration | Component | collection view', function (hooks) {
...
+      .hasText('screwdriver-cd/models');
+    assert
+      .dom('.collection-pipeline:nth-of-type(2) .app-id')
+      .hasText('screwdriver-cd/screwdriver');
+    assert
+      .dom('.collection-pipeline:nth-of-type(3) .app-id')
+      .hasText('screwdriver-cd/ui');
+    assert
+      .dom('.collection-pipeline:nth-of-type(4) .app-id')
+      .hasText('screwdriver-cd/zzz');
```
adong, line 763:
```
assert.dom('.collection-pipeline:nth-of-type(4) .app-id p').hasText('zzz');
```
Is not sufficient, because `screwdriver-cd/zzz` also has `zzz`

I would say, add another check that ensures:
```
      .hasText('zzz');
```
And
```
      . doesNotIncludeText('screwdriver-cd/zzz');
```
see: https://github.com/simplabs/qunit-dom/blob/master/lib/__tests__/does-not-include-text.ts

adong (changes requested):

tests/integration/components/collection-view/component-test.js
adong: These comments should go to previous code block, https://github.com/screwdriver-cd/ui/pull/828/files/007392b2ee221f2c4b0a55e3dc91b611e9520808#diff-1d40fc3763f6a0d8c53bccb59972eb04fc0386972af290e245f57918576cb977R668

tests/integration/components/collection-view/component-test.js
```diff
@@ -541,6 +657,24 @@ module('Integration | Component | collection view', function (hooks) {
...
+      .hasText('screwdriver')
+      .doesNotIncludeText('screwdriver-cd/screwdriver');
+    assert
+      .dom('.collection-pipeline:nth-of-type(3) .app-id p')
+      .hasText('ui')
+      .doesNotIncludeText('screwdriver-cd/ui');
+    assert
+      .dom('.collection-pipeline:nth-of-type(4) .app-id p')
+      .hasText('zzz')
+      .doesNotIncludeText('screwdriver-cd/zzz');
```
adong, line 676: LGTM

adong (approved):

PR #829, opened 2022-09-21
fix: update userPreference for deleteRecord used in reset

adong:
## Context

Click on `reset` button once, will reset the user settings, click again, it will throw errors.

See: 
image

## Objective

This PR will fix the reset issue. The reason it throws errors is because the old reference was kept in memory. Prior Ember-Data 3.28, `deleteRecord + save()` will not `unloadRecord` from `peekRecord`, but `peekAll` will not have the stale data.

So, we need to manually call [`unloadRecord`](https://api.emberjs.com/ember-data/3.16/classes/Model/methods/unloadRecord?anchor=unloadRecord). 

## References

Later on in Ember-Data 3.28 (https://github.com/emberjs/data/pull/7258/) that `deleteRecord + save()` will NOT `unloadRecord`, but `destroyRecord` will `unloadRecord`, see: 
 https://github.com/emberjs/data/blob/master/packages/model/addon/-private/model.js#L735-L747

if you are interested in seeing the discussion on the difference between `deleteRecord` vs `destroyRecord` vs `unloadRecord` please check out discord thread: 
https://discord.com/channels/480462759797063690/486549196837486592/1022272715761123390
## License

### DIFF
--- a/app/components/pipeline-options/component.js
+++ b/app/components/pipeline-options/component.js
@@ -155,7 +155,7 @@ export default Component.extend({
       await this.shuttle.updatePipelineSettings(pipeline.id, {
         aliasName
       });
-      this.set('successMessage', 'Pipeline alias-name updated successfully');
+      this.set('successMessage', 'Pipeline alias updated successfully');
     } finally {
       this.set('aliasName', aliasName);
     }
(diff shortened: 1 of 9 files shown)


banusha2 (dismissed):
LGTM

jithine (approved):

PR #834, opened 2022-10-06
fix: Last run column 

banusha2:
## Context
Last run job column's data is not correct.

## Objective
Display correct data in Last run job column in collection.
card view:
Screen Shot 2022-10-27 at 1 19 35 PM
List view:
Screen Shot 2022-10-27 at 1 20 37 PM

## References

## License

### DIFF
--- a/app/components/collection-table-row/component.js
+++ b/app/components/collection-table-row/component.js
@@ -3,12 +3,6 @@ import { computed, getWithDefault } from '@ember/object';
 import { and } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { formatMetrics } from 'screwdriver-ui/utils/metric';
-import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
-import ObjectProxy from '@ember/object/proxy';
-import templateHelper from 'screwdriver-ui/utils/template';
-
-const ObjectPromiseProxy = ObjectProxy.extend(PromiseProxyMixin);
-const { getLastUpdatedTime } = templateHelper;
 
 export default Component.extend({
   store: service(),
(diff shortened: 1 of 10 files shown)


app/components/collection-table-row/component.js
```diff
@@ -30,13 +34,21 @@ export default Component.extend({
...
     return rootDir ? `${branch}#${rootDir}` : branch;
   }),
-  lastRun: computed('lastRun', function get() {
-    const { createTime } = this.pipeline;
-    const lastRun = getLastUpdatedTime({
-      createTime
-    });
+  lastRun: computed('pipeline', function get() {
+    return ObjectPromiseProxy.create({
+      promise: this.pipeline.get('metrics').then(metrics => {
```
jithine, line 39: does this cause extra API calls ?

jithine (commented):
don't we already have lastEventInfo, why is this needed

banusha2:
> don't we already have lastEventInfo, why is this needed

lastEventInfo don't have the createTime information. and afaik event and job are different things

app/components/collection-table-row/component.js
banusha2: No it just gets data from already fetched model. We have to treat it like a promise because Ui gets rendered before we compute the data and ui breaks

jithine:
@banusha2 last run event's data is good enough. Both should be more or less matching. Infact you brought an important point. Column should not be about last run job, it should be event. 
I propose to remove `job` from the column name, and just call it  `Last Run`

jithine (changes requested):
see my last comment, it should be last run event's info, not job's.

banusha2:
> see my last comment, it should be last run event's info, not job's.

Yes I have made changes please have look again

jithine (changes requested):
1. get rid of `last run` stuff in this PR and  rename `start date` to `last run`
2. same change in card view also
3. in card view have `last run` first before the `duration` in the row. you will understand once you see the view

app/pipeline/model.js
```diff
@@ -74,12 +75,14 @@ export default DS.Model.extend({
       return failedBuildCount;
     }
   }),
-  lastRunJob: computed('metrics.[]', {
+  lastRunEvent: computed('metrics.[]', {
```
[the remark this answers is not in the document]
banusha2: yes, but we need it in collection-view component for sorting feature.

jithine (approved):

PR #854, opened 2022-11-14
feat(2580): Disable actions that allow user to manually start an event/build

sagar1312:
## Context
Child pipelines are marked Inactive when the associated SCM URLs have been removed from the external configuration. No new events/builds should be allowed for such inactive child pipelines.

Backend API https://github.com/screwdriver-cd/screwdriver/pull/2724 now fails creation of new events for inactive pipelines.

## Objective
This PR disables all the action in the UI which submits a request to the Backend API to create a new event.
Here is the page level breakdown of UI changes...

### Job list view 

1.  Disable 'Start' and 'Restart' buttons/icons under 'ACTIONS' column against each job listed in the table
image

### Pull Requests card list view

1. Disable 'Start' button

image

### Pipeline Events view

1. Disable 'Start' button for pipeline with parameters
2. Hide 'Submit' button in the 'START PIPELINE WITH PARAMETERS' drop down menu
3. Hide 'Submit' button in the 'START PIPELINE WITH PARAMETERS' modal (when pipeline has large number of parameters)

image

image

## License

### DIFF
--- a/app/components/pipeline-list-view/component.js
+++ b/app/components/pipeline-list-view/component.js
@@ -5,6 +5,7 @@ import { inject as service } from '@ember/service';
 import { toCustomLocaleString } from 'screwdriver-ui/utils/time-range';
 import Table from 'ember-light-table';
 import isEqual from 'lodash.isequal';
+import { isActivePipeline } from 'screwdriver-ui/utils/pipeline';
 
 export default Component.extend({
   store: service(),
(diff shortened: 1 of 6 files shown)


tkyi:
Do we also make sure INACTIVE child pipeline event will not start through API call? or only through UI?

sagar1312:
> Do we also make sure INACTIVE child pipeline event will not start through API call? or only through UI?

As mentioned in the PR description, API changes are already in place to fail creation of new events for inactive pipelines.

jithine (approved):

PR #855, opened 2022-11-15
fix: Sorting on collection dashboard

banusha2:
## Context

sorting on status column.
## Objective

To sort status column based on priority sort.
## References

## License

### DIFF
--- a/app/components/collection-view/component.js
+++ b/app/components/collection-view/component.js
@@ -3,6 +3,14 @@ import { isEmpty, isEqual } from '@ember/utils';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import ENV from 'screwdriver-ui/config/environment';
+import {
+  sortByBranch,
+  sortByLastRunStatus,
+  sortByName,
+  sortByLastRun,
+  sortByHistory,
+  sortByDuration
+} from 'screwdriver-ui/utils/sort-functions';
 
 const viewOptions = [
   {
@@ -14,64 +22,6 @@ const viewOptions = [
     value: 'List'
   }
 ];
-
-const sortByLastRunStatus = (a, b) => {
-  const priorities = [
-    'success',
-    'running',
-    'queued',
-    'created',
-    'unstable',
-    'aborted',
-    'collapsed',
-    'frozen',
-    'failure',
-    'blocked'
-  ];
-  const aStatus = get(a, 'lastRunEvent.status');
-  const bStatus = get(b, 'lastRunEvent.status');
-
-  return priorities.indexOf(aStatus) - priorities.indexOf(bStatus);
-};
-
-const sortByName = (a, b) => {
-  const aName = get(a, 'scmRepo.name');
-  const bName = get(b, 'scmRepo.name');
-
-  if (aName > bName) {
-    return 1;
-  }
-  if (aName < bName) {
-    return -1;
-  }
-
-  return 0;
-};
-
-const sortByLastRun = (a, b) => {
-  const aCreateTime = get(a, 'lastRunEvent.createTime');
-  const bCreateTime = get(b, 'lastRunEvent.createTime');
-
-  return new Date(aCreateTime) - new Date(bCreateTime);
-};
-
-const sortByHistory = (a, b) => {
-  const aFailedBuildCount = get(a, 'failedBuildCount');
-  const bFailedBuildCount = get(b, 'failedBuildCount');
-
-  return aFailedBuildCount - bFailedBuildCount;
-};
-
-const sortByBranch = (a, b) => {
-  return a.scmRepo.branch - b.scmRepo.branch;
-};
-
-const sortByDuration = (a, b) => {
-  const aDuration = get(a, 'lastRunEvent.duration');
-  const bDuration = get(b, 'lastRunEvent.duration');
-
-  return aDuration - bDuration;
-};
 
 export default Component.extend({
   store: service(),
--- a/app/components/collection-view/component.js
+++ b/app/components/collection-view/component.js
@@ -15,7 +15,7 @@ const viewOptions = [
   }
 ];
 
-const sortByLastRunStatus = function (a, b) {
+const sortByLastRunStatus = (a, b) => {
   const priorities = [
     'success',
     'running',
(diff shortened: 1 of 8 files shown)


app/components/collection-view/component.js
```diff
@@ -221,6 +221,23 @@ export default Component.extend({
     organize() {
       this.set('isOrganizing', true);
     },
+    sortByStatus() {
+      const priorities = {
+        'ABORTED' : 6,
```
adong, line 226: This is an eye sore, can you reorder it in asc / desc order?

adong: Also what if the status passed in was not defined?

app/components/collection-view/component.js
```diff
@@ -221,6 +221,29 @@ export default Component.extend({
...
+      ];
+
+      const newSortedPipelines = this.sortedPipelines.sort((a, b) => {
+        const aStatus = get(a, 'lastRunEvent.status');
+        const bStatus = get(b, 'lastRunEvent.status');
+
+        return priorities.indexOf(aStatus) - priorities.indexOf(bStatus);
+      });
+
+      this.set('sortedPipelines', newSortedPipelines);
```
adong, line 245: Hmm, `sortedPipelines: sort('collectionPipelines', 'sortBy'),` see https://github.com/screwdriver-cd/ui/pull/855/files#diff-454c2200cdb1986fe26070e9113b98f9523bf68bdf5a35c3a96155a9b0549884R103


is a computed property, by doing so, it will break other sorting criteria, since this property is no longer a computed property, and will not react and update if dependent properties change

adong (changes requested):

adong (commented):
Change `sort` to `computed` to debug:  

Old
```
  sortedPipelines: sort('collectionPipelines', 'sortBy'),
```
New
```
  sortedPipelines: computed('collectionPipelines', 'sortBy', function() {
   // console what sortBy is, see if it expected
 }),
```

To debug what `sortBy` value is, and see why after `lastRunEvent.status:asc` and `lastRunEvent.status:desc` are set, and didn't work as expected.


PS: 
I have suspect something related to 

```
app/pipeline/model.js
79:  lastRunEvent: computed('metrics.[]', { 
```

If neither worked, we can sort the old way.

app/components/collection-view/component.js
```diff
@@ -100,7 +101,39 @@ export default Component.extend({
...
+      ];
+
+      sorted = this.collectionPipelines.toArray().sort((a,b) => {
+        const aStatus = get(a, 'lastRunEvent.status');
+        const bStatus = get(b, 'lastRunEvent.status');
+
+        return priorities.indexOf(aStatus) - priorities.indexOf(bStatus);
+      })
+    } else if(this.sortBy === 'scmRepo.name') {
+      sorted = this.collectionPipelines.toArray().sort((a,b) => {
```
adong, line 127: 2 comments. 

1) Extract these `sort(a, b)` functions to the top of the file, so it would be easier to modify and test individually without mocking the whole thing. 

2) Another comment on this, does it work on all column sorts? I see right now, only 2 columns are supported, i.e. `lastRun` and `scmRepo.name`

app/components/collection-view/component.js
```diff
@@ -100,7 +101,39 @@ export default Component.extend({
...
+        return a.scmRepo.name - b.scmRepo.name;
+      });
+    }
+    if (this.sortOrder === 'asc') {
+     return sorted;
```
adong, line 132: Minor comment: indention

app/components/collection-view/component.js
adong: Here's what I mean:

```
const sortByLastRunStatus = function(a, b) { // logics }
const sortByRepoName =  function(a, b) { // logics }
```

Then use as

```
if (this.sortBy === 'lastRun') {
  sorted = this.collectionPipelines.toArray().sort(sortByLastRunStatus)
}

```

banusha2: No, it doesn't work on all the columns.we have to fix all other column as well.

app/components/collection-view/component.js
```diff
@@ -100,7 +146,25 @@ export default Component.extend({
       return description;
     }
   }),
-  sortedPipelines: sort('collectionPipelines', 'sortBy'),
+  sortedPipelines: computed('collectionPipelines', 'sortBy', 'sortOrder', function sortedPipelines() {
+    let sorted;
+    if (this.sortBy === 'lastRunStatus') {
+      sorted = this.collectionPipelines.toArray().sort(sortByLastRunStatus);
+    } else if(this.sortBy === 'scmRepo.name') {
```
adong, line 153: Please refactor this `this.collectionPipelines.toArray()` to be a variable beforehand.

banusha2: Ok

app/components/collection-view/component.js
```diff
@@ -215,7 +300,10 @@ export default Component.extend({
           this.set('sortBy', [`${option}:desc`]);
           break;
         default:
-          this.set('sortBy', [option]);
+          this.setProperties({
```
adong, line 303: Question, do we need these
```
switch ()
 ...
```

Anymore? If not, please delete these lines. Just keep

```
          this.setProperties({
            sortBy: option,
            sortOrder: order
          });
```

app/components/collection-view/component.js
```diff
@@ -16,10 +15,69 @@ const viewOptions = [
   }
 ];
 
+const sortByLastRunStatus = (a, b) => {
```
adong, line 18: Please use `export const`, and add tests for these sorting functions

adong (changes requested):

tests/unit/utils/sort-functions-test.js
```diff
@@ -0,0 +1,276 @@
...
+
+const expectedPipelinesSortedByLastRun = [
+  '2020-10-06T17:57:53.388Z',
+  '2020-10-07T17:47:55.089Z',
+  '2020-10-08T18:07:55.089Z',
+  '2020-10-10T18:27:55.089Z'
+];
+
+module('Unit | Utility | sort functions', function () {
+  test('it sorts pipelines based upon names', function (assert) {
```
adong, line 233: Can you add to the description that the sorting is specific in either ascending or descending order, so it's easier to understand?

adong: Hmm, on second thought, lets leave as is for now

adong (approved):

PR #869, opened 2022-12-14
fix: sorting issues for collection page

adong:
## Context

1) Current sorting breaks on unknown `status` for the collection page
2) Break down the collection into `knownStatus` and `unknownStatus` pipelines, and latter remain at the bottom of the list when sorting.

3) Fix up/down arrow when another column was selected

## Objective

image

## License

### DIFF
--- a/app/components/collection-table-row/template.hbs
+++ b/app/components/collection-table-row/template.hbs
@@ -21,7 +21,11 @@
   <span title={{branch}}>{{branch}}</span>
 </td>
 <td class="last-run">
-  {{lastEventInfo.startTime}}
+  {{#if (eq lastEventInfo.createTime 0)}}
+    --/--/----
+  {{else}}
+    {{lastEventInfo.startTime}}
+  {{/if}}
 </td>
 <td class="duration">
   {{lastEventInfo.durationText}}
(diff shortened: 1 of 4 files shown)


app/components/collection-table-row/template.hbs
```diff
@@ -21,7 +21,11 @@
   <span title={{branch}}>{{branch}}</span>
 </td>
 <td class="last-run">
-  {{lastEventInfo.startTime}}
+  {{#if (eq lastEventInfo.createTime 0)}}
+    --/--/----
+  {{else}}
+    {{lastEventInfo.startTime}}
```
adong, line 27: I think the difference between `startTime` and `createTime` is createTime is time/date object, and startTime is just a string format of `mm/yy/yyyy`

image

app/utils/metric.js
```diff
@@ -114,7 +114,7 @@ const formatMetrics = metrics => {
     commitMessage: lastEvent.commit.message,
     commitUrl: lastEvent.commit.url,
     createTime: moment(lastEvent.createTime).format(),
-    duration: lastEvent.duration
+    duration: lastEvent.duration || 0
```
adong, line 117: `lastEvent.duration` could be `null`, so we need to fill it with 0. because `null` value breaks the `sort`ing

jithine (dismissed):

tkyi (approved):

jithine (approved):

jithine (approved):

PR #872, opened 2022-12-21
feat(2805): make label clickable if contains links

banusha2:
## Context
To set hyperlink to 'label' like 'commit' field.

## Objective
Make [label](https://docs.screwdriver.cd/user-guide/metadata#event-labels) more usable, so that users can put links inside `label` content, and instead of showing as text, it shows as hyperlinks

Screen Shot 2022-12-22 at 12 24 07 PM

## References

https://github.com/screwdriver-cd/screwdriver/issues/2805

## License


app/components/pipeline-graph-nav/component.js
```diff
@@ -53,5 +53,14 @@ export default Component.extend({
 
       return startDate;
     }
+  }),
+  label: computed('selectedEventObj.label', {
+    get() {
+      let label = this.get('selectedEventObj.label');
+
+      label = label.replace(/(https?\/\/[^\s+])/g, '<a href="$1">$1</a>');
```
adong, line 61: I would refactor this logic to a utility function called, does the text contain hyperlinks, `isTextLinkable`, and another function like `TransformTextToClickableContent`, something like that.

Also this logic only detects and replaces the first url, I think.

adong: Please refer this: 
https://www.cluemediator.com/find-urls-in-string-and-make-a-link-using-javascript

adong: 😀 It turns out frequent enough to as an interview question, and solution is in-place. https://www.pcoroneos.com/cracking-the-code-interview-1-3-urlify

app/components/pipeline-graph-nav/component.js
```diff
@@ -53,5 +54,18 @@ export default Component.extend({
 
       return startDate;
     }
+  }),
+  label: computed('selectedEventObj.label', {
+    get() {
+      let label = this.get('selectedEventObj.label');
+
+      const isTextLinkable = DoesTextContainsLink(label);
```
adong, line 62: `DoesTextContainsLink`,  `TransformTextToClickableContent` camel case please

Also, please refactor `isTextLinkable` to be a computed property and use in the template

app/components/pipeline-graph-nav/template.hbs
```diff
@@ -87,7 +87,7 @@
         {{#if selectedEventObj.label}}
           <div class="col">
             <span class="title">Label</span><br>
-            <span class="customize-label" title={{selectedEventObj.label}}>{{selectedEventObj.label}}</span>
```
adong, line 90: When `isTextLinkable` is a computed property of a boolean, 
Check if `isTextLinkable` true, then show `{{{label}}}`, otherwise, show `{{selectedEventObj.label}}`

app/utils/convert-url.js
```diff
@@ -0,0 +1,19 @@
+const URL_REGEX = /(((https?:\/\/)|(www\.))[^\s]+)/g;
+
+export function DoesTextContainsLink(text) {
+  return new RegExp(URL_REGEX).test(text);
```
adong, line 4: Can you rename this file to be `url-helper`, instead of convert url? `convert url` sounds like a function name and only convert url, rather than file name that contains all url related functions

app/utils/convert-url.js
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function TransformTextToClickableContent(text) {
```
adong, line 7: camel case

app/utils/convert-url.js
```diff
@@ -0,0 +1,19 @@
+const URL_REGEX = /(((https?:\/\/)|(www\.))[^\s]+)/g;
+
+export function DoesTextContainsLink(text) {
```
adong, line 3: camel case

app/utils/convert-url.js
```diff
@@ -0,0 +1,19 @@
...
+  return new RegExp(URL_REGEX).test(text);
+}
+
+export function TransformTextToClickableContent(text) {
+  if(!text) return;
+
+  return text.replace(URL_REGEX, function (url) {
```
adong, line 10: After I think more about it, I suspect this logic has a potential issue for cross-site scripting. For example, what would be rendered in `{{{label}}}` if `text` contains both valid hyperlinks and malicious scripts? 

i.e.

```
this is my build, http://screwdriver.cd/build/112233, here's my script <script>alert('XSS')</script>
```

Just want to point out, the current implementation _would_ run into issues.

app/components/pipeline-graph-nav/template.hbs
adong: `isTextLinkable` => `hasContainLinkableText`

app/utils/convert-url.js
adong: Hmm. Interestingly, github does something similar. 

image

image

adong: Please check out

>  [Usage of triple curly braces to allow raw HTML to be injected into the DOM is a large vector for exploits of your application (especially when the raw HTML is user-controllable). Instead of using `{{{foo}}}`, you should use appropriate helpers or computed properties that return a `SafeString` (via `Ember.String.htmlSafe` generally) and ensure that user-supplied data is properly escaped.](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-triple-curlies.md)

and [htmlSafe](https://api.emberjs.com/ember/release/functions/@ember%2Ftemplate/htmlSafe) helper

banusha2: yes done

app/utils/url-helper.js
```diff
@@ -0,0 +1,37 @@
...
+  return text.replace(URL_REGEX, function transformUrl(url) {
+    let hyperlink = url;
+
+    // eslint-disable-next-line no-useless-escape
+    if (!hyperlink.match('^https?:\/\/')) {
+      // eslint-disable-next-line prefer-template
+      hyperlink = 'http://' + hyperlink;
+    }
+
+    return (
```
adong, line 28: Instead of a string concatenation, how about just returning [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) ?

app/utils/url-helper.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @param {String} text to transform into clickable content
+ * @return {String} string with clickable content
+ */
+export function transformTextToClickableContent(text) {
+  if (!text) return;
+
+  // eslint-disable-next-line consistent-return
+  return text.replace(URL_REGEX, function transformUrl(url) {
```
adong, line 19: As I mentioned before, this function needs to filter out only to allow links to be processed and filter anything else. 

See:

image

And `https://cd.screwdriver.cd/pipelines/7290/events/718950?jobId=`
image

adong (changes requested):

app/utils/url-helper.js
banusha2: Sorry I didn't get it. what we have to filter out.

tests/unit/utils/url-helper-test.js
```diff
@@ -0,0 +1,25 @@
...
+import {
+  doesTextContainsLink,
+  transformTextToClickableContent
+} from 'screwdriver-ui/utils/url-helper';
+
+const string =
+  'Lorem ipsum dolor sit amet, https://abc.com consectetur adipiscing elit, sed do eiusmod tempor incididunt ut http://xyz.org labore et dolore magna aliqua.';
+
+const expectedTransformText =
+  'Lorem ipsum dolor sit amet, <a href="https://abc.com" target="_blank" rel="noopener noreferrer">https://abc.com</a> consectetur adipiscing elit, sed do eiusmod tempor incididunt ut <a href="http://xyz.org" target="_blank" rel="noopener noreferrer">http://xyz.org</a> labore et dolore magna aliqua.';
```
adong, line 11: Try adding some script tag and image tag to check the functionalities for `transformTextToClickableContent`. I dont think it would work.

adong (changes requested):
Hmm, I don't think the current implementation handles the XSS issue I brought up earlier. Can you implement a small XSS sanitizer file with an allow list to filter out the non-allowed tags?

In that case, the function `transformTextToClickableContent`'s transformation part can remain the same

app/utils/sanitize.js
```diff
@@ -0,0 +1,82 @@
...
+  blockElements = DEFAULT_OPTIONS.blockElements,
+  allowElements
+} = {}) {
+  const parser = new DOMParser();
+
+  /**
+   * @param {node}
+   * @return {}
+   */
+  function sanitiseNode(node) {
```
adong, line 26: 1) Rename the file from `sanitize` to `sanitizer`
2) Please update the word `sanitise` to `sanitize` to for consistency's sake.

app/utils/url-helper.js
```diff
@@ -0,0 +1,34 @@
+import { Sanitizer } from './sanitize';
+
+const URL_REGEX = /(((https?:\/\/)|(www\.))[^\s]+)/g;
+
```
adong, line 4: Please put `const sanitizer = new Sanitizer();` here, outside export, otherwise, this Sanitizer gets initialized everytime `transformTextToClickableContent` gets called.

app/utils/url-helper.js
```diff
@@ -0,0 +1,34 @@
+import { Sanitizer } from './sanitize';
+
+const URL_REGEX = /(((https?:\/\/)|(www\.))[^\s]+)/g;
```
adong, line 3: This regex is not working properly. It will transform `https://x.com,abcdef` as well, which shouldn't.

app/components/pipeline-graph-nav/component.js
```diff
@@ -53,5 +58,23 @@ export default Component.extend({
...
+      let label = this.get('selectedEventObj.label');
+
+      const isTextLinkable = doesTextContainsLink(label);
+
+      return isTextLinkable;
+    }
+  }),
+  label: computed('selectedEventObj.label', {
+    get() {
+      let label = this.get('selectedEventObj.label');
```
adong, line 73: Please rename this computed property from `label` to `linkableTextLabel`. 

`label` is confusing, because the code shows here is actually transformed, not raw label.

app/utils/url-helper.js
```diff
@@ -0,0 +1,34 @@
...
+
+  const sanitizer = new Sanitizer();
+  const filteredString = sanitizer.parseFromString(text);
+
+  // eslint-disable-next-line consistent-return
+  return filteredString.replace(URL_REGEX, function transformUrl(url) {
+    let hyperlink = url;
+
+    // eslint-disable-next-line no-useless-escape
+    if (!hyperlink.match('^https?://')) {
```
adong, line 28: I don't think this code does what is intended to do. 

This code can be simplified as it assumes the `url` already has `http` or `https`, or `www`, this if statement only runs for url with `www` and prefix with `http://`. 


By the way, why sanitizer runs before regex replacement? I don't get the order of the logic here. We should be filtering at the end to make sure the final code aligns with allow/blocked tags.

adong (changes requested):
Not working properly

tests/unit/utils/url-helper-test.js
```diff
@@ -0,0 +1,25 @@
...
+  transformTextToClickableContent
+} from 'screwdriver-ui/utils/url-helper';
+
+const string =
+  'Lorem ipsum dolor sit amet, https://abc.com consectetur adipiscing elit, sed do <img src="lorem" />eiusmod tempor incididunt ut <script>document.getElementById("lorem").innerHTML = "lorem epsum!";</script> http://xyz.org labore et dolore magna aliqua.';
+
+const expectedTransformText =
+  'Lorem ipsum dolor sit amet, <a href="https://abc.com" target="_blank" rel="noopener noreferrer">https://abc.com</a> consectetur adipiscing elit, sed do eiusmod tempor incididunt ut  <a href="http://xyz.org" target="_blank" rel="noopener noreferrer">http://xyz.org</a> labore et dolore magna aliqua.';
+
+module('Unit | Utility | url-helper', function () {
```
adong, line 13: This is a good one _overall_ test case. Can you add more test cases to handle each scenario?

app/utils/url-helper.js
banusha2: because first sanitizer removes all the extra tags then which user has put link then that link is converted into a `<a>` tag.

tests/unit/utils/url-helper-test.js
banusha2: sure

app/utils/url-helper.js
banusha2: Now sanitization is done at last. 

i didn't you get on second statement here.

adong: Let me add more explanations. 

I'm saying the code block of 

```
if (!hyperlink.match('^https?://')) {  
...

```

Only converts url string starts with `www`, and will not work for other links like `yahoo.com`, will not work for `github.io`, or even `screwdriver.cd`. 

Coming up with our own regex is very unlikely to cover all possible urls. 

I need to think more about it, and there has to be a better way.

adong:
Actually, this PR became overly complicated for issue https://github.com/screwdriver-cd/screwdriver/issues/2805, and does not handle

Check out this library, https://github.com/johno/ember-linkify, which uses https://github.com/Hypercontext/linkifyjs under the hood, in my opinion, more industry-standard,  which handles 1) link detection and transformation and 2) the XSS Sanitization.  See [demo](https://linkify.js.org/#demo)

Try: 

`ember install ember-linkify`

Should be much easier 

`app/components/pipeline-graph-nav/template.hbs`

```
<span class="customize-label" title={{selectedEventObj.label}}>{{linkify selectedEventObj.label urlLength=30}}</span>
```

Thus, I'm closing this PR, please raise a new PR.

PR #873, opened 2023-01-03
feat(2669): Render Stages in the workflow graph if configured

sagar1312:
## Context

Stage represents a group of closely related jobs acting on a specific environment in the pipeline. We need to visualize these groupings in the workflow graph for an event.
Stage will have a name and description (optional).

## Objective

Render stages in the workflow graph

**Examples:**
https://github.com/sagar1312/screwdriver-config/blob/main/stage-with-linear-jobs/screwdriver.yaml
image

https://github.com/sagar1312/screwdriver-config/blob/main/stage-with-parallel-jobs/screwdriver.yaml
image

https://github.com/sagar1312/screwdriver-config/blob/main/stage-with-linear-jobs-not-in-first-row/screwdriver.yaml
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/2669

## License

### DIFF
--- a/app/components/workflow-graph-d3/component.js
+++ b/app/components/workflow-graph-d3/component.js
@@ -8,6 +8,7 @@ import {
   subgraphFilter,
   removeBranch
 } from 'screwdriver-ui/utils/graph-tools';
+import groupBy from 'lodash.groupby';
 
 export default Component.extend({
   shuttle: service(),
@@ -18,6 +19,13 @@ export default Component.extend({
   displayJobNames: true,
   showPRJobs: true,
   graph: { nodes: [], edges: [] },
+  showStages: computed('minified', 'stages', {
+    get() {
+      const stages = this.stages === undefined ? [] : this.stages;
+
+      return !(this.minified || stages.length === 0);
+    }
+  }),
   decoratedGraph: computed(
     'builds.@each.{id,status}',
     'completeWorkflowGraph',
@@ -31,6 +39,7 @@ export default Component.extend({
     'startFrom',
     'workflowGraph',
     'prChainEnabled',
+    'stages',
     {
       get() {
         const showDownstreamTriggers =
@@ -41,6 +50,7 @@ export default Component.extend({
 
         const { startFrom } = this;
 
+        const stages = this.stages === undefined ? [] : this.stages;
         const prJobs = this.prJobs === undefined ? [] : this.prJobs;
         const jobs = (this.jobs === undefined ? [] : this.jobs).concat(prJobs);
 
@@ -110,7 +120,8 @@ export default Component.extend({
           jobs,
           start: startFrom,
           chainPR: this.prChainEnabled,
-          prNum: this.selectedEventObj?.prNum
+          prNum: this.selectedEventObj?.prNum,
+          stages
         });
       }
     }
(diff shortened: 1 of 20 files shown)


adong:
I tried a sample PR: 
https://cd.screwdriver.cd/pipelines/9762/events/704140

The `Y_DISPLAYMENT` on row1 and row2 are good, and on row3 is not.

image

pritamstyz4ever:
The UI doesn't look as per the mocks, the jobs inside the stage should be vertically aligned we can set the arrows from top to bottom. I think that would be more intuitive for users rather than having horizontal alignment and forcing users to scroll right which is not a good experience.

pritamstyz4ever (changes requested):
Please work on the UI design as per the mocks

sagar1312:
> I tried a sample PR: https://cd.screwdriver.cd/pipelines/9762/events/704140
> 
> The `Y_DISPLAYMENT` on row1 and row2 are good, and on row3 is not.
> 
> image

@adong 
This has been fixed by the last commit

image

sagar1312:
Latest changes also dynamically compute and set stage position and height based on stage name and description

sagar1312:
> Please work on the UI design as per the mocks

@pritamstyz4ever 
The pipeline/job example in the mock is different. It is not mandatory that stages job/nodes needs to be positioned one below the other

app/pipeline/events/template.hbs
```diff
@@ -61,6 +61,7 @@
       completeWorkflowGraph=completeWorkflowGraph
       workflowGraph=pipeline.workflowGraph
       jobs=jobs
+      stages=stages
```
adong, line 64: I would recommend to keep a copy of the original `workflow-graph-d3` component, and duplicate and refactor the new changes in a new template, i.e. `workflow-graph-stage-d3`. 

So we can move forward, without worrying about: 
1) existing workflow 
2) messing something corner case for the main when soft launching 
3) more flexible to add/remove codes 


```
{{#if stages}}
 {{workflow-graph-stage-d3}}
{{else}}
 {{workflow-graph-d3}}
{{/if}}
```

adong (changes requested):
See my comment in https://github.com/screwdriver-cd/ui/pull/873/files#r1084458729

app/pipeline/events/template.hbs
sagar1312: Majority of the code (to render stage elements in the workflow graph) with in `workflow-graph-d3` component is already isolated and executed only if stages exist.
https://github.com/screwdriver-cd/ui/pull/873/files#diff-3a03bfd3e46c3280d50d0738c305e3fe169c2729a63e2f0af768070f7ffbbcf9R270

Only [common code](https://github.com/screwdriver-cd/ui/pull/873/files#diff-3a03bfd3e46c3280d50d0738c305e3fe169c2729a63e2f0af768070f7ffbbcf9R219) is applying vertical displacement for existing graph elements like nodes and edges. This is no op when stages are not defined as it always returns `0`.
```
const getVerticalDisplacementByRowPosition = pos => {
      return verticalDisplacementByRowPosition[pos] || 0;
    };
```

I do not see a compelling reason to duplicate 350+ lines of code (plus redundant tests) when there is no complex interleaving logic.

adong (dismissed):

app/components/workflow-graph-d3/component.js
```diff
@@ -211,6 +222,18 @@ export default Component.extend({
 
     const { ICON_SIZE, TITLE_SIZE, ARROWHEAD } = this.elementSizes;
 
+    // Need this if stage is in first row
+    const Y_DISPLACEMENT = this.showStages ? ICON_SIZE / 2 : 0;
+    // gap between stages
+    const STAGE_GAP = ICON_SIZE / 9;
+
+    const stagesGroupedByRowPosition = groupBy(data.stages, 'pos.y');
```
adong, line 230: Minor: can you add more comments to explain the calculation?

adong (approved):

pritamstyz4ever (approved):
Blanket approving now, I will review later

PR #874, opened 2023-01-04
fix: update pipelines from different tabs

banusha2:
## Context

Fix: adding pipelines to collections from different browser tabs is overwriting collection data.

https://user-images.githubusercontent.com/15989893/213005924-c9bdac9d-b905-4d55-b2d3-df849e1b6615.mp4

## Objective

Using new API call, instead of UI asks the backend to store what the pipeline to a given collection, UI will just pass the necessary pipeline ids to either delete or add, to avoid override the data in the backend. 

## References

Collection https://github.com/screwdriver-cd/screwdriver/issues/2765

Backend PR: https://github.com/screwdriver-cd/screwdriver/pull/2797

## License

### DIFF
--- a/app/dashboard/show/controller.js
+++ b/app/dashboard/show/controller.js
@@ -1,51 +1,25 @@
 import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
-import { getWithDefault, set } from '@ember/object';
 import Controller from '@ember/controller';
 
 export default Controller.extend({
   collection: alias('model.collection'),
   shuttle: service(),
   actions: {
-    async removePipeline(pipelineId) {
-      const collectionId = this.get('collection.id');
-      try {
-        await this.shuttle.removePipeline(collectionId, pipelineId);
-        const collection = await this.store.findRecord('collection', collectionId);
-
-        return collection;
-      }
-      catch (e) {
-
-       return e;
-      }
+    removePipeline(pipelineId, collectionId) {
+      return this.shuttle.removePipeline(collectionId, pipelineId);
     },
-    async removeMultiplePipelines(removedPipelineIds) {
-      const collectionId = this.get('collection.id');
-
-      try {
-        await this.shuttle.removeMultiplePipelines(collectionId, removedPipelineIds);
-        const collection = await this.store.findRecord('collection', collectionId);
-
-        return collection;
-      }
-      catch (e) {
-        return e;
-      }
+    removeMultiplePipelines(removedPipelineIds, collectionId) {
+      return this.shuttle.removeMultiplePipelines(
+        collectionId,
+        removedPipelineIds
+      );
     },
     onDeleteCollection() {
       this.transitionToRoute('home');
     },
-    async addMultipleToCollection(addedPipelineIds, collectionId) {
-      try {
-        await this.shuttle.updateCollection(collectionId, addedPipelineIds);
-        const collection = await this.store.findRecord('collection', collectionId);
-
-        return collection;
-      }
-      catch (e) {
-        return e;
-      }
+    addMultipleToCollection(addedPipelineIds, collectionId) {
+      return this.shuttle.updateCollection(collectionId, addedPipelineIds);
     }
   }
 });
--- a/app/dashboard/show/controller.js
+++ b/app/dashboard/show/controller.js
@@ -13,10 +13,11 @@ export default Controller.extend({
         await this.shuttle.removePipeline(collectionId, pipelineId);
         const collection = await this.store.findRecord('collection', collectionId);
 
-        return collection.save();
+        return collection;
       }
       catch (e) {
-        return e;
+
+       return e;
       }
     },
     async removeMultiplePipelines(removedPipelineIds) {
@@ -26,7 +27,7 @@ export default Controller.extend({
         await this.shuttle.removeMultiplePipelines(collectionId, removedPipelineIds);
         const collection = await this.store.findRecord('collection', collectionId);
 
-        return collection.save();
+        return collection;
       }
       catch (e) {
         return e;
@@ -40,7 +41,7 @@ export default Controller.extend({
         await this.shuttle.updateCollection(collectionId, addedPipelineIds);
         const collection = await this.store.findRecord('collection', collectionId);
 
-        return collection.save();
+        return collection;
       }
       catch (e) {
         return e;
--- a/app/dashboard/show/controller.js
+++ b/app/dashboard/show/controller.js
@@ -7,50 +7,44 @@ export default Controller.extend({
   collection: alias('model.collection'),
   shuttle: service(),
   actions: {
-    removePipeline(pipelineId) {
+    async removePipeline(pipelineId) {
       const collectionId = this.get('collection.id');
+      try {
+        await this.shuttle.removePipeline(collectionId, pipelineId);
+        const collection = await this.store.findRecord('collection', collectionId);
 
-      return this.store
-        .findRecord('collection', collectionId)
-        .then(collection => {
-          const pipelineIds = getWithDefault(collection, 'pipelineIds', []);
-
-          set(
-            collection,
-            'pipelineIds',
-            pipelineIds.filter(id => id !== pipelineId)
-          );
-
-          this.shuttle.removePipeline(collectionId, pipelineId);
-        });
+        return collection.save();
+      }
+      catch (e) {
+        return e;
+      }
     },
-    removeMultiplePipelines(removedPipelineIds) {
+    async removeMultiplePipelines(removedPipelineIds) {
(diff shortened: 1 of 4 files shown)


adong (changes requested):
1) Semantic error, these shuttle APIs are async. So await is needed, with try catch logics to guard. 

```
    try {
      const x = await y();
      // logics
    } catch (e) {
      // stuff
    } 
```

2) How does using the new API endpoints solves the multi-tab issues? 

_(For context purposes, Can you add 1) the issue link, 2) and the related PR backend change if any to the current PR description? )_

Say, one has 2 pages to edit the same collection.

1. Deleting pipelines from collection on page 1
2. Check page 2, not reflecting the changes, and one has to refresh the page to see it

The reason one sees such behavior is that: 
Each webpage has its own data stored in memory. `localStorage` and `sessionStorage` are shared across tabs within the same browser.

Say, we could use `localStorage` to handle it. Then what about the case that page 1 is now in Chrome, and page 2 is now in Firefox. `localStorage` will not share across different apps. 


__HINT__: 
I think, just refactoring the API endpoints from `collection.save()` method to `shuttle`'s api service, will NOT solve the problem underneath. 

The issue to me still inside the part 1) that how one handles the `success` and `error` case

adong:
1) fetch the latest pipelines before updating anything. (logics already implemented) 
```
      return this.store
        .findRecord('collection', collectionId)
        .then(collection => {
```

2) fetch alone is not enough, also need to _store_ the fetched data, such as pipelines, so if there were newly added or deleted pipelines, the UI will reflect that (_I think_ this is the part that current PR stucks on)

3) Handle the **success** and **error** cases, what if the pipeline ids that are being deleted have already been deleted in the backend, or, what about such a collection itself already being deleted?

app/shuttle/service.js
```diff
@@ -348,6 +348,29 @@ export default Service.extend({
 
     const url = `/pipelines?${query}`;
 
+    return this.fetchFromApi(method, url);
+  },
+
+  async updateCollection(collectionId, pipelineIds) {
+    const method = 'put';
+    const query = decodeURIComponent($.param({ ids: pipelineIds }));
```
adong, line 356: Why do you use `decodeURIComponent` and `$.param`?

I dont think these functions are needed

app/shuttle/service.js
```diff
@@ -348,6 +348,29 @@ export default Service.extend({
...
+  async removePipeline(collectionId, pipelineId) {
+    const method = 'delete';
+    const url = `/collections/${collectionId}/pipelines?ids[]=${pipelineId}`;
+
+    return this.fetchFromApi(method, url);
+  },
+
+  async removeMultiplePipelines(collectionId, pipelineIds) {
+    const method = 'delete';
+    const query = decodeURIComponent($.param({ ids: pipelineIds }));
```
adong, line 371: Same here, `decodeURIComponent` and `$.param` are not needed

app/dashboard/show/controller.js
```diff
@@ -46,13 +48,8 @@ export default Controller.extend({
...
         .findRecord('collection', collectionId)
         .then(collection => {
-          const pipelineIds = collection.get('pipelineIds');
 
-          collection.set('pipelineIds', [
-            ...new Set([...pipelineIds, ...addedPipelineIds])
-          ]);
-
-          return collection.save();
+          this.shuttle.updateCollection(collectionId, addedPipelineIds);
```
adong, line 52: FYI, `shuttle.updateCollection` is an `async` function, so needs to add `return` in front of it when inside `then` for using promise syntax. 

Also want to point out that using shuttle api to call the backend, will result in the `collection.pipelineIds` being out of sync because `collection.save` were not used.

app/dashboard/show/controller.js
```diff
@@ -19,7 +21,7 @@ export default Controller.extend({
             pipelineIds.filter(id => id !== pipelineId)
           );
 
-          return collection.save();
+          this.shuttle.removePipeline(collectionId, pipelineId);
```
adong, line 24: How about moving `this.shuttle.removePipeline(collectionId, pipelineId);` to right before the `return this.store`? 


It would make the logic easier: UPDATE data, and GET data.

app/dashboard/show/controller.js
```diff
@@ -36,7 +38,7 @@ export default Controller.extend({
             pipelineIds.filter(id => !removedPipelineIds.includes(id))
           );
 
-          return collection.save();
+          this.shuttle.removeMultiplePipelines(collectionId, removedPipelineIds);
```
adong, line 41: How about moving `this.shuttle.removePipeline(collectionId, pipelineId);` to right before the `return this.store`? 


It would make the logic easier: UPDATE data, and GET data.

adong: For sequential order purpose, either change to chained promises or use async/await. I prefer the latter approaches with try catch.

app/dashboard/show/controller.js
adong: Something like:
```
try {
  await this.shuttle.removePipeline(collectionId, pipelineId);
  await this.store.findRecord('collection', collectionId)
} catch 
```

app/shuttle/service.js
banusha2: Removed `decodeURIComponent` , but `$.param` is needed I think to make a query params like this `/collections/123/pipelines?ids[]=11&ids[]=23` from an array

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
-            pipelineIds.filter(id => id !== pipelineId)
-          );
-
-          return collection.save();
-        });
+      try {
+        await this.shuttle.removePipeline(collectionId, pipelineId);
+        const collection = await this.store.findRecord('collection', collectionId);
+
+        return collection.save();
```
adong, line 16: I think `return collection.save();` is not necessary. Since `collection` itself was not changed after fetch.

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
 
-          set(
-            collection,
-            'pipelineIds',
-            pipelineIds.filter(id => !removedPipelineIds.includes(id))
-          );
-
-          return collection.save();
-        });
+        return collection.save();
```
adong, line 29: I think `return collection.save();` is not necessary. Since `collection` itself was not changed after fetch.

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
-          ]);
-
-          return collection.save();
-        });
+    async addMultipleToCollection(addedPipelineIds, collectionId) {
+      try {
+        await this.shuttle.updateCollection(collectionId, addedPipelineIds);
+        const collection = await this.store.findRecord('collection', collectionId);
+
+        return collection.save();
```
adong, line 43: I think `return collection.save();` is not necessary. Since `collection` itself was not changed after fetch.

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
+      }
+      catch (e) {
+        return e;
```
adong, line 46: Also here in the catch, I think there can be a case depending on the returned response, that says: "pipeline x is has been already added to collection y", or something like that. I'm not sure how the backend response to duplicate, and you can explore more.

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
-            'pipelineIds',
-            pipelineIds.filter(id => !removedPipelineIds.includes(id))
-          );
-
-          return collection.save();
-        });
+        return collection.save();
+      }
+      catch (e) {
+        return e;
```
adong, line 32: Similar here, like "pipeline x, y, z already removed from collection y"

app/dashboard/show/controller.js
```diff
@@ -1,59 +1,50 @@
...
-          return collection.save();
-        });
+      try {
+        await this.shuttle.removePipeline(collectionId, pipelineId);
+        const collection = await this.store.findRecord('collection', collectionId);
+
+        return collection.save();
+      }
+      catch (e) {
+        return e;
```
adong, line 19: Similar here, like "pipeline x already removed from collection y" in the catch case.

app/dashboard/show/controller.js
banusha2: Sorry, didnt get you here

app/dashboard/show/controller.js
banusha2: Backend does not return an error in case of duplications, just checked it still returns `204`

adong:
PR https://github.com/screwdriver-cd/ui/pull/875/files merged, and see PR files for more details

adong (approved):

PR #880, opened 2023-02-16
feat: Add template details page

banusha2:
## Context

If users use a template inside their sd.yaml for jobs, the build details page should show something similar as we do for the container images.

## Objective
Screen Shot 2023-02-16 at 11 35 32 AM 1

## References
https://github.com/screwdriver-cd/screwdriver/issues/2822

## License

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -68,7 +68,7 @@ export default Component.extend({
     }
   }),
 
-  getTemplate: computed('build', {
+  getTemplate: computed('templateId', {
     get() {
       const templateId = this.get('templateId');
 
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -68,18 +68,18 @@ export default Component.extend({
     }
   }),
 
-  getTemplate: computed('templateId', {
+  template: computed('templateId', {
     get() {
       const templateId = this.get('templateId');
 
       return ObjectPromiseProxy.create({
-        promise: this.shuttle.getTemlateDetails(templateId).then(template => {
+        promise: this.shuttle.getTemplateDetails(templateId).then(template => {
           const { namespace, name, version } = template;
 
           return {
-            link: `/templates/${namespace}/${name}/${version}`,
             name,
-            version
+            version,
+            namespace
           };
         })
       });
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -3,7 +3,11 @@ import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { isActiveBuild, isPRJob } from 'screwdriver-ui/utils/build';
+import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
+import ObjectProxy from '@ember/object/proxy';
 import { getTimestamp } from '../../utils/timestamp-format';
+
+const ObjectPromiseProxy = ObjectProxy.extend(PromiseProxyMixin);
 
 export default Component.extend({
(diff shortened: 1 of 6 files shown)


app/components/build-banner/template.hbs
```diff
@@ -78,11 +78,18 @@
     <span class="banner-value">{{buildContainer}}</span>
     <span class="banner-label">Container</span>
   </li>
+  {{#if templateId}}
+    <li class="template-info">
```
adong, line 82: Also need to check if `getTemplate.content` has value or not. Otherwise, it will throw errors.

app/shuttle/service.js
```diff
@@ -381,5 +381,23 @@ export default Service.extend({
...
+  async getTemlateDetails(templateId) {
+    const method = 'get';
+    const url = `/template/${templateId}`;
+
+    try {
+      const temlateDetails = await this.fetchFromApi(method, url);
+
+      return temlateDetails;
+    } catch (e) {
+      return {};
```
adong, line 400: Non-valid value such as `{}` needs to be checked in `template.hbs`

adong (changes requested):
Needs to add tests

app/components/build-banner/template.hbs
```diff
@@ -78,11 +78,18 @@
...
+      <span class="banner-value"><a href={{getTemplate.content.link}}>{{getTemplate.content.name}}:{{getTemplate.content.version}}</a></span>
```
adong, line 83: Use `<a>` link redirection is not good, please use ember `{{link-to}}`, you can search our codebase for example

app/components/build-banner/component.js
```diff
@@ -63,6 +68,24 @@ export default Component.extend({
     }
   }),
 
+  getTemplate: computed('templateId', {
```
adong, line 71: Please rename `getTemplate` to `template` if not already used. We don't need to use verb in the variable, usually verb is used for actions.

app/components/build-banner/component.js
```diff
@@ -63,6 +68,24 @@ export default Component.extend({
...
+    get() {
+      const templateId = this.get('templateId');
+
+      return ObjectPromiseProxy.create({
+        promise: this.shuttle.getTemlateDetails(templateId).then(template => {
```
adong, line 76: like `shuttle.getTemlateDetails` is fine to use `get`

app/shuttle/service.js
```diff
@@ -381,5 +381,23 @@ export default Service.extend({
...
+   * @return {Promise} temlate details
+   */
+  async getTemlateDetails(templateId) {
+    const method = 'get';
+    const url = `/template/${templateId}`;
+
+    try {
+      const temlateDetails = await this.fetchFromApi(method, url);
+
+      return temlateDetails;
```
tkyi, line 398:
```suggestion
   * getTemplateDetails
   * @param {Number} templateId template id
   * @return {Promise} template details
   */
  async getTemplateDetails(templateId) {
    const method = 'get';
    const url = `/template/${templateId}`;

    try {
      const templateDetails = await this.fetchFromApi(method, url);

      return templateDetails;
```

app/components/build-banner/component.js
tkyi, line 76:
```suggestion
        promise: this.shuttle.getTemplateDetails(templateId).then(template => {
```

tkyi (changes requested):
tests?

app/components/build-banner/component.js
banusha2: removing`this` gives error that shuttle is not define

adong (approved):

PR #885, opened 2023-02-24
fix: template label and link

banusha2:
## Context

To fix template label and link.
## Objective
Screen Shot 2023-02-24 at 2 35 04 PM

## License

### DIFF
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -83,20 +83,19 @@ export default Component.extend({
 
       const templateProxy = ObjectPromiseProxy.create({
         promise: resolve(this.shuttle.getTemplateDetails(templateId))
-      })
+      });
 
-      
       templateProxy.then(template => {
-       const usageTemplate = {
-          'name': template.name,
-          'namsespace': template.namespace,
-          'version': template.version
-        }
+        const usageTemplate = {
+          name: template.name,
+          namespace: template.namespace,
+          version: template.version
+        };
 
-        this.set('template', usageTemplate)
-      })
+        return this.set('template', usageTemplate);
+      });
     },
-    set(_, value ) {
+    set(_, value) {
       return set(this, '_template', value);
     }
   }),
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -92,7 +92,9 @@ export default Component.extend({
           version: template.version
         };
 
-        return this.set('template', usageTemplate);
+        this.set('template', usageTemplate);
+
+        return this.template;
       });
     },
     set(_, value) {
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -1,7 +1,6 @@
 import { set, computed } from '@ember/object';
 import { alias } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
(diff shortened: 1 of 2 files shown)


app/components/build-banner/component.js
```diff
@@ -83,20 +83,19 @@ export default Component.extend({
...
-        }
+        const usageTemplate = {
+          name: template.name,
+          namespace: template.namespace,
+          version: template.version
+        };
 
-        this.set('template', usageTemplate)
-      })
+        return this.set('template', usageTemplate);
```
jithine, line 95: is this `set` needed here ?

adong: I dont think this syntax is correct.

How about? 

```
this.set('template', usageTemplate);

return this.template;
```

app/components/build-banner/component.js
```diff
@@ -83,20 +83,21 @@ export default Component.extend({
...
+
+        this.set('template', usageTemplate);
 
-        this.set('template', usageTemplate)
-      })
+        return this.template;
+      });
     },
-    set(_, value ) {
+    set(_, value) {
```
adong, line 100: I don't think `set` method is needed.

adong: Also, the variable is not the same, one is setting `_template` and one is setting `template`. So, I would say just delete `set()` method

app/components/build-banner/component.js
```diff
@@ -83,20 +83,21 @@ export default Component.extend({
...
-          'namsespace': template.namespace,
-          'version': template.version
-        }
+        const usageTemplate = {
+          name: template.name,
+          namespace: template.namespace,
+          version: template.version
+        };
+
+        this.set('template', usageTemplate);
```
adong, line 95: Variable `template` is a computed property, `template: computed(...`

and set on the same variable name inside its own function does not really do anything. 


So this whole thing can be simplified as 

```js
  template: computed('templateId', {
    get() {
      const templateId = this.get('templateId');

      const templateProxy = ObjectPromiseProxy.create({
        promise: resolve(
          this.shuttle.getTemplateDetails(templateId)
            .then(template => {
              const usageTemplate = {
                name: template.name,
                namespace: template.namespace,
                version: template.version
              };

              return usageTemplate;
            })
          )
      });
    }
  })
```

adong: The issue here actually is that this function itself has no return value, i.e. `templateProxy` was not returned. 

It ought to be something like this:

```
  template: computed('templateId', {
    get() {
      const templateId = this.get('templateId');

      return ObjectPromiseProxy.create({
        promise: this.shuttle.getTemplateDetails(templateId)
          .then(template => {
            const usageTemplate = {
              name: template.name,
              namespace: template.namespace,
              version: template.version
            };

            return usageTemplate;
          })
      });
    }
  }),
```

adong:
Just need to 
1) Revert the changes for `app/components/build-banner/template.hbs`, since https://github.com/screwdriver-cd/ui/pull/886 is merged
2) Update file `app/components/build-banner/component.js` per comment https://github.com/screwdriver-cd/ui/pull/885#discussion_r1120660966

adong (approved):

PR #868, opened 2022-12-14
fix: revamp user setting ui

banusha2:
## Context

Collection Dashboard required changes.

## Objective
UI changes on User settings.

Screen Shot 2022-12-13 at 2 32 56 PM
">

## References
https://github.com/screwdriver-cd/screwdriver/issues/2791

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #844, opened 2022-10-20
BREAKING CHANGES: major bump upgrade ember@4.4.3

yk634:
## Context

Update to LTS version (4.4) as the version of Ember framework used is out of date.

## Objective

Modify code, tests to adapt to newer versions of Ember.

- Library Changes
  - ember-cli-jstree -> [ember-basic-tree](https://github.com/dwickern/ember-basic-tree)
    - Because ember-cli-jstree does not support Ember@4.
- High-impact updates to the library
  - `ember-bootstrap: "^5.0.0"`
    - bootstrap@3 deprecated in Ember@4.
- Newly started using library
  - [ember-svg-jar](https://github.com/voltidev/ember-svg-jar)
    - Used to display some icons.

We have confirmed that each function of Screwdriver.cd is available from the UI, except for the following
- UI does not follow when referring to Arifact from URL in Artifact view page.

## References

screwdriver-cd/screwdriver#2608

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #891, opened 2023-03-02
fix(2838): Problems caused by ember upgrade

yk634:
## Context

Fix some display issues caused by Ember upgrade in https://github.com/screwdriver-cd/ui/pull/844.

## Objective

- Overall
  - Font looks a bit large.
    - [In bootstrap4, the default font size is `16px`](https://getbootstrap.com/docs/4.6/content/typography/#global-settings), so set `14px` in the root as before.
  - Display position of the part using `pull-{right,left}`
    - [`pull-{right,left}` is obsolete in bootstrap4](https://getbootstrap.com/docs/4.1/migration/#utilities).
    - e.g. before:
image
after:
image

- Templates/Commands list page
  - There are more than that, but only 10 are shown.
    - Since default page size is 10, set `pageSize` to the number that should be displayed.
  - Cannot sort by `Name` and `Namespace`.
    - In `ember-models-table`, If `propertyName` isn't provided, sorting and filtering options for current column are ignored.
- Collections page
  - The display position of the `Name` value is biased upward on the table.
- Build details page
  - Steps and Artifacts tabs are small
before:
image
after:
image

- Pipeline secrets page
  - Display corruption on access tokens section.
    - To be precise, this is not due to Ember Upgrade, but to https://github.com/screwdriver-cd/ui/pull/868.
    - Fix the Access Token section of the Pipeline Secrets page so that the User Settings page is not affected.
      - padding must function properly.
      - No unnatural borders should appear.
        - Since the border is [defined as a user-preference](https://github.com/screwdriver-cd/ui/blob/e4faaa9b8b399c77fc9011afbe322dc19219defa/app/styles/user-settings.scss#L11), removing it from the component will not affect the User Setting page.

before:
image
after:
image

## References

- [issue](https://github.com/screwdriver-cd/screwdriver/issues/2838)
- [ember-models-table docs](https://onechiporenko.github.io/ember-models-table/v.3/docs/classes/Utils.ModelsTableColumn.html)

## License

### DIFF
--- a/app/pipeline/secrets/template.hbs
+++ b/app/pipeline/secrets/template.hbs
@@ -17,6 +17,7 @@
   @onRefreshToken={{action "refreshPipelineToken"}}
   @tokenName="Pipeline"
   @tokenScope="this pipeline"
+  style="padding: 25px;"
 />
 
 {{outlet}}
--- a/app/pipeline/secrets/template.hbs
+++ b/app/pipeline/secrets/template.hbs
@@ -17,7 +17,6 @@
   @onRefreshToken={{action "refreshPipelineToken"}}
   @tokenName="Pipeline"
   @tokenScope="this pipeline"
-  style="padding: 25px;"
 />
 
 {{outlet}}
--- a/app/components/build-step-collection/styles.scss
+++ b/app/components/build-step-collection/styles.scss
@@ -13,6 +13,10 @@
     }
   }
 
+  li.nav-item {
+    padding: 10px;
+  }
+
   ul {
     list-style: none;
     padding: 0;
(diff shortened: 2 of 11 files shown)


app/pipeline/secrets/template.hbs
```diff
@@ -17,6 +17,7 @@
   @onRefreshToken={{action "refreshPipelineToken"}}
   @tokenName="Pipeline"
   @tokenScope="this pipeline"
+  style="padding: 25px;"
```
adong, line 20: Can this inline style move to `app/components/token-list/style.scss` as well?

yk634: Fixed.

kumada626 (approved):

tkyi (approved):

PR #899, opened 2023-03-10
fix: syntax errors after ember 4 upgrade

adong:
## Context
Sometimes we see broken pages for these 2 reasons: 

 

1) Strict mode in handlebars template files `template.hbs`. 

Error message sample:
```
Uncaught (in promise) Error: Attempted to resolve a value in a strict mode template, but that value was not in scope: isShowingModal
```

image

Before: 
```hbs
{{#if isShowingModal}}

```

After
```
{{#if this.isShowingModal}}
```
See https://github.com/emberjs/rfcs/blob/master/text/0496-handlebars-strict-mode.md#2-no-implicit-this-fallback

2) Change the implicit `set` to explicit `this.set`, for Ember version [3.20.0+](https://github.com/emberjs/ember.js/pull/18961) for an implicit set.

Error Message: 
```
Uncaught (in promise) Error: Assertion Failed: You attempted to update <screwdriver-ui@model:job::ember545:4816773>._isDisabled to "false", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.
```

Before:

```js
routeParams: computed('model', {
    set(_, value) {
      return (this._routeParams = value);
    }
}
```

After: 
```js
routeParams: computed('model', {
    set(_, value) {
      this.set('_routeParams', value);

      return value;
    }
}
```

----

> You must return the new intended value of the computed property from the setter function.

I also updated to have return value explicitly as well, from

```js
return this.set('_isInactivePipeline', value);
```

To
 
```js
this.set('_isInactivePipeline', value);

return value;
```

## Objective
This PR is merely to solve the blank pages issue, and we need to consider a path to move forward and use the latest glimmer syntax. 

Such as: we need to move away from [computed properties](https://guides.emberjs.com/v3.5.0/object-model/computed-properties/#toc_setting-computed-properties) 
image

To `tracked` [computed values](https://guides.emberjs.com/v4.4.0/components/component-state-and-actions/#toc_computed-values)
image

## References
1) [computed properties](https://guides.emberjs.com/v3.5.0/object-model/computed-properties/#toc_setting-computed-properties) 
2) [Why do I not need to mark the property as tracked?](https://discuss.emberjs.com/t/why-do-i-not-need-to-mark-the-property-as-tracked/18093/3)

https://github.com/screwdriver-cd/screwdriver/issues/2838

## License

### DIFF
--- a/app/commands/controller.js
+++ b/app/commands/controller.js
@@ -14,7 +14,9 @@ export default Controller.extend({
       return params;
     },
     set(_, value) {
-      return (this._routeParams = value);
+      this.set('_routeParams', value);
+
+      return value;
     }
   }),
   crumbs: computed('routeParams', {
--- a/app/components/build-banner/component.js
+++ b/app/components/build-banner/component.js
@@ -40,7 +40,9 @@ export default Component.extend({
       return url.split('/').pop();
     },
     set(_, value) {
-      return set(this, '_prNumber', value);
+      set(this, '_prNumber', value);
+
+      return value;
     }
   }),
 
--- a/app/components/build-step-collection/component.js
+++ b/app/components/build-step-collection/component.js
@@ -29,7 +29,9 @@ export default Component.extend({
       return '';
     },
     set(_, value) {
-      return set(this, '_iframeUrl', value);
+      set(this, '_iframeUrl', value);
+
+      return value;
     }
   }),
   selectedStep: computed(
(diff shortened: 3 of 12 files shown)


jithine (approved):
👏

PR #908, opened 2023-03-21
fix: unable to see build display name

banusha2:
## Context

Unable to see display name on Build Details or Options, and unable to see "real" name on Overview.

## Objective
All UI views should be consistent about which name is used for a job.
image

## References
https://github.com/screwdriver-cd/screwdriver/issues/2773

## License

### DIFF
--- a/app/components/pipeline-list-job-cell/template.hbs
+++ b/app/components/pipeline-list-job-cell/template.hbs
@@ -15,7 +15,7 @@
   {{/if}}
   <div class="job-name">
     <BsTooltip @placement="bottom">
-      {{this.record.job.jobName}} {{this.build.status}}
+      {{this.record.job.jobName}}
     </BsTooltip>
     {{this.jobName}}
   </div>
--- a/app/components/pipeline-list-job-cell/template.hbs
+++ b/app/components/pipeline-list-job-cell/template.hbs
@@ -14,6 +14,9 @@
     <span class="fa"></span>
   {{/if}}
   <div class="job-name">
+    <BsTooltip @placement="bottom">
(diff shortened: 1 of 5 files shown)


tkyi:
Can't tell from the screenshot what is different. Is it just the display name for the pipeline at the top?

tests/integration/components/pipeline-list-view/component-test.js
```diff
@@ -613,7 +646,10 @@ module('Integration | Component | pipeline list view', function (hooks) {
             startTime: '2020-04-16T01:30:01.447',
             endTime: null
           }
-        ]
+        ],
+        annotations: {
```
adong, line 650: These are fixing the existing tests, can you add a test to check the displayName is used instead of the job name?

adong (changes requested):

tests/integration/components/pipeline-list-view/component-test.js
banusha2: Yes

app/components/pipeline-list-job-cell/template.hbs
```diff
@@ -14,6 +14,9 @@
     <span class="fa"></span>
   {{/if}}
   <div class="job-name">
+    <BsTooltip @placement="bottom">
+      {{this.record.job.jobName}} {{this.build.status}}
```
adong, line 18: can you remove the `{{this.build.status}}` when hovering over the name?

adong: I think the `status` is can be shown when hovering over the `job-cell`, which has the dot icon.

adong (approved):

PR #912, opened 2023-03-30
user opt auto scroll

banusha2:
## License

### DIFF
--- a/app/components/build-log/component.js
+++ b/app/components/build-log/component.js
@@ -16,6 +16,7 @@ export default Component.extend({
   fullScreen: false,
   lineWrap: true,
   autoscroll: true,
+  autoscrollSelected: true,
   isFetching: false,
   isDownloading: false,
   inProgress: false,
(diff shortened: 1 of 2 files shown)


app/components/build-log/template.hbs
```diff
@@ -16,6 +16,12 @@
         <span class="content"></span>
       </div>
       <div>
+        <a>
+          <BsTooltip @placement="bottom" >
+            auto scroll bottom when new log
+          </BsTooltip>
+          <Input @type="checkbox" @checked={{this.autoscrollSelected}} class="checkbox-input" />
```
adong, line 23: variable case autoscrollSelected => autoScrollSelected

app/components/build-log/component.js
```diff
@@ -424,6 +428,9 @@ export default Component.extend({
         this.element.querySelectorAll('.bottom')[0].getBoundingClientRect()
           .top < 1500
       );
+      if(event.target.scrollTop === 0) {
```
adong, line 431:
```
      if(event.target.scrollTop === 0) {
```

I dont get the logics here, what are you trying to do?

adong: also indention

adong (changes requested):

app/components/build-log/template.hbs
banusha2: Done

app/components/build-log/component.js
```diff
@@ -424,6 +428,9 @@ export default Component.extend({
         this.element.querySelectorAll('.bottom')[0].getBoundingClientRect()
           .top < 1500
       );
+      if (event.target.scrollTop === 0) {
```
adong, line 431:
```
If the user opt-in to new behavior: 
	if the user scrolled up:
		stop auto-scroll to the bottom.
	if the user scrolled down to the bottom:
		if has more logs to load:
			re-enable auto-scroll down to bottom.
		if does not have more logs to load:
			do nothing

if the user does not opt-in to new behavior:
	do nothing
```


This code does not reflect the pseudo logic here. `event` variable does not even exist.

adong (changes requested):

PR #921, opened 2023-04-24
fix: selected event from url is not filtered

adong:
## Context

With the event filter ON, and users visit the the url with given event, which belongs to the part of the filters, we should show the event.

## Objective

For example, event 735440 is a scheduled event, which ought not to be displayed, yet if users share the URL to other users and visit the URL directly, 
https://cd.screwdriver.cd/pipelines/11326/events/735440

Users should see the event on the right side regardless
image

## License

### DIFF
--- a/app/components/pipeline-events/component.js
+++ b/app/components/pipeline-events/component.js
@@ -385,7 +385,6 @@ export default Component.extend(ModelReloaderMixin, {
           filteredPaginateEvents
         );
         const selectedEventId = this.selected;
-        const selectedEvent = this.modelEvents.findBy('id', selectedEventId);
 
         let filteredEvents = pipelineEvents;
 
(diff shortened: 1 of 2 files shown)


app/components/pipeline-events/component.js
```diff
@@ -404,6 +407,11 @@ export default Component.extend(ModelReloaderMixin, {
           );
         }
 
+        // ensure selected event from url is not filtered
+        if (!filteredEvents.findBy('id', selectedEventId)) {
```
adong: This is the original approach, just add it back.

But this way will not persevere the event list order, so it's not used. 

The new approach is add the logics when filtering out events, see the latest code

tkyi (dismissed):

tkyi (approved):

PR #932, opened 2023-05-23
fix(2838): icon changes without reloading when switching job state

yk634:
## Context

This is a bug fix for No.1 of comment below.
https://github.com/screwdriver-cd/screwdriver/issues/2838#issuecomment-1549931089

## Objective

In Ember@3, pipeline events were implemented as controllers, but with Ember Upgrade, they have been moved to components.

The action refreshModel() defined in router can no longer be executed, so it should be passed as an argument from the contoroller and executed.

## References

Ember Upgrade PR (https://github.com/screwdriver-cd/ui/pull/844)

## License


ibu1224 (approved):
LGTM

kumada626 (approved):

PR #950, opened 2023-08-31
fix: build warning should fetch from meta.build instead

klu909:
## Context

it's an oversight that warning still fetching from build.meta.meta.warning, it should be build.meta.build.warning instead
## Objective

build warning should fetch from meta.build instead
## License


jithine:
so `meta.warning` is not used at all ? it should have been `build.warning` ?

adong (approved):

adong:
Approved, please update the doc if needed. 👍

PR #952, opened 2023-09-07
feat(2862): Allow template owners (repo admins) to delete a template version

sagar1312:
## Context
PR https://github.com/screwdriver-cd/ui/pull/937 add in template detail view to delete a specific version on the template.
But, the delete action was only available to Screwdriver admins.

API to delete a template version also allow template owners (repo admins) to delete a specific version

## Objective
Make delete action available to all users.

If a user who do not have admin permission on the template repo attempts to delete a template version, API would fail with 403 and below error will be displayed in the UI
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/2862

## License

### DIFF
--- a/app/components/template-versions/component.js
+++ b/app/components/template-versions/component.js
@@ -92,62 +92,55 @@ export default Component.extend({
     this.set('jobsDetails', []);
   },
 
-  data: computed(
-    'isAdmin',
-    'removeVersion',
-    'templates.[]',
-    'timestampPreference',
-    {
-      get() {
-        const { templates, isAdmin, timestampPreference } = this;
+  data: computed('removeVersion', 'templates.[]', 'timestampPreference', {
+    get() {
+      const { templates, timestampPreference } = this;
 
-        return templates.map(template => {
-          const {
-            name,
-            namespace,
-            fullName,
-            version,
-            createTime,
-            trusted,
-            metrics,
-            tag
-          } = template;
+      return templates.map(template => {
+        const {
+          name,
+          namespace,
+          fullName,
+          version,
+          createTime,
+          trusted,
+          metrics,
+          tag
+        } = template;
 
-          let publishedTime;
+        let publishedTime;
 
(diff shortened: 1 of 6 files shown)


app/components/template-versions/component.js
```diff
@@ -81,61 +81,54 @@ export default Component.extend({
...
-    'templates.[]',
-    'timestampPreference',
-    {
-      get() {
-        const { templates, isAdmin, timestampPreference } = this;
+  data: computed('removeVersion', 'templates.[]', 'timestampPreference', {
+    get() {
+      const { templates, timestampPreference } = this;
 
-        return templates.map(template => {
```
adong, line 93: I wonder if there's a way to expand the scope of `isAdmin` to
`hasAccess: sdAdmins || templateOwners`

This way only some people see the deletion button. Others dont see the buttons so, they won't try to click on them and see the error message.

I took a look at both templates' responses, and we don't expose the repo owners and only have the email info. So we don't have that piece info of `templateOwners` to compare current user to.

tkyi: I agree, I generally feel like things should not show up if a person doesn't have permission.

sagar1312: I too started to make changes to include templateOwners in the permission check assuming permissions check are being done in other pages in the UI.

Looks only check UI performs is to distinguish SD admins vs non SD admins based on the `scope` from the token.
Repo Permissions check (or permission checks in general) are being deferred to the API across the board.
For Ex: 
1. All the actions (except user preferences) on the pipeline `Options` tab are exposed to all the users and API throws `403` when a user who is not the pipeline admin performs those actions.
2. Adding `secrets` or `tokens` to a pipeline
3. Removing a pipeline from a collection not owned by the user
4. Deleting inactive child pipelines
5. Starting/creating events
... more

Performing permission check needs to be addressed across the board. I believe, this PR should not overloaded with that task.

adong (approved):

PR #956, opened 2023-10-13
fix: only mark warning on successful build

klu909:
## Context

it could confusing to users when failed builds are marked with warning
## Objective

only mark successful builds
## License

### DIFF
--- a/app/components/workflow-graph-d3/component.js
+++ b/app/components/workflow-graph-d3/component.js
@@ -62,12 +62,7 @@ export default Component.extend({
         // set warning status if has warning
         if (builds.length) {
           builds.forEach(build => {
-            if (
-              build.meta &&
-              build.meta.build &&
-              build.meta.build.warning &&
-              build.status === 'SUCCESS'
-            ) {
+            if (build?.meta?.build?.warning && build.status === 'SUCCESS') {
               build.status = 'WARNING';
             }
           });
--- a/app/components/workflow-graph-d3/component.js
+++ b/app/components/workflow-graph-d3/component.js
@@ -62,7 +62,12 @@ export default Component.extend({
         // set warning status if has warning
         if (builds.length) {
           builds.forEach(build => {
-            if (build.meta && build.meta.build && build.meta.build.warning) {
+            if (
+              build.meta &&
+              build.meta.build &&
+              build.meta.build.warning &&
+              build.status === 'SUCCESS'
+            ) {
               build.status = 'WARNING';
             }
           });
--- a/app/pipeline/jobs/index/controller.js
+++ b/app/pipeline/jobs/index/controller.js
@@ -122,12 +122,7 @@ export default Controller.extend(ModelReloaderMixin, {
 
           if (nextJobDetail.builds) {
             nextJobDetail.builds.forEach(build => {
-              if (
-                build.meta &&
-                build.meta.build &&
-                build.meta.build.warning &&
-                build.status === 'SUCCESS'
-              ) {
+              if (build?.meta?.build?.warning && build.status === 'SUCCESS') {
                 build.status = 'WARNING';
               }
             });
--- a/app/pipeline/jobs/index/controller.js
+++ b/app/pipeline/jobs/index/controller.js
@@ -122,7 +122,12 @@ export default Controller.extend(ModelReloaderMixin, {
 
           if (nextJobDetail.builds) {
             nextJobDetail.builds.forEach(build => {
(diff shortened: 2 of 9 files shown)


jithine (commented):
This looks like something UI shouldn't handle directly. Should we leave the STATUS override logic to API ?

klu909:
> This looks like something UI shouldn't handle directly. Should we leave the STATUS override logic to API ?

we will keep it this way for now, main value was warning in build detail page which you wouldn't really get when fetching builds by calling api directly

jithine (approved):

app/components/workflow-graph-d3/component.js
```diff
@@ -62,7 +62,12 @@ export default Component.extend({
         // set warning status if has warning
         if (builds.length) {
           builds.forEach(build => {
-            if (build.meta && build.meta.build && build.meta.build.warning) {
+            if (
+              build.meta &&
+              build.meta.build &&
+              build.meta.build.warning &&
+              build.status === 'SUCCESS'
+            ) {
```
adong, line 70: I'd say use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

`build?.meta?.build?.warning && build.status === 'SUCCESS'`

app/pipeline/jobs/index/controller.js
```diff
@@ -122,7 +122,12 @@ export default Controller.extend(ModelReloaderMixin, {
 
           if (nextJobDetail.builds) {
             nextJobDetail.builds.forEach(build => {
-              if (build.meta && build.meta.build && build.meta.build.warning) {
+              if (
+                build.meta &&
+                build.meta.build &&
+                build.meta.build.warning &&
+                build.status === 'SUCCESS'
+              ) {
```
adong, line 130: Same `build?.meta?.build?.warning && build.status === 'SUCCESS'`

app/utils/metric.js
```diff
@@ -100,13 +103,38 @@ const formatMetrics = metrics => {
...
-    duration: event.duration || 0,
-    statusColor: getColor(event.status.toLowerCase())
-  }));
+  const eventsInfo = metrics.map(event => {
+    let eventStatus = event.status.toLowerCase();
+
+    // only overwrite success event with warning
+    if (event && event.builds && eventStatus === 'success') {
+      event.builds.forEach(build => {
+        if (build.meta && build.meta.build && build.meta.build.warning) {
```
adong, line 112: `if (build?.meta?.build?.warning)`

app/utils/metric.js
```diff
@@ -100,13 +103,38 @@ const formatMetrics = metrics => {
...
+      statusColor: getColor(eventStatus)
+    };
+  });
+
+  let lastEventStatus = lastEvent.status.toLowerCase();
+
+  // only overwrite success event with warning
+  if (lastEvent && lastEvent.builds && lastEventStatus === 'success') {
+    lastEvent.builds.forEach(build => {
+      if (build.meta && build.meta.build && build.meta.build.warning) {
```
adong, line 129: Same `if (build?.meta?.build?.warning)`

app/utils/metric.js
```diff
@@ -100,13 +103,38 @@ const formatMetrics = metrics => {
...
+    return {
+      duration: event.duration || 0,
+      statusColor: getColor(eventStatus)
+    };
+  });
+
+  let lastEventStatus = lastEvent.status.toLowerCase();
+
+  // only overwrite success event with warning
+  if (lastEvent && lastEvent.builds && lastEventStatus === 'success') {
```
adong, line 127: For consistency reasons from the current file, I would use upper-case instead.
`lastEvent?.builds && build.status === 'SUCCESS'` 

Remove 
```
let lastEventStatus = lastEvent.status.toLowerCase();
```

app/utils/metric.js
```diff
@@ -100,13 +103,38 @@ const formatMetrics = metrics => {
...
+      if (build.meta && build.meta.build && build.meta.build.warning) {
+        lastEventStatus = 'warning';
+      }
+    });
+  }
+
   const lastEventInfo = {
     startTime: `${lastEventStartMonth}/${lastEventStartDay}/${lastEventStartYear}`,
-    statusColor: getColor(lastEvent.status.toLowerCase()),
+    statusColor: getColor(lastEventStatus),
```
adong, line 137: And keep `statusColor: getColor(lastEvent.status.toLowerCase())`

adong (changes requested):
See comments for code readability improvements

adong (approved):

PR #959, opened 2023-10-31
feat: add sonar badge [4]

adong:
## Context

Add sonar badge display to the pipeline page and configurable in pipeline option page

## Objective
image
image

Tests pending
## License


jithine:
what we need is for a way to get the sonar badge information auto populated. I wonder if the coverage bookend can do this. It does make an API call explicitly to create the project, I think that API call should update the Pipeline information as well.

adong:
> what we need is for a way to get the sonar badge information auto populated. I wonder if the coverage bookend can do this. It does make an API call explicitly to create the project, I think that API call should update the Pipeline information as well.

Okay, maybe in the coverage bookend script, we can set this url inside the `pipeline.badge.sonar`

tkyi (dismissed):

app/components/pipeline-header/component.js
```diff
@@ -26,19 +48,30 @@ export default Component.extend({
       };
     }
   }),
-  sameRepoPipeline: computed('pipeline', {
+  sameRepoPipeline: computed('pipeline.scmRepo.name', 'pipeline.{id,scmUri}', {
```
jithine, line 51: what are these changes for ?

adong: I think these codes are added when rebased with master

PR #960, opened 2023-11-06
fix(2862): Improve default display performance of template metrics

yk634:
## Context

Fix the problem in the following comment.
https://github.com/screwdriver-cd/screwdriver/issues/2862#issuecomment-1786509652

When tallying the number of builds using a template, it may take longer if createTime is specified.

## Objective

When the template detail page is opened, the default is to display it without specifying a time period, thereby reducing the time required to display it.

Also add `all` button to aggregate all periods.
image

## License

### DIFF
--- a/app/components/time-range-picker/component.js
+++ b/app/components/time-range-picker/component.js
@@ -20,6 +20,8 @@ export default Component.extend({
   // flatpickr addon seems to prefer dates in string
   customRange: computed('startTime', 'endTime', {
     get() {
+      if (!this.get('startTime')) return null;
+
       return ['startTime', 'endTime'].map(t =>
         toCustomLocaleString(new Date(this.get(t)), {
           options: {
--- a/app/components/time-range-picker/component.js
+++ b/app/components/time-range-picker/component.js
@@ -15,7 +15,8 @@ export default Component.extend({
     { alias: '1mo', value: '1mo' },
     { alias: '3mo', value: '3mo' },
     { alias: '6mo', value: '180d' },
-    { alias: '1yr', value: '1yr' }
+    { alias: '1yr', value: '1yr' },
+    { alias: 'all', value: null }
   ],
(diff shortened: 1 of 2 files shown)


kumada626 (dismissed):

ibu1224 (approved):

y-oksaku (approved):

sagar1312:
We had discussed to whether to include the preset option `All` to show the all time usage of templates in builds.
We decided not to have this option as we usually cleanup to delete builds older than 1 year from the system.

Also we want, last 1 year as the default range.

Please take a look at https://github.com/screwdriver-cd/data-schema/pull/541 for fix the performance issue

yk634:
@sagar1312
Thanks for updating the index to improve performance!
But some people operate without deleting builds older than 1 year, so I think it is appropriate to default to show the values for the entire period.

sagar1312:
I am not against having "All" option.
But it need not be the default option.

Before introducing date/time filter, we always used to show all time usage.
With the introduction of date/time filter, we want to highlight the recent activity (IMO 1yr is also a very long period. May be we should highlight last 3 months) on each version.

With this, we also want to encourage the template owners to deprecate and delete older versions.

I would let @jithine weigh in on what should be the default date range.

jithine:
Making `all` default would put pressure on DB for templates with very large number of builds.

yk634:
@jithine
I understand. But by default, there is a double aggregation for the entire period for template-header and for 1 year for per-version metrics.
https://github.com/screwdriver-cd/ui/blob/f127799f0d4a572aaf9502ca30aec0b7555db7cf/app/templates/detail/route.js#L37-L50

Since aggregating the entire period for the template-header would be a load on the DB, modify it to refer to the 1 year aggregate that is retrieved by default for per-version metrics.

Close this PR and create this modification as a new PR.

PR #961, opened 2023-11-09
fix(2862): Improve the load on the DB for template metrics

yk634:
## Context

Fix the problem in the following comment.
https://github.com/screwdriver-cd/ui/pull/960#issuecomment-1803212548

Currently, the total for the entire period is being aggregated to show the number of builds in the template-header, and the total for one year to show the number of builds per version.

## Objective

Because double aggregation is a burden on the DB, modify the values displayed in template-header to refer to one year's aggregation.
Also, add an `all` button to display version metrics for the entire period

## License

### DIFF
--- a/app/templates/detail/route.js
+++ b/app/templates/detail/route.js
@@ -20,6 +20,7 @@ export default Route.extend({
     // these are used for querying, so they are in ISO8601 format
     this.set('startTime', startTime);
     this.set('endTime', endTime);
+
     this.set('fetchAll', true);
     this.set('fetchFiltered', false);
   },
--- a/app/templates/detail/route.js
+++ b/app/templates/detail/route.js
@@ -1,6 +1,7 @@
 import { inject as service } from '@ember/service';
 import RSVP from 'rsvp';
 import Route from '@ember/routing/route';
+import timeRange from 'screwdriver-ui/utils/time-range';
 
 export default Route.extend({
   router: service(),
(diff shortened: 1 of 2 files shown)


yakanechi (approved):

y-oksaku (approved):

ibu1224 (approved):

tkyi (approved):

app/templates/detail/route.js
```diff
@@ -33,28 +33,33 @@ export default Route.extend({
...
+
+    if (startTime) {
+      query = {
+        startTime,
+        endTime
+      };
+    }
+
     return RSVP.all([
-      fetchAll
```
sagar1312, line 37: I agree that we need to optimize the API request.

The date/time filter should be applicable only for the table where we list all the versions.

In the header (or detail view) where we display the details of the selected version, we need to display the all time usage.

So here, we still need to fetch the template without date/time filter. But limit it to the version being displayed in the header (or detail view)

yk634: I figured that getting metrics for all periods by default would overload the DB and should be fixed.
https://github.com/screwdriver-cd/ui/pull/960#issuecomment-1799469421

But is it correct that it is necessary to get metrics for all periods in order to display them in the template-header?

If so, modify to run two GET APIs, one specifying all periods and one specifying one year, as before.

PR #966, opened 2023-11-16
fix:(2215): fix issue of moving between pipelines in the same branch (rootDir)

itleigns:
## Context
In #964, we made sure to verify if each of the pipeline options for a switch was of the same repository. However, this has caused improper functioning when root directory is set.

## Objective
We have ensured correct operation even when root directory is set. In cases where the root directory is set, the format of `scmUri` is `<SCM>:<repository ID>:<branch>:<rootDir>`, so we decided to compare up to the second `:`.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2215

## License

### DIFF
--- a/app/components/pipeline-header/component.js
+++ b/app/components/pipeline-header/component.js
@@ -28,8 +28,9 @@ export default Component.extend({
   }),
   sameRepoPipeline: computed('pipeline', {
     get() {
+      const repositoryUri = this.pipeline.scmUri.substring(0,this.pipeline.scmUri.indexOf(':', this.pipeline.scmUri.indexOf(':') + 1))
       return this.pipelineService.getSiblingPipeline(this.pipeline.scmRepo.name).then(value =>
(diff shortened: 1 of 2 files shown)


app/components/pipeline-header/component.js
```diff
@@ -28,8 +28,9 @@ export default Component.extend({
   }),
   sameRepoPipeline: computed('pipeline', {
     get() {
+      const repositoryUri = this.pipeline.scmUri.substring(0,this.pipeline.scmUri.indexOf(':', this.pipeline.scmUri.indexOf(':') + 1))
```
y-oksaku, line 31: I think this might be better.

```js
const [scm, repo] = this.pipeline.scmUri.split(':');

...

filter(pipe => {
  const [s, r] = pipe.scmUri.split(':');
  return this.pipeline.id === pipe.id && scm === s && repo === r;
})....
```

or

```js
const  repositoryUri = this.pipeline.scmUri.split(':').slice(0, 2).join(':');
```

itleigns: Thanks!

y-oksaku (approved):

ibu1224 (approved):

tkyi (approved):

PR #969, opened 2023-11-28
fix(2924): Adjust font size in workflow to prevent text overflow

itleigns:
## Context
With jobs that include a large number of wide-width characters, the text overflows in the workflow graph.

Q7

## Objective
We reduce the font size when overflowing.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2924

## License

### DIFF
--- a/app/components/workflow-graph-d3/component.js
+++ b/app/components/workflow-graph-d3/component.js
@@ -348,6 +348,15 @@ export default Component.extend({
           return 'graph-label';
         })
         .attr('font-size', `${TITLE_SIZE}px`)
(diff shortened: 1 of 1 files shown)


itleigns:
After
IssuePipeline

QQQQPipeline

itleigns:
There was also an option to modify [this part](https://github.com/screwdriver-cd/ui/blob/master/app/components/workflow-graph-d3/component.js#L216-L219) and increase the width, but for example with 'Q', we need to set the value to 10, which is too large.

tkyi (approved):

PR #974, opened 2023-12-07
fix: sanitize user input before displaying info message

klu909:
## Context

sanitize user input before displaying info message
## Objective

before
Screenshot 2023-12-07 at 3 44 49 PM
after
Screenshot 2023-12-07 at 3 44 22 PM

## License

### DIFF
--- a/app/components/info-message/template.hbs
+++ b/app/components/info-message/template.hbs
@@ -9,7 +9,7 @@
         {{! template-lint-disable no-triple-curlies }}
         <FaIcon @icon={{@icon}} />
         <span>
-          {{{@message}}} Please login to
+          {{{this.sanitizedMessgae}}} Please login to
         </span>
 
         <a href="#authenticate" {{action "authenticate" @scmContext}}>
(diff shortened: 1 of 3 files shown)


app/helpers/sanitize-string.js
```diff
@@ -0,0 +1,25 @@
...
+import { htmlSafe } from '@ember/template';
+
+/**
+ * Sanitizes given string
+ * @method sanitizeString
+ * @param  {String}  string  String to be sanitized
+ * @return {String}          Sanitized string
+ */
+export function sanitizeString(string) {
+  const escape = {
```
jithine, line 11: why are we having this manual mapping? I would imagine a standard library to handle this.

klu909: i didn't find native library/api for this. Open to change it though

app/components/info-message/template.hbs
```diff
@@ -18,14 +18,14 @@
       {{else}}
         <FaIcon @icon={{@icon}} />
         <span>
-          {{{@message}}}
+          {{{this.sanitizedMessgae}}}
```
adong, line 21: I dont think you need the computed property stuff. How about this?

```
{{{(sanitizeString @message}}
```

adong (approved):

PR #977, opened 2024-01-02
feat(#2970): use DOM parser instead of Regex to sanitize user html input

adong:
## Context

Previous PR https://github.com/screwdriver-cd/ui/pull/974 will filter almost every tags, and will not allow `anchor` tag to work

## Objective

Instead of using regex, use DOM Parser to filter out allow/deny elements and attributes

Screenshots:

image

image

## References

Issue: https://github.com/screwdriver-cd/screwdriver/issues/2970

Test Links: 
1) https://cd.screwdriver.cd/pipelines/10293/builds/913516
2) https://cd.screwdriver.cd/pipelines/12888/builds/915788

## License

### DIFF
--- a/app/helpers/input-sanitizer.js
+++ b/app/helpers/input-sanitizer.js
@@ -1,5 +1,57 @@
 import { helper } from '@ember/component/helper';
-import { htmlSafe } from '@ember/template';
+
+/**
+ * Sanitizer which filter out deny elements, and only keep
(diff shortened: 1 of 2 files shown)


jithine (approved):
My only concern is the custom code we are having, I would imaging some standard Ember/JS library to abstract this logic

klu909 (approved):

PR #981, opened 2024-01-12
fix(2982): Reload page after sync pipeline

kumada626:
## Context

If there is an update to the pipeline information when the sync button is clicked, users have to reload page manually to show updated info on UI.
Therefore, the page is automatically reloaded after the sync button is clicked.

## Objective

Reload page after sync pipeline.
Sync webhooks and PRs do not update pipeline info. So we have to reload page only after sync pipeline.

## References

https://github.com/screwdriver-cd/screwdriver/issues/2982

## License

### DIFF
--- a/app/components/pipeline-options/component.js
+++ b/app/components/pipeline-options/component.js
@@ -307,12 +307,16 @@ export default Component.extend({
 
       return this.sync
         .syncRequests(this.get('pipeline.id'), syncPath)
-        .then(() =>
-          this.setProperties({
-            successMessage: 'Pipeline sync successful',
-            errorMessage: ''
-          })
-        )
+        .then(() => {
+          if (syncPath) {
+            this.setProperties({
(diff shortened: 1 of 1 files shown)


app/components/pipeline-options/component.js
```diff
@@ -225,6 +225,9 @@ export default Component.extend({
       this.set('errorMessage', error);
     }
   },
+  reloadPage() {
+    window.location.reload(true);
```
jithine, line 229: is there an ember way to refresh the page data without a physical page reload ?

kumada626: Since it was feasible by reloading the model, I modified to reload the pipeline model without a physical page reload.

itleigns (approved):

ibu1224 (approved):

jithine (approved):

PR #984, opened 2024-01-25
feat: add functions to validate pipeline template

pritamstyz4ever:
## Context

Enhance validator to support pipeline templates

## Objective

This PR adds logic to check if the supplied yaml is a pipeline template and calls the back API accordingly

## License
https://github.com/screwdriver-cd/screwdriver/issues/2135


pritamstyz4ever:
@adong should we also include text for pipeline templates here
Screenshot 2024-01-25 at 12 38 12 PM

app/validator/service.js
```diff
@@ -12,7 +12,11 @@ export default Service.extend({
    * @return {Boolean}
    */
   isTemplate(yaml) {
-    return /^name|\n+name: |\n+namespace: /.test(yaml);
+    return /^name|\n+name: |\n+namespace: |\n+config:\s*\n\s*jobs: /.test(yaml);
```
adong, line 15: Maybe change to `isJobTemplate`. Yea, it's tedious

app/validator/service.js
```diff
@@ -22,10 +26,12 @@ export default Service.extend({
    * @return {Promise}
    */
   getValidationResults(yaml) {
-    let url = `${ENV.APP.SDAPI_HOSTNAME}/v4/validator`;
+    let url = `${ENV.APP.SDAPI_HOSTNAME}/v4`;
 
-    if (this.isTemplate(yaml)) {
-      url += '/template';
+    if (this.isPipelineTemplate(yaml)) {
+      url += '/pipeline/template/validate';
```
adong, line 32: We dont validate against API directly. We validate against this template validator.

Repo is: https://github.com/screwdriver-cd/template-validator

Can we move this endpoint to `/validator/template/pipeline`

tests/unit/validator/service-test.js
```diff
@@ -54,6 +59,28 @@ module('Unit | Service | validator', function (hooks) {
         JSON.stringify(EXAMPLE_TEMPLATE_PAYLOAD)
       ];
     });
+
+    server.post(
+      'http://localhost:8080/v4/pipeline/template/validate',
```
adong, line 64: I think it needs some refactoring for the encapsulation in validator

adong (changes requested):

app/validator/service.js
pritamstyz4ever: I don't think we validate directly via template-validator, we call the API which in turn calls the templateValidator
Screenshot 2024-01-25 at 1 08 07 PM

adong: Correct, we still calls API. API contains https://github.com/screwdriver-cd/screwdriver/blob/d165071e9ec163c1c746f45781b3176d8a68c4e2/plugins/validator.js, which calls https://github.com/screwdriver-cd/screwdriver/blob/d165071e9ec163c1c746f45781b3176d8a68c4e2/plugins/template-validator.js#L5

I think the template validator has been updated to support job template, https://github.com/screwdriver-cd/screwdriver/commit/2bf13ed3971e9c627aeb5657d4eee6dc211e7e85#diff-12e32795e740f0d1bae40c44544e6f6ce975a060c1caef154ce087c7245a4d3dL6

Maybe @klu909 could help you out.

pritamstyz4ever: @adong the pipeline template validation API is here https://github.com/screwdriver-cd/screwdriver/blob/master/plugins/pipelines/templates/validate.js#L10

adong: I'm saying, our validator APIs are:

```
/v4/validator
/v4/validator/template   # for job template
```

So the new pipeline template API ought to be
`/v4/validator/template/pipeline`

Instead of 
`/v4/pipeline/template/validate`

VonnyJap: I do agree that consistency matters, like in this case `/v4/validator/template/pipeline` seems to be a better endpoint for validating the pipeline template. 

@pritamstyz4ever - would that be lots of refactoring to do so?

app/validator/service.js
```diff
@@ -7,12 +7,24 @@ export default Service.extend({
...
+  },
+
+  /**
+   * Simple test to determine if yaml looks like a pipeline template file
+   * @method isPipelineTemplate
+   * @param  {String}   yaml Raw yaml text
+   * @return {Boolean}
+   */
+  isPipelineTemplate(yaml) {
+    return /^(?=.*\bnamespace:)(?=.*\bname:)(?=.*\bconfig:.*\bjobs:).*/.test(
```
adong, line 25: Question: what is the subtle difference between `isJobTemplate` and `isPipelineTemplate` in these 2 regex(es)? How does UI know a given string is a `job` or `pipeline` template?

app/validator/service.js
pritamstyz4ever: @VonnyJap  
No, we will not be using `/v4/validator` endpoint as it breaks the design.
All the APIs are structured as per the entity they belong to like `pipelines` , `jobs`, `builds`, `commands` etc. , the validator API is an anti-pattern that has `jobs` and `commands` combined and we don't want to add anything new in there. In the future, we will retire this endpoint and move the validation for jobs to `/templates/validate` and similarly for commands to `/commands/validate`

app/validator/service.js
pritamstyz4ever: both types of templates have different configuration, for pipeline templates we need the config to always specify jobs so we check for `config.jobs`, `namespace` and `name`

VonnyJap (approved):

PR #988, opened 2024-02-06
fix: graph render race condition

minghay:
## Context
There's a race condition when rendering graphs based on whether or not the data is already in the ember data store.

## Objective
Changing the lifecycle back to `didUpdateAttrs` fixes the race condition on the graph being rendered multiple times in a single update.

## License


adong:
Can you attach a screenshot, maybe a gif?

VonnyJap (approved):

PR #991, opened 2024-02-14
fix: update logic for determining when a node can start and restart

minghay:
## Context
The ability for a job to start or restart from the graph node depends on job data returned from the API.  If data is missing from the API, this prevents the UI from being able to create the tooltips correctly.

## Objective
Update the tooltip logic to only rely on the workflow graph information.  Additional logic around the current view is used to further determine which jobs, and their associated workflow graph, can be triggered.

## References
https://github.com/screwdriver-cd/screwdriver/issues/3011

## License

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -2,7 +2,11 @@ import { isActivePipeline } from 'screwdriver-ui/utils/pipeline';
 import Component from '@ember/component';
 import { get, computed, set, setProperties } from '@ember/object';
 import { reject } from 'rsvp';
-import { isRoot } from 'screwdriver-ui/utils/graph-tools';
+import {
+  isRoot,
+  reverseGraph,
+  subgraphFilter
+} from 'screwdriver-ui/utils/graph-tools';
 import { isActiveBuild } from 'screwdriver-ui/utils/build';
 import { copy } from 'ember-copy';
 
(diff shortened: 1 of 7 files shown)


tkyi (approved):

app/components/pipeline-workflow/component.js
```diff
@@ -61,46 +65,37 @@ export default Component.extend({
...
+          return isOnPrPath;
+        }
+        // In the events view:
+        const isOnCommitPath =
+          subgraph.nodes.filter(node => node.name === '~commit').length > 0;
+
+        // If the job is detached, allow restart:  isOnPrPath = false, isOnCommitPath = false
+        // If the job is only on the commit path, allow restart: isOnPrPath = false, isOnCommitPath = true
+        // If the job is on the commit and pr path, allow restart: isOnPrPath = true, isOnCommitPath = true
+        // If the job is only on the pr path, do not allow restart: isOnPrPath = true, isOnCommitPath = false
```
VonnyJap, line 93: I think somehow I recalled that pr job can be started as long as there is a path leading to `~pr` node, regardless of the commit path. Maybe I am wrong?

minghay: That only applies when in the pr view.  This particular section is for the events view, and the new logic is that jobs that are only reachable from the `~pr` node can not be restarted (the user will need to change to the PR view in order to trigger those jobs).

y-oksaku:
@minghay 

Jobs that only have `~pr`, such as the `foo` job below, can no longer use the `Start pipeline from here` in the workflow graph.

```yaml
jobs:
  foo:
    requires: [ ~pr ]
```

On the other hand, I can start a `foo` build from the job list view.
Is this behavior expected?

Thank you.

minghay:
@y-oksaku
For jobs triggered by `~pr`, those jobs can only be restarted when you are in the pull request view (i.e. `/pipelines/{pipelineId}/pulls`), not the events view (i.e. `/pipelines/{pipelineId}/events/{eventId}`).  There are no restrictions for restarting jobs from the job list view.  This is expected behavior.

y-oksaku:
@minghay 

> For jobs triggered by ~pr, those jobs can only be restarted when you are in the pull request view (i.e. /pipelines/{pipelineId}/pulls), not the events view (i.e. /pipelines/{pipelineId}/events/{eventId}). There are no restrictions for restarting jobs from the job list view. This is expected behavior.

I noted it. Thank you.

PR #997, opened 2024-02-22
feat(2669): Optimization of fetching stages for a pipeline

sagar1312:
## Context

Stages associated with a pipeline are being fetched each time an event is selected in the events view.
Since the stages are associated with the `pipeline` and not with individual events, we should cache the stages once fetched for a pipeline.

## Objective

Cache the stages in Ember store that are fetched for a pipeline.

## References
https://github.com/screwdriver-cd/screwdriver/issues/2669

## License


app/pipeline/events/show/route.js
```diff
@@ -39,23 +39,9 @@ export default class PipelineEventsShowRoute extends Route {
...
-  async model() {
-    const { event_id: eventId } = this.paramsFor(this.routeName);
-    const { pipeline_id: pipelineId } = this.paramsFor('pipeline');
-
-    const stages = await this.store.query('stage', {
-      pipelineId,
-      eventId
-    });
-
-    return { stages };
```
adong, line 51: Okay, so instead of getting `stage` data from event show route, we will get that from event route.

sagar1312: yes, similar to `job` data

adong (approved):

PR #998, opened 2024-02-23
fix: updating pipeline settings only requires 1 API call

minghay:
## Context
Updating pipeline settings requires 2 API calls.  Currently, to update the pipeline settings 2 API calls are required because the pipeline setting adapter does not serialize the user settings.  As a result, the current implementation makes an API call to get the current settings and update only the modified pipeline's settings.  As there is no `PATCH` API endpoint, the full settings body must be recreated.

## Objective
Updates the pipeline setting adapter to serialize the user settings to remove the need for an API call.

## License

### DIFF
--- a/app/preference/pipeline/adapter.js
+++ b/app/preference/pipeline/adapter.js
@@ -1,10 +1,13 @@
 import Adapter from 'ember-local-storage/adapters/local';
 import { inject as service } from '@ember/service';
+import { preparePayload } from 'screwdriver-ui/preference/user/serializer';
 
 export default class PreferencePipelineAdapter extends Adapter {
   modelNamespace = 'preference';
 
   @service shuttle;
+
+  @service userSettings;
 
   async createRecord(store, type, snapshot) {
     const serializer = store.serializerFor(type.modelName);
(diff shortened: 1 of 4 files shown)


app/preference/pipeline/adapter.js
```diff
@@ -14,19 +17,22 @@ export default class PreferencePipelineAdapter extends Adapter {
...
+    const data = await this.createPayload();
+
+    data[id] = pipelinePreference;
 
-    return this.shuttle.updateUserPreference(id, pipelinePreference);
+    return this.shuttle.updateUserSetting(data);
   }
 
-  async updateRecord(store, type, snapshot) {
-    const serializer = store.serializerFor(type.modelName);
```
minghay, line 22: There isn't a need to serialize the latest changes anymore since the `preparePayload` function in the user preference serializer already serializes all of the pipeline preferences that have been set.

app/preference/pipeline/adapter.js
```diff
@@ -14,19 +17,22 @@ export default class PreferencePipelineAdapter extends Adapter {
       snapshot,
       { includeId: true }
     );
+    const data = await this.createPayload();
+
+    data[id] = pipelinePreference;
```
minghay, line 22: When a new pipeline preference setting is created, it hasn't been saved to the store quite yet at this point.  As such, we need to manually add it into the body that is sent to the API.

app/shuttle/service.js
```diff
@@ -261,76 +261,6 @@ export default Service.extend({
...
-    const remotePipelineConfig =
-      get(remotePreferences, pipelineId) === undefined
-        ? {}
-        : get(remotePreferences, pipelineId);
-    const localPipelinePreference = await this.store
-      .peekAll('preference/pipeline')
-      .findBy('id', pipelineId);
-
-    // local preference takes precedence
-    if (localPipelinePreference && remotePipelineConfig) {
```
adong, line 284: I wonder how this local/remote settings logic is handled in this PR, since these lines of code are removed.

pritamstyz4ever (approved):

app/shuttle/service.js
minghay: https://github.com/screwdriver-cd/ui/commit/5e0244fe6d19086fbe46ba2e5cb2345eaaa95d3f has the changes where this entire function was migrated to the pipeline service: https://github.com/screwdriver-cd/ui/blob/master/app/pipeline/service.js#L26-L54

adong (approved):

tkyi (approved):

PR #1000, opened 2024-02-27
fix: pr pipeline graph rendering

minghay:
## Context
https://github.com/screwdriver-cd/ui/pull/989 had introduced a side-effect where pipelines in the PR view would always use the workflow graph.  The correct rendering of the pipeline graph in the PR view is to have the complete pipeline graph rendered instead; however, the workflow graph contains the correct build metadata when in the PR view.  As such, to properly render the PR view pipeline graph, data from both graphs need to be merged together.

## Objective
Fixes an issue with PR pipeline graph rendering not displaying the full pipeline graph and also hiding the PR pipeline when that pipeline setting is toggled.

## License


adong (approved):

app/components/workflow-graph-d3/component.js
```diff
@@ -69,7 +71,52 @@ export default Component.extend({
               }
             : this.completeWorkflowGraph;
 
-        const graph = showDownstreamTriggers ? completeGraph : workflowGraph;
+        let graph = showDownstreamTriggers ? completeGraph : workflowGraph;
+
+        if (!this.minified && this.selectedEventObj?.prNum) {
+          // When initially onboarding the screwdriver configuration, there is a node named ~pr:/.*/ in completeGraph that must be merged with the node data in the workflow graph.  Additional edge data clean up must also be done.
+          const prJobWildcard = '~pr:/.*/';
+          const completePrGraph = JSON.parse(JSON.stringify(completeGraph));
```
adong, line 79: I think this will cause errors somehow, just want to point out.

Maybe catch in a try catch block?

PR #1005, opened 2024-03-05
Fix cache checking of event

minghay:
## Context
When loading the pipeline page, multiple API calls are being made to the `/events/{eventId}` endpoint.  Limit the number of API calls to the `events/{eventId}` endpoint by properly checking the store cache first.

## Objective
`Peek` into the ember store cache to get the event data and only make an API call if the store cache does not contain the event.

## License


adong (approved):

tkyi (approved):

sagar1312:
Shouldn't we need to refresh the `event` esp certain dynamic attributes (Ex: `meta.parameters`) could get stale as the builds get executed for that event.

PR #1010, opened 2024-03-07
feat: add validator ui support for pipeline templates

pritamstyz4ever:
## Context

Added validator support to display pipeline template and pipeline template usage

## Objective

The validator should correctly display the pipeline template name, namespace, jobs, etc.
The validator should correctly display the pipeline template used in a pipeline

Template:
Screenshot 2024-03-06 at 7 17 55 PM
Screenshot 2024-03-06 at 7 18 03 PM

Usage:
Screenshot 2024-03-07 at 2 41 11 PM

## References
https://github.com/screwdriver-cd/screwdriver/issues/3052

## License

### DIFF
--- a/tests/integration/components/validator-results/component-test.js
+++ b/tests/integration/components/validator-results/component-test.js
@@ -447,6 +447,37 @@ module('Integration | Component | validator results', function (hooks) {
     assert.dom('h4').hasText('batman/batmobile@1.0.0');
   });
 
+  test('it renders a pipeline template', async function (assert) {
+    this.set('validationMock', {
+      errors: [],
+      template: {
+        namespace: 'batman',
+        name: 'batmobile',
+        version: '1.0.0',
+        config: {
+          shared: {
+            image: 'int-test:1'
+          },
+          jobs: {
+            main: {
+              steps: [{ forgreatjustice: 'ba.sh' }]
+            },
+            publish: {
+              steps: [{ joker: 'ha' }]
+            }
+          }
+        }
+      }
+    });
+
+    await render(
+      hbs`<ValidatorResults @results={{this.validationMock}} @isPipelineTemplate={{true}} />`
+    );
+
+    assert.dom('.error').doesNotExist();
+    assert.dom('h4').hasText('batman/batmobile@1.0.0');
+  });
+
   test('it renders joi error results', async function (assert) {
     this.set('validationMock', {
       errors: [{ message: 'there is an error' }],
--- a/app/components/pipeline-header/component.js
+++ b/app/components/pipeline-header/component.js
@@ -59,7 +59,7 @@ export default Component.extend({
       };
     }
   }),
-  sameRepoPipeline: computed('pipeline', {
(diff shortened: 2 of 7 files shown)


jithine (dismissed):

tests/integration/components/validator-results/component-test.js
```diff
@@ -110,6 +110,37 @@ module('Integration | Component | validator results', function (hooks) {
...
+    this.set('validationMock', {
+      errors: [],
+      template: {
+        namespace: 'batman',
+        name: 'batmobile',
+        version: '1.0.0',
+        config: {
+          shared: {
+            image: 'int-test:1'
+          },
```
sagar1312, line 123: We should use/render the configuration under `shared`.
If few jobs inherit `steps`, `settings` from shared, that is not being visualized in the validator results

sagar1312 (changes requested):

sagar1312:
Also the workflow is not visualized for pipeline template
We should show how the jobs are interconnected in the workflow graph

image

pritamstyz4ever:
> Also the workflow is not visualized for pipeline template We should show how the jobs are interconnected in the workflow graph
> 
> image

Current workflow jobs are not returned by the API for templates. We will need to add that

tests/integration/components/validator-results/component-test.js
pritamstyz4ever: I think shared steps have the same behavior as pipeline, for settings and environment we may need to tweak API to return the merged values

sagar1312:
> > Also the workflow is not visualized for pipeline template We should show how the jobs are interconnected in the workflow graph
> > image
> 
> Currently workflow jobs are not returned by the API for templates. We will need to add that

Yes, the validator API may need to all the config-parser again to merge the `shared` into `jobs`. That would also generate workflow graph.

VonnyJap:
@sagar1312 

> Also the workflow is not visualized for pipeline template
> We should show how the jobs are interconnected in the workflow graph

We can do the visualization as next item to work on. At least the validator can make sure that the config in template is working. Although yes, I agree, without the workflow graph, pipeline template validator is missing the important purpose and becomes mainly just to validate the syntax.

sagar1312:
> @sagar1312
> 
> > Also the workflow is not visualized for pipeline template
> > We should show how the jobs are interconnected in the workflow graph
> 
> We can do the visualization as next item to work on. At least the validator can make sure that the config in template is working. Although yes, I agree, without the workflow graph, pipeline template validator is missing the important purpose and becomes mainly just to validate the syntax.

Sure, we can push changes incrementally. 
Besides visualizing workflow graph, we also need to make sure config under `shared` as explained earlier needs to be somehow (merging it with `jobs`) displayed in the result.

sagar1312 (approved):

sagar1312 (approved):

adong:
Validator: https://cd.screwdriver.cd/validator

```
template: sd-test/example-template@latest
shared:
    environment: 
        FOO: bar
```

PR #1031, opened 2024-03-29
fix: delete pipeline by version only delete the latest one

adong:
## Context

Currently, delete pipeline template version only deletes the latest one.

## Objective
1) delete the pipeline template by version by the target version from table
2) Add more tests

## License

### DIFF
--- a/app/templates/pipeline/detail/controller.js
+++ b/app/templates/pipeline/detail/controller.js
@@ -161,7 +161,7 @@ export default class TemplatesPipelineDetailController extends Controller {
   }
 
   @action
-  removeVersion(fullName, version) {
+  removeVersion(_, version) {
     const { namespace, name } = this.selectedVersionTemplate;
 
     return (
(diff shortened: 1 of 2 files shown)


app/templates/pipeline/detail/controller.js
```diff
@@ -161,8 +161,8 @@ export default class TemplatesPipelineDetailController extends Controller {
   }
 
   @action
-  removeVersion() {
-    const { namespace, name, version } = this.selectedVersionTemplate;
+  removeVersion(fullName, version) {
```
VonnyJap, line 164: when is the `fullName` option here being used in the function? I could not locate the use case.

adong: `fullName` is not used. So, rename `removeVersion(fullName, version)` to ` removeVersion(_, version)`

VonnyJap (approved):

PR #1033, opened 2024-04-03
fix: commands table is not rendered correctly and add more tests

adong:
## Context

commands table is not rendered correctly 

## Objective
To add more integration tests to cover the code changes

## License

### DIFF
--- a/app/components/tc-collection-list/component.js
+++ b/app/components/tc-collection-list/component.js
@@ -18,6 +18,14 @@ export default Component.extend({
     get() {
       const prefix = this.collectionType.toLowerCase();
 
+      if (prefix === 'commands') {
+        return {
+          prefix: `${prefix}`,
+          detail: `${prefix}.detail`,
+          namespace: `${prefix}.namespace`
+        };
+      }
+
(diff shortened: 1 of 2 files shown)


jithine (approved):

app/components/tc-collection-list/component.js
```diff
@@ -18,6 +18,14 @@ export default Component.extend({
     get() {
       const prefix = this.collectionType.toLowerCase();
 
+      if (prefix === 'commands') {
+        return {
+          prefix: `${prefix}`,
+          detail: `${prefix}.detail`,
```
DekusDenial, line 24: nothing big, just question about the purpose of `detail` and `namespace` fields here and what are they being used and for?

adong: I think
`detail` is to go to the one's specific details (detail route)
`namespace`  is to go back up a level, like directories (namespace route)

DekusDenial: where are they being used? I only see `prefix` being used

PR #1045, opened 2024-04-30
fix(3098): Workflow graph - Non stage jobs overlap into the stage

sagar1312:
## Context

Non stage jobs are being positioned under stage on the workflow graph in the UI giving a false impression that those jobs are part of the stage.

**Example (screwdriver.yaml)**
```
## Stages by grouping
shared:
    image: node:20
    steps:
        - init: echo 'init'
jobs:
    triggering-stage:
        requires: [~pr, ~commit]
    overlapping-non-stage-job-at-start-1:
        requires: [triggering-stage]
    overlapping-non-stage-job-at-start-2a:
        requires: [ overlapping-non-stage-job-at-start-1 ]
    overlapping-non-stage-job-at-start-2b:
        requires: [ overlapping-non-stage-job-at-start-1 ]
    ci-deploy:
        requires: [ ~stage@integration:setup ]
    ci-test-batch-1:
        requires: [ ci-deploy ]
    ci-test-batch-2:
        requires: [ ci-deploy ]
    ci-certify:
        requires: [ ci-test-batch-1, ci-test-batch-2 ]
    overlapping-non-stage-job-at-end-1:
        requires: [ ci-test-batch-1 ]
    overlapping-non-stage-job-at-end-2:
        requires: [ ci-test-batch-2 ]
    triggered-by-stage:
        requires: [ ~stage@integration:teardown ]

stages:
    integration:
        requires: [triggering-stage]
        jobs: [ci-deploy, ci-test-batch-1, ci-test-batch-2, ci-certify]
        description: "This stage will deploy the latest application to CI environment and certifies it after the tests are passed."
```

**Workflow graph (Before)**
image

## Objective

Non stage jobs must be positioned outside stage matrix.

**Workflow graph (After)**
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3098

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1049, opened 2024-05-09
fix(3098): Mini workflow in the event card is not rendered when there are stages

sagar1312:
## Context

Changes made in https://github.com/screwdriver-cd/ui/pull/1045 introduced a regression due to which mini workflow graph are not rendered in the event card as shown below
image

## Objective

Add check to skip the logic to assign position to stage when stages is empty
image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3098

## License


adong:
Can you attach an after screenshot?

adong (approved):

PR #1054, opened 2024-05-14
fix: use standard SCSS

minghay:
## Context
Refactored and cleaned up handlebars and CSS for the the new V2 routes.  By using sass's `@use` and `@include`, we can include styles where we need; begin migration off of the `ember-component-css` library; and properly start using Glimmer components.  One additional benefit is that we can now include styles at the same level as the templates for routes as well.

## Objective
1. Clean up some of the new styles for the V2 routes.
2. Clean up some of the associated templates.
3. Provide component nesting for easier navigation of the codebase.

## License


VonnyJap (approved):

app/v2/pipeline/styles.scss
```diff
@@ -1,36 +1,19 @@
-.new-ui {
-  // to accommodate and offset old ui margins
-  margin-right: -15px;
-  margin-left: -15px;
-  height: 100%;
-  // height: 100vh;
-
-  .grid-container {
+@use 'screwdriver-colors' as colors;
```
adong, line 1: I do like this notion 👍

PR #1066, opened 2024-06-14
fix: extract parameter normalization logic to utility

minghay:
## Context
Parameters require a bit of data massaging to be used in the UI.  With the new UI work, this logic should be extracted to a utility class to be shared across the different UI components.

## Objective
Refactor shared code into a utility for use in the new UI.  Also adds in missing tests around the existing logic.

## License


adong (approved):
Looks good to me, thank you, Ming!

PR #1070, opened 2024-06-19
fix: Converts parameters into a format ready for API use

minghay:
## Context
The API consumes parameters in a concise form.  Through the various layers of UI components, the API responses are massaged into a shape that the UI can use.  In reverse, when the UI needs to send parameters back to the API, the parameters need to get massaged back into a shape that the API expects.  The ownership of this massaging is now part of the `PipelineParametersComponent`.

## Objective
Moves ownership of returning parameters in a form that can be used by the API endpoint to live with the `PipelineParametersComponent`.

## License


adong (approved):
LGTM

PR #1072, opened 2024-06-21
feat: add v2 secrets route

adong:
## Context

Add `secrets` route to v2.
## Objective

Screenshot:
image

## License

### DIFF
--- a/app/v2/pipeline/secrets/controller.js
+++ b/app/v2/pipeline/secrets/controller.js
@@ -1,0 +1,65 @@
+import Controller from '@ember/controller';
+import { reads } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+
+export default Controller.extend({
+  store: service(),
+  session: service(),
+  errorMessage: '',
+  pipeline: reads('model.pipeline'),
+  secrets: reads('model.secrets'),
+  pipelineTokens: reads('model.tokens'),
+  pipelineId: reads('model.pipeline.id'),
+  newToken: null,
+  refreshService: service('pipeline.secrets'),
+  actions: {
+    createSecret(name, value, pipelineId, allowInPR) {
+      const newSecret = this.store.createRecord('secret', {
+        name,
+        value,
+        pipelineId,
+        allowInPR
+      });
+
+      return newSecret.save().then(
+        s => {
+          this.set('errorMessage', '');
+          this.secrets.reload();
+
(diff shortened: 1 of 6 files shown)


minghay:
Can you try using Glimmer components?  I think it's best to take the opportunity now with the new UI work to migrate over to Glimmer components.

app/v2/pipeline/secrets/route.js
```diff
@@ -1,3 +1,43 @@
+// import Route from '@ember/routing/route';
+
+// export default class NewPipelineSecretsRoute extends Route {}
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+import { get } from '@ember/object';
+
+export default Route.extend({
```
minghay, line 8: You can use a native class here

app/v2/pipeline/secrets/controller.js
```diff
@@ -0,0 +1,65 @@
+import Controller from '@ember/controller';
+import { reads } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+
+export default Controller.extend({
```
minghay, line 5: You should be able to use a native class for the controller

app/v2/pipeline/template.hbs
```diff
@@ -1,7 +1,8 @@
 {{page-title "Pipeline"}}
 <div class="pipeline-content">
   <Pipeline::Nav />
-  <div class="pipeline-main-content">
+  <div class="pipeline-main-content {{if this.isSecretsRoute "unset-grid-template-columns"}}"
```
minghay, line 4: This is going to be problematic for a few other pipeline routes.  You should either add a class for the grid layout for the ones around events or refactor the CSS so that it's only applied at a lower level element

adong: Very valid point, I'm making changes, please take a look.

app/v2/pipeline/secrets/controller.js
```diff
@@ -0,0 +1,87 @@
...
+  @service store;
+
+  @service session;
+
+  @service('pipeline.secrets')
+  refreshService;
+
+  @tracked newToken = '';
+
+  @tracked newToken = null;
```
minghay, line 16:
```suggestion
  @tracked newToken = null;
```

minghay (approved):

PR #1068, opened 2024-06-18
feat: New pipeline parameters component

minghay:
## Context
#1065 added a new Glimmer component that wraps Ember Power Select for the use with pipeline/job parameters.  This new Glimmer component wraps up all pipeline/job parameters into a single entity that has some styles applied and some basic user interactions.  Conceptually, users will interact with this component and modify parameters and any changes to these parameters are made available to a parent component via the `@onUpdateParameters` argument.

## Objective
Provide a new Glimmer component that is compatible with the new UI.  There are some minor CSS changes to what exists in the old UI, but it's largely the same.
Screenshot 2024-06-17 at 19-01-00 New Pipeline Events Show

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1075, opened 2024-06-27
feat: New confirm action modal UI

minghay:
## Context
New UI for the modal to confirm start/restarting of a job from the workflow graph.

## Objective
As part of the new UI work, this provides a new modal that is a Glimmer component and leverages the pipeline parameters component from #1068.

The new component also properly displays the desired action that is to be taken with the job (e.g., start or restarting a job)
Screenshot 2024-06-27 at 14-52-08 New Pipeline Events Show

Screenshot 2024-06-27 at 14-52-21 New Pipeline Events Show

Screenshot 2024-06-27 at 14-52-30 New Pipeline Events Show

## License

### DIFF
--- a/app/components/pipeline/modal/confirm-action/util.js
+++ b/app/components/pipeline/modal/confirm-action/util.js
@@ -1,0 +1,80 @@
+import { flattenParameters } from 'screwdriver-ui/utils/pipeline/parameters';
+
+/**
+ * Initialize parameters from an event API response object
+ * @param event
+ * @returns {undefined|*}
+ */
+export function initializeParameters(event) {
+  if (event.meta?.parameters && event.meta.parameters.length > 0) {
+    return flattenParameters(event.meta.parameters);
+  }
+
+  return undefined;
+}
+
+/**
+ * Truncate commit message to 150 characters
+ * @param commitMessage
+ * @returns {string|*}
+ */
+export function truncateMessage(commitMessage) {
+  const cutOff = 150;
+
+  return commitMessage.length > cutOff
+    ? `${commitMessage.substring(0, cutOff)}...`
+    : commitMessage;
+}
+
+/**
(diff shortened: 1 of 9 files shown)


adong (approved):

app/components/pipeline/modal/confirm-action/util.js
```diff
@@ -0,0 +1,80 @@
...
+    : commitMessage;
+}
+
+/**
+ * Capitalize first letter of string
+ * @param string
+ * @returns {string}
+ */
+export function capitalizeFirstLetter(string) {
+  return string.charAt(0).toUpperCase() + string.slice(1);
```
adong, line 35: can we have a default `string = ''` ?

app/components/pipeline/modal/confirm-action/util.js
```diff
@@ -0,0 +1,80 @@
...
+ * Initialize parameters from an event API response object
+ * @param event
+ * @returns {undefined|*}
+ */
+export function initializeParameters(event) {
+  if (event.meta?.parameters && event.meta.parameters.length > 0) {
+    return flattenParameters(event.meta.parameters);
+  }
+
+  return undefined;
```
adong, line 13: It intentionally returns `undefined`, rather than `{}`. Either one should be fine, and I'm curious to know the reason for that.

app/components/pipeline/modal/confirm-action/util.js
minghay: It already returns `''` if the input string is empty.

app/components/pipeline/modal/confirm-action/util.js
minghay: Mainly to keep things in line with the original API response.  The `meta` field returned on the event from the API would resolve to `undefined` if there are no parameters set.

PR #1108, opened 2024-08-03
fix: moved default workflow graph for pipeline template

VonnyJap:
## Context

The workflow graph value init [here](https://github.com/screwdriver-cd/ui/blob/6202078d8216a6c89dd8abd8f5377fe993f9a6e3/app/templates/pipeline/detail/controller.js#L23-L26) overwrites the value from [the config of selected version](https://github.com/screwdriver-cd/ui/blob/6202078d8216a6c89dd8abd8f5377fe993f9a6e3/app/templates/pipeline/detail/controller.js#L74-L78) 

## Objective

The default value of workflow graph which is 
```
  workflowGraph = {
    nodes: [],
    edges: []
  };
 ```
 should be used only when the selected version is missing the workflow graph config. 
 
## License

### DIFF
--- a/app/templates/pipeline/detail/controller.js
+++ b/app/templates/pipeline/detail/controller.js
@@ -20,11 +20,6 @@ export default class TemplatesPipelineDetailController extends Controller {
 
   pipelineTemplateTags = [];
 
-  workflowGraph = {
-    nodes: [],
-    edges: []
-  };
-
   @tracked startTime;
 
   @tracked endTime;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/templates/pipeline/detail/controller.js [hunk 1]
```diff
@@ -72,9 +67,15 @@ export default class TemplatesPipelineDetailController extends Controller {
   }
 
   get workflowGraph() {
+    console.log('selectedVersionTemplate', this.selectedVersionTemplate);
```
tkyi, line 70: Remove log

minghay (approved):

PR #1109, opened 2024-08-05
fix: Util to check if an event was skipped

minghay:
## Context
Creates a new utility function that checks whether an event has was skipped due to having the `skip ci` or `ci skip` text in the commit.

## Objective
The new UI needs to determine if an event was skipped and this utility function extracts that logic out so that it can be used by the new workflow graph.

## License

### DIFF
--- a/app/utils/pipeline/event.js
+++ b/app/utils/pipeline/event.js
@@ -1,0 +1,14 @@
+/**
+ * Determines if the event has been skipped
+ * @param event
+ * @param builds
+ * @returns {boolean}
+ */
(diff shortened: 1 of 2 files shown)


VonnyJap (approved):

app/utils/pipeline/event.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Determines if the event has been skipped
+ * @param event
+ * @param builds
```
tkyi, line 4: missing jsdocs

PR #1110, opened 2024-08-05
feat: New workflow graph component

minghay:
## Context
New glimmer component for the V2 workflow graph

## Objective
Creates a new glimmer component and associated stylesheet for the workflow graph.  The primary difference between the new component and the existing component is that any logic around creating and massaging data related to the workflow graph is moved up a layer and the new component is effectively a dumb display layer.

## License

### DIFF
--- a/app/components/pipeline/workflow/graph/component.js
+++ b/app/components/pipeline/workflow/graph/component.js
@@ -1,0 +1,95 @@
+/* global d3 */
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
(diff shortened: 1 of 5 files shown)


VonnyJap:
Any screenshot to show in this PR @minghay?

VonnyJap (commented):
LGTM, but which layer will the data be since this has become just a component for rendering?

minghay:
The routes provide the models and there is a controller that will help out with providing the proper parameters into the component.

VonnyJap (approved):

PR #1136, opened 2024-08-29
feat: New event card for v2 UI

minghay:
## Context
Adding in a new component for event cards in the v2 UI.  The old ones were built using ember data, and are incompatible with how the v2 UI operates.

## Objective
Provide a new component for the event cards for the v2 UI.  The new cards follow the general pattern of using modals for actions that require user confirmations.

Basic event card:
Screenshot 2024-08-29 at 11-33-25 New Pipeline Events Show

Event card with background color used for hover states (also note that the commit hash is highlighted indicating it is the latest commit):
Screenshot 2024-08-29 at 11-35-28 New Pipeline Events Show

Event card with background highlight indicating the event is the currently selected:
Screenshot 2024-08-29 at 11-35-03 New Pipeline Events Show

Event card showing the placement of the lab
Screenshot 2024-08-29 at 11-36-39 New Pipeline Events Show
el metadata:

Event card showing the placement of the configured parameters of an event:
Screenshot 2024-08-29 at 12-01-45 New Pipeline Events Show

Event card showing the abort button for an event that is still running:
Screenshot 2024-08-29 at 12-03-19 New Pipeline Events Show

## License

### DIFF
--- a/app/components/pipeline/event/card/component.js
+++ b/app/components/pipeline/event/card/component.js
@@ -1,0 +1,115 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+import { statusIcon } from 'screwdriver-ui/utils/build';
+import { getStatus } from 'screwdriver-ui/utils/pipeline/event';
+import {
+  getDurationText,
+  getFailureCount,
+  getExternalPipelineId,
+  getStartDate,
+  getTruncatedMessage,
+  getTruncatedSha,
+  getWarningCount,
+  isCommitterDifferent,
+  isExternalTrigger
+} from './util';
+
+export default class PipelineEventCardComponent extends Component {
+  @service router;
+
+  @service shuttle;
+
+  @tracked status;
+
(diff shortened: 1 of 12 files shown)


tkyi:
Any reason why the `Stop Builds` button doesn't have a white background like the `Parameters` button?
Also is it confusing that the Parameters button looks so similar to the Start/Stop buttons since it's not actually an action?

minghay:
I missed updating the style for the `Stop Builds` button to have the white background, will get that updated.

The `Parameters` button is also an action in so far as it opens up a modal that shows all of the parameters.  That's to be consistent with how the parameters are all done in the UI (opened up in a modal).

tkyi:
Ah I guess what I mean by "action" is that Start/Stop will actually make API calls and affect the build but Parameters is just showing information

minghay:
That's true, there's a small odd overlap between the Start/Stop actions using a modal and the parameters also using a modal on the event card.  I currently don't have any better ideas on how to present the parameters, so going with how the other parameters is typically presented seemed to be the best option for now.

tkyi (approved):

PR #1159, opened 2024-09-18
fix: Update start modal to allow for optional messaging

minghay:
## Context
The start event modal is used in multiple places.  There may be some additional messaging that needs to happen to better alert the user to pay attention to certain specifics of their pipeline configuration.

## Objective
Update the start event modal to allow for optional messaging.
Optional notice

## License


tkyi:
How is this optional messaging set?

minghay:
The component takes the `notice` parameter (which is assumed to be a string value).  The component just dumb displays the string value in the warning.

tkyi (approved):

PR #1169, opened 2024-09-18
feat: Table cell for actions

minghay:
## Context
The new jobs UI makes use of custom components for each cell in the table. This creates a new component for handling the actions that a user can take for a job.

## Objective
Create a glimmer component for the jobs. Screenshot will be captured in the linked PR

## References
#1171 for screenshot 
https://github.com/screwdriver-cd/screwdriver/issues/3200

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1161, opened 2024-09-18
feat: Table cell component for jobs

minghay:
## Context
The new jobs UI makes use of custom components for each cell in the table.  This creates a new component for handling the job name and the build status.

## Objective
Create a glimmer component for the jobs.  Screenshot will be captured in the linked PR

## References
#1171 for screenshot 
https://github.com/screwdriver-cd/screwdriver/issues/3200

## License

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1171, opened 2024-09-18
feat: v2 jobs table glimmer component

minghay:
## Context
The new v2 UI for the jobs route.

## Objective
Creates the new v2 UI glimmer component which is a customized table using `ember-models-table` as the base.  This table also incorporates automatic data refreshing to keep all the table data up to date.
Screenshot 2024-09-18 at 14-15-53 New Pipeline Jobs

## References
Incorporates the glimmer components for each cell:
#1161 for the cells in the `JOB` column
#1162 for the cells in the `HISTORY` column 
#1163 for the cells in the `DURATION` column
#1164 for the cells in the `START TIME` column
#1165 for the cells in the `COVERAGE` column
#1166 for the cells in the `METRICS` column
#1169 for the cells in the `ACTIONS` column

The background refresh logic is from #1170

## License

### DIFF
--- a/app/components/pipeline/jobs/styles.scss
+++ b/app/components/pipeline/jobs/styles.scss
@@ -1,0 +1,5 @@
+@use 'table/styles' as table;
+
+@mixin styles {
+  @include table.styles;
+}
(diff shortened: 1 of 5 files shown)


tkyi:
Small comment: doesn't have to be in this PR, but would be nice if there was some way to indicate which builds are more recent (under HISTORY). We could maybe have an opacity gradient?

tkyi (approved):

PR #1195, opened 2024-10-16
fix(1749): Add ability to download artifact directory

tkyi:
## Context

Should be able to download artifacts by directory.

## Objective

This PR adds ability to download artifact directory.

## References

Related to https://github.com/screwdriver-cd/screwdriver/issues/1749

## License


app/components/artifact-tree/component.js
```diff
@@ -88,6 +88,10 @@ export default Component.extend({
         const artifactPath = instance.get_path(node, '/');
 
         if (type === 'directory') {
+          this.setProperties({
+            href,
+            type: 'directory'
+          });
```
minghay, line 94:
```suggestion
```

app/components/artifact-tree/component.js
```diff
@@ -101,7 +105,8 @@ export default Component.extend({
           this.setProperties({
             href,
             iframeUrl: `${href}?type=preview`,
-            isModalOpen: true
+            isModalOpen: true,
+            type: 'file'
```
minghay, line 109:
```suggestion
```

app/components/artifact-preview/component.js
```diff
@@ -41,11 +41,17 @@ export default Component.extend({
 
   actions: {
     download() {
-      const downloadLink = this.iframeUrl.replace(
+      let downloadLink = this.iframeUrl.replace(
```
minghay, line 44: Don't use the `iframeUrl` as it doesn't get set properly (I'm not currently sure why the tree structure isn't setting it properly for directories).

Inject the Ember router service into this component and you can use the `.currentRoute` property to extract the artifact path within it:
```javascript
  this.router.currentRoute.params.file_path
```

The artifact that is returned as a string will have trailing `/` for directories, so you can use that to do the rest of the needed work of recreating the proper URL.

One last thing, you'll need to recreate the base API URL like what is down in the `downloadAll` function just below this function

minghay (changes requested):

app/components/artifact-preview/component.js
tkyi: updated

app/components/artifact-preview/component.js
```diff
@@ -41,11 +42,20 @@ export default Component.extend({
...
         'type=preview',
         'type=download'
       );
 
+      if (this.type === 'directory') {
```
minghay, line 50: `this.type` is always undefined.  The type was set in code that was used in a previous library.

minghay (approved):

PR #1200, opened 2024-10-21
fix: Event card update

minghay:
## Context
Multiple event cards can be displayed in the UI at any given point so there are a couple of elements in the card that should not use ids for styling.  Additionally, the button to display the parameters configured for the event do not need to be available for all instances of the event cards.

## Objective
1. Update the event card styling to better address variable content sizes
2. Moves event timing information to a new location in the card for improved legibility
3. Changes the buttons in the event card to not use text to help save space and allow for more functionality within the card component

Updated layout showing the new Objectives 2 and 3:
Screenshot 2024-10-21 at 13-15-12 New Pipeline Events

## References

## License

### DIFF
--- a/app/components/pipeline/event/card/styles.scss
+++ b/app/components/pipeline/event/card/styles.scss
@@ -85,8 +85,6 @@
 
       .abort-event-button {
         margin-left: auto;
-        margin-top: auto;
-        margin-bottom: auto;
         background-color: colors.$sd-white;
       }
     }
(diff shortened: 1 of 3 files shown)


tkyi:
Nitpick: 
- do we need the stop icon in a circle and the button? or would it be okay to still use text for Start/Stop/Restart since we seem to have enough space there
- the icons don't look centered in the buttons (they look a little too high?)

minghay:
> * do we need the stop icon in a circle and the button? or would it be okay to still use text for Start/Stop/Restart since we seem to have enough space there
Keeping with the iconography instead of text keeps the design consistent with the other buttons within the card. 

> * the icons don't look centered in the buttons (they look a little too high?)
Fixed the styles so that the icons are now centered
Screenshot 2024-10-24 at 11-57-33 New Pipeline Events

tkyi (approved):

PR #1205, opened 2024-10-25
feat: Event group modal

minghay:
## Context
Users should be able to view all events that make up a group.  In the V1 interface, there is an expandable option for displaying events, however the drawback of the V1 interface is that older events in the group are completely removed from the view and in certain aspects makes it difficult to really understand the order of events that have happened for the pipeline.

The V2 interface will instead keep all events in the event rail as they occur and have a button in within the event card that will open up a modal to display all events that make up an event group.  This will hopefully make it easier for users to navigate the full event history of a pipeline while providing the additional grouping information if they want to dive into the event history of an event group. 

## Objective
Create a modal to display all events that make up an event group.  Also adds the button to display the event group modal into the event card itself.

Event group modal will display the event group in the modal title and the events that make up the event group:
Screenshot 2024-10-21 at 12-27-17 New Pipeline Events

The modal will also overflow with a vertical scroll:
Screenshot 2024-10-21 at 12-27-53 New Pipeline Events

## References

## License

### DIFF
--- a/app/components/pipeline/event/card/component.js
+++ b/app/components/pipeline/event/card/component.js
@@ -47,6 +47,8 @@ export default class PipelineEventCardComponent extends Component {
 
   @tracked showAbortBuildModal;
 
+  @tracked showEventHistoryModal;
+
   queueName;
 
   userSettings;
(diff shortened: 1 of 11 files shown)


minghay:
> * would it be better to have separating lines between the events?
This might be beneficial to have, I'll play around with the idea when all of the event card lists are in and see what works well across all of those instances.

> * will there be any actionable buttons from the modal?
At the moment, only the abort event button would be actionable.  I will probably disable all actionable buttons from this view and only allow users to navigate to a new event.

> * is there any other way to indicate which events are more recent vs having to look at the time started
The events are all sorted by date so there shouldn't be any other particular need at the moment to add additional indicators of event order.

> * how useful/meaningful is it for users to see the event group number?
The group number isn't critical to be displayed within the modal as it's just additional information.  It may be helpful for users who aren't super familiar with the fact that Screwdriver has event grouping as the group number will provide some data that they can reference when exploring their pipeline.

tkyi (approved):

PR #1209, opened 2024-10-29
fix: Revert "build(deps): bump cookie, express and socket.io"

tkyi:
Reverts screwdriver-cd/ui#1186

The PR pages are missing everything under the build banner and you cannot navigate away. Trying to see if reverting this change fixes it.

Screenshot 2024-10-29 at 2 40 01 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11188,9 +11188,9 @@
       "dev": true
     },
     "node_modules/cookie": {
-      "version": "0.7.1",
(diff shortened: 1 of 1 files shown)


minghay (approved):

kumada626:
@tkyi I think that missing PR build page bug is not caused by those dependency update but caused by below PR.
https://github.com/screwdriver-cd/ui/pull/1201
Console shows `TypeError: Cannot read properties of undefined (reading 'virtual')` error. And virtual property is introduced in that PR.

PR #1210, opened 2024-10-30
feat: Modal to search for an event

minghay:
## Context
The v2 UI adds in a feature to search for an event by SHA.  To support the search feature, a new modal will be employed to provide the text input field and display of the results.  This modal will do dynamic searches so as valid characters are input into the text input field the results will be updated.  For invalid inputs, the text input field will change colors to provide immediate feedback.

The escape key is configured to allow clearing of the input field when an input has been provided.  If the input field is empty, the escape key will close the modal (which is the default behavior for all of the modals being used in the v2 UI).

## Objective
Create a new modal to support event searching.

Basic modal that is presented to the user:
Screenshot 2024-10-29 at 17-54-03 New Pipeline Events

Valid input with search results:
Screenshot 2024-10-29 at 18-17-06 New Pipeline Events

Invalid input showing text color change and the search results not changing due to the invalid input:
Screenshot 2024-10-29 at 18-17-33 New Pipeline Events

## References

## License

### DIFF
--- a/app/components/pipeline/modal/search-event/component.js
+++ b/app/components/pipeline/modal/search-event/component.js
@@ -1,0 +1,57 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+
+export default class PipelineModalSearchEventComponent extends Component {
+  @service shuttle;
+
+  @tracked sha;
+
+  @tracked searchResults;
+
+  @tracked inputClass;
+
+  constructor() {
+    super(...arguments);
+
+    this.searchResults = [];
+    this.inputClass = null;
+  }
+
+  @action
+  handleInput(inputEvent) {
+    const inputValue = inputEvent.target.value;
+
+    if (inputValue && inputValue.length > 0) {
+      if (!/^[0-9a-f]{1,40}$/.test(inputValue)) {
+        if (!this.inputClass) {
+          this.inputClass = 'invalid';
+        }
+      } else {
+        this.inputClass = null;
+        this.shuttle
+          .fetchFromApi(
+            'get',
+            `/pipelines/${this.args.pipeline.id}/events?sha=${inputValue}`
+          )
+          .then(events => {
+            this.searchResults = events;
+          });
+      }
+    } else {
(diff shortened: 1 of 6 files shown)


app/components/pipeline/modal/search-event/component.js
```diff
@@ -0,0 +1,57 @@
...
+    this.searchResults = [];
+    this.inputClass = null;
+  }
+
+  @action
+  handleInput(inputEvent) {
+    const inputValue = inputEvent.target.value;
+
+    if (inputValue && inputValue.length > 0) {
+      if (!/^[0-9a-f]{1,40}$/.test(inputValue)) {
```
tkyi, line 27: can we save this regex as a constant and/or somehow use the data-schema's definition? (https://github.com/screwdriver-cd/data-schema/blob/master/models/event.js#L41-L45)

tkyi:
Some questions:
- Do we have plans to add other searchable parameters in the future (status, committer, date, etc)?
- For invalid input case: currently if the Checkout URL is formatted incorrectly we highlight the outline of the form red, should we keep that behavior consistent? And/or alternatively should we have small text in red underneath with words as well, something like "Invalid format for SHA, must be hex (e.g. a0d4389)"

app/components/pipeline/modal/search-event/component.js
minghay: I'll move it into a constant.  It's just a basic test for hex values anyway, so there's no particular need to pull it from the data-schema definition.  Additionally, since it's used as a partial match rather than an exact match, we want the regex to not be fixed to 40 as in the data-schema.

minghay:
> * Do we have plans to add other searchable parameters in the future (status, committer, date, etc)?

No plans for other searchable parameters on the horizon yet.

> * For invalid input case: currently if the Checkout URL is formatted incorrectly we highlight the outline of the form red, should we keep that behavior consistent? And/or alternatively should we have small text in red underneath with words as well, something like "Invalid format for SHA, must be hex (e.g. a0d4389)"

How about starting with this?
Screenshot 2024-10-30 at 11-37-18 New Pipeline Events

tkyi (approved):

PR #1215, opened 2024-11-01
fix(3227): Show download button when the flag is false

kumada626:
## Context

#1206 has a minor bug with the if-else conditions in the template which causes the download button to go missing when feature flag is false.

## Objective

Fixes the missing download button when feature flag is false.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3227

## License


yk634 (approved):

tkyi (approved):
🙇‍♀️ thank you sorry

PR #1230, opened 2024-11-10
feat(3197): Table cell component for stages in pipeline list jobs view

tkyi:
## Context
The new jobs UI makes use of custom components for each cell in the table.  This creates a new component for handling the stage name for the job.

## Objective
Create a glimmer component for the jobs.
V1:
Screenshot 2024-11-09 at 10 03 13 PM

V2:
Screenshot 2024-11-09 at 10 05 27 PM

## References
https://github.com/screwdriver-cd/screwdriver/issues/3197

## License


app/components/pipeline-list-stage-cell/component.js
```diff
@@ -0,0 +1,10 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+
+export default Component.extend({
+  stageName: computed('record.stage.stageName', {
```
VonnyJap, line 5: Locally I got an error when viewing something without a stage. 
Screenshot 2024-11-12 at 4 47 49 PM

VonnyJap: Config for reference https://github.com/screwdriver-cd-test/functional-commands/blob/master/screwdriver.yaml

VonnyJap: After some investigation, this issue is caused when a pipeline has jobs that never run before, the change to the jobs list view to include the stages is not able to handle this at the moment. 

To reproduce: create any new pipeline and don't run any builds, check into the jobs list view and console.

VonnyJap (approved):

app/components/pipeline-list-stage-cell/component.js
tkyi: Strangely this bug doesn't show up the first time the jobs list view loads, only after refreshing. Fixed it by putting a default value `N/A`

PR #1234, opened 2024-11-15
feat(3241): Add options to select historical number of builds in jobs list view

VonnyJap:
## Context

The job list view does not have any form of dynamic options to allow number of historical builds and sometimes it might be necessary to get higher number of history. 

## Objective

Add a select options to choose number of historical builds to be displayed in jobs list view. 
Screenshot 2024-11-13 at 4 00 55 PM

## References

https://github.com/screwdriver-cd/screwdriver/issues/3241

## License

### DIFF
--- a/app/components/pipeline-list-view/component.js
+++ b/app/components/pipeline-list-view/component.js
@@ -313,8 +313,19 @@ export default Component.extend({
     'jobsDetails.[]',
     function jobsObserverFunc({ jobsDetails }) {
       const rows = this.getRows(jobsDetails);
+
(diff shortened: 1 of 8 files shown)


jithine:
can you share screenshots on how this would look with 30 builds ? Are these options sticky ? Can they also be treated like a user preference for the pipeline ?

VonnyJap:
> can you share screenshots on how this would look with 30 builds ? Are these options sticky ? Can they also be treated like a user preference for the pipeline ?

@jithine - it is not sticky, this is filter and not user settings.

PR #1266, opened 2024-11-27
fix: Disable job restart when job has no builds

minghay:
## Context
When selecting a job from the pipeline workflow graph in the V2 UI, jobs that have no builds for that event should not be allowed to restart.

## Objective
Updates the logic for determining whether a job can be restarted from the pipeline workflow graph.

## References

## License


tkyi:
Hm is that consistent with current behavior?

PR #1275, opened 2024-12-03
fix(3247): virtual job icon is not used when the graph redraw

y-oksaku:
## Context

Virtual job icon is not used after the workflow graph is redrew. (e.g. Uses stops the build)

## Objective

Use virtual job icon correctly when the workflow graph is redrew.

## References

Issue: https://github.com/screwdriver-cd/screwdriver/issues/3247

## License

### DIFF
--- a/app/utils/pipeline/graph/d3-graph-util.js
+++ b/app/utils/pipeline/graph/d3-graph-util.js
@@ -28,7 +28,7 @@ const STATUS_MAP = {
  * @return {String}        Unicode character that maps to an icon in screwdriver icon font
  */
 export function icon(status, isVirtual) {
-  if (isVirtual && (!status || ['SUCCESS', 'WARNING', 'CREATED'])) {
+  if (isVirtual) {
     return STATUS_MAP.VIRTUAL.icon;
   }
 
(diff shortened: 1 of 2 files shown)


app/utils/pipeline/graph/d3-graph-util.js
```diff
@@ -28,7 +28,7 @@ const STATUS_MAP = {
  * @return {String}        Unicode character that maps to an icon in screwdriver icon font
  */
 export function icon(status, isVirtual) {
-  if (isVirtual && (!status || ['SUCCESS', 'WARNING', 'CREATED'])) {
```
y-oksaku, line 31: Right hand side of this statement is always `true`.

I think that this might means `['SUCCESS', 'WARNING', 'CREATED'].includes(status)`.
But if use this statement then the icon will be able to change `CREATED/QUEUED` to `VIRTUAL`.

In this case, we need to recalculate the icon size and its position when redrawing. This would be difficult.

This change explicitly allowed the virtual icon to have created/queued status (i.e. allow blue virtual icon).

yakanechi (approved):

tkyi (approved):

PR #1282, opened 2024-12-04
fix: Fix job name search

minghay:
## Context
When searching for a job in the v2 jobs table, jobs that have a display name annotation configured are not able to be search for properly

## Objective
Fixes the job name so that it is searchable.

## References

## License


tkyi:
With this change, if displayName annotation is set can you not search by original job name?

minghay:
> With this change, if displayName annotation is set can you not search by original job name?

That would be the case for now, yes.  I need to spend a little bit more time to see if I can get the search to work for both the original job name and the display name.  As it currently stands, the display name, if configured, would be the one that is expected to be found as the original name requires a mouse-over to display

tkyi:
I'm not sure what is expected behavior from the user. But I guess for now it makes sense that you can search by what you can see

tkyi (approved):

PR #1301, opened 2024-12-16
fix(3253): allow time range picker to render based on correct input selection

VonnyJap:
## Context

Currently, when users select a time range in the job template view, the dropdown always defaults to "1 yr" instead of retaining the selected time range. This behavior results in a poor user experience, as the time range is not properly applied or displayed according to user input.

## Objective

The goal of this PR is to fix the time range selection behavior in the job template view. The time range dropdown should accurately display and apply the user's selected value, ensuring that the selection is preserved and reflected correctly in the UI.

## References

https://github.com/screwdriver-cd/screwdriver/issues/3253

## License


minghay (changes requested):

VonnyJap:
I encountered issues with the commit history, so I’m closing this PR to address them.

PR #1304, opened 2024-12-16
fix(3253): allow time range picker to render based on correct input selection

VonnyJap:
## Context

Currently, when users select a time range in the job template view, the dropdown always defaults to "1 yr" instead of retaining the selected time range. This behavior results in a poor user experience, as the time range is not properly applied or displayed according to user input.

## Objective

The goal of this PR is to fix the time range selection behavior in the job template view. The time range dropdown should accurately display and apply the user's selected value, ensuring that the selection is preserved and reflected correctly in the UI.

https://github.com/user-attachments/assets/2a0e813d-c986-4ed5-8daa-a1250366816f

## References

https://github.com/screwdriver-cd/screwdriver/issues/3253

## License


app/components/time-range-picker/component.js
```diff
@@ -46,13 +61,13 @@ export default Component.extend({
...
 
-        this.onTimeRangeChange(startTime, endTime);
+        this.onTimeRangeChange(startTime, endTime, this.selectedRange);
       } else {
         this.onTimeRangeChange();
       }
     },
     setCustomRange([start, end]) {
-      this.set('selectedRange');
+      this.set('selectedRange', 'none');
```
minghay, line 70: `setCustomRange` should account for a non-date range selection by not changing the `selectedRange` instead of setting the value to `none`.  By doing that, closing the date picker will not force an unnecessary reload of the page.

app/components/time-range-picker/component.js
```diff
@@ -46,13 +61,13 @@ export default Component.extend({
       if (range) {
         const { startTime, endTime } = timeRange(new Date(), range);
 
-        this.onTimeRangeChange(startTime, endTime);
+        this.onTimeRangeChange(startTime, endTime, this.selectedRange);
       } else {
         this.onTimeRangeChange();
       }
     },
```
minghay, line 68: `setTimeRange` can be simplified because the template has the button group configured as radio buttons which will prevent the `setTimeRange` from executing unless a different radio button selection is selected.

minghay (changes requested):

minghay (dismissed):

app/components/template-versions/template.hbs
```diff
@@ -4,9 +4,13 @@
 
 {{#unless this.isPipelineTemplatePage}}
 <div class="table-controls">
+  <span class="note">
```
minghay, line 7:
```suggestion
  <span>
```

app/components/template-versions/styles.scss
```diff
@@ -17,7 +17,16 @@
 
   .table-controls {
     display: flex;
-    justify-content: flex-end;
+    justify-content: space-between; /* Space between the note and the picker */
+    align-items: center; /* Keep vertical alignment centered */
+    flex-wrap: wrap;
```
minghay, line 22:
```suggestion
    flex-direction: column;
```

app/components/template-versions/styles.scss
```diff
@@ -17,7 +17,16 @@
...
     margin-bottom: 10px;
+    max-width: 100vw;
+    overflow: auto;
+  }
```
minghay, line 26:
```suggestion
```

app/components/template-versions/styles.scss
```diff
@@ -17,7 +17,16 @@
...
+
+  .table-controls .note {
+    margin-right: auto; /* Push the TimeRangePicker to the end */
+    align-self: center; /* Ensure the note aligns vertically */
```
minghay, line 30:
```suggestion
  > * {
    margin-left: auto;
  }
```

minghay (commented):
The blue outline is getting cut off a little bit.  I've suggested a couple of CSS changes that will fix the outline issue and also stacks the note above the date picker

minghay (approved):

PR #1316, opened 2024-12-20
feat(3200): css fix to give round border of stages and right padding in WG container

VonnyJap:
## Context

The rectangle border around a stage in the new interface lacks a border radius, deviating from the current design. Additionally, the workflow graph container’s right side has no padding, making it visually unappealing.

Screenshot 2024-12-19 at 4 32 48 PM

## Objective

Add border radius to the stage rectangles in the new interface and apply padding to the right side of the workflow graph container for improved visual consistency.

Screenshot 2024-12-19 at 4 43 34 PM

## References

## License


app/components/pipeline/workflow/graph/styles.scss
```diff
@@ -99,5 +99,10 @@
         margin-top: 5px;
       }
     }
+
+    .stage-container {
+      rx: 5;
+      ry: 5;
+    }
```
minghay, line 106: `rx` and `ry` aren't valid CSS properties, they are for SVG elements.  The proper fix is to actually remove the CSS here and in `workflow-graph-d3/styles.scss` and then modify `utils/pipeline/graph/d3-graph-util.js` so that the `rx` and `ry` attributes are set on the SVG `rect` element.

app/components/pipeline/workflow/styles.scss
```diff
@@ -26,6 +26,7 @@
     #workflow-display-container {
       grid-area: workflow;
       overflow: auto;
+      padding-right: max(1rem, 0.8vw);
```
minghay, line 29:
```suggestion
      padding-right: 1rem;
```
You'll want to keep this as 1rem to keep things lined up along the right side of the window

minghay (changes requested):

minghay (approved):

PR #1322, opened 2025-01-22
feat(3241): add more options to build historical view

VonnyJap:
## Context

As of now users can only see up to 30 historical builds in the jobs list view. 

## Objective

This change is to add more options to allow users to view up to 50 historical builds. 

## References

https://github.com/screwdriver-cd/screwdriver/issues/3241

## License

### DIFF
--- a/app/components/pipeline-list-view/component.js
+++ b/app/components/pipeline-list-view/component.js
@@ -23,7 +23,7 @@ export default Component.extend({
   lastRows: [],
   moreJobs: true,
   timestampPreference: null,
-  buildsHistoryOptions: [5, 10, 15, 20, 25, 30],
+  buildsHistoryOptions: [5, 10, 15, 20, 25, 30, 35, 40, 45, 50],
   columns: [
     {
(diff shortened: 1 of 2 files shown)


app/components/pipeline-list-view/component.js
```diff
@@ -23,7 +23,7 @@ export default Component.extend({
   lastRows: [],
   moreJobs: true,
   timestampPreference: null,
-  buildsHistoryOptions: [5, 10, 15, 20, 25, 30],
+  buildsHistoryOptions: [5, 10, 15, 20, 25, 30, 35, 40, 45, 50],
```
jithine, line 26: why 35, 50, 45 ? At higher limits, just jump by 20/10 at least.

minghay:
I think the table should be changed to not have predefined history sizes in a dropdown at this point.  We should change it so that it is an input field for users to set to whatever value they desire.

minghay:
Closed until further discussions/plans for addressing the historical builds is finalized.

PR #1331, opened 2025-02-10
feat: Inject pipeline

minghay:
## Context
The pipeline page has a lot of static data that is being passed through many layers of glimmer components.  This static data can be injected into the components as necessary.  For this initial change, only the pipeline data will be set up, but future changes will start incorporating additional pieces of data.

## Objective
Sets up an injectable ember service for capturing static data starting with the pipeline data. 

## References

## License

### DIFF
--- a/tests/unit/pipeline-page-state/service-test.js
+++ b/tests/unit/pipeline-page-state/service-test.js
@@ -1,0 +1,13 @@
+import { module, test } from 'qunit';
+import { setupTest } from 'screwdriver-ui/tests/helpers';
+
+module('Unit | Service | pipelinePageState', function (hooks) {
+  setupTest(hooks);
+
+  // TODO: Replace this with your real tests.
+  test('it exists', function (assert) {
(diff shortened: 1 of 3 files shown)


tests/unit/pipeline-page-state/service-test.js
```diff
@@ -0,0 +1,13 @@
...
+
+module('Unit | Service | pipelinePageState', function (hooks) {
+  setupTest(hooks);
+
+  // TODO: Replace this with your real tests.
+  test('it exists', function (assert) {
+    let service = this.owner.lookup('service:pipeline-page-state');
+
+    assert.ok(service);
+  });
```
VonnyJap, line 12: planning to do it later?

VonnyJap (approved):

tests/unit/pipeline-page-state/service-test.js
minghay: Yes, additional tests will be filled in later when it expands to have more complex logic.

PR #1342, opened 2025-02-18
feat(3065): Ability to collapse and expand stage in the UI

sagar1312:
## Context

Right now stage visualization default behavior is expanded.

image

## Objective

Should have a way to collapse and expand a stage so it can minimize the real estate take in workflow graph.

**When Expanded**
image

**When Collapsed**
image

## License

### DIFF
--- a/app/components/pipeline-workflow/component.js
+++ b/app/components/pipeline-workflow/component.js
@@ -52,6 +52,8 @@ export default Component.extend({
     }
   ),
 
+  collapsedStages: new Set([]),
+
   canRestartPipeline: computed('authenticated', 'pipeline.state', {
     get() {
       return this.authenticated && isActivePipeline(this.get('pipeline'));
@@ -230,6 +232,7 @@ export default Component.extend({
     // hide graph tooltip when event changes
     set(this, 'showTooltip', false);
     this.setStageBuilds();
+    set(this, 'collapsedStages', new Set([]));
   },
   actions: {
     graphClicked(job, mouseevent, sizes) {
@@ -405,6 +408,17 @@ export default Component.extend({
       };
 
       setProperties(this, properties);
+    },
+    onToggleStageView(stageName, isCollapsed) {
+      const collapsedStages = new Set(this.collapsedStages);
+
+      if (isCollapsed) {
+        collapsedStages.add(stageName);
+      } else {
+        collapsedStages.delete(stageName);
+      }
+
+      setProperties(this, { collapsedStages });
     },
     confirmStartBuild(newEventStartFromType) {
       setProperties(this, {
--- a/app/components/pipeline-workflow/template.hbs
+++ b/app/components/pipeline-workflow/template.hbs
@@ -16,9 +16,11 @@
     @causeMessage={{this.selectedEventObj.causeMessage}}
     @graphClicked={{action "graphClicked"}}
     @onShowStageActionsMenu={{action "onShowStageActionsMenu"}}
+    @onToggleStageView={{action "onToggleStageView"}}
     @isSkipped={{this.selectedEventObj.isSkipped}}
     @prChainEnabled={{this.prChainEnabled}}
     @stages={{this.stages}}
+    @collapsedStages={{this.collapsedStages}}
     @displayStageMenuHandle={{this.displayStageMenuHandle}}
   >
     <WorkflowTooltip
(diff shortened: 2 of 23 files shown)


minghay (changes requested):
1. Doesn't work in V2 UI
2. Stages collapse to different heights
Screenshot 2025-02-18 at 14-45-29 minghay_stages
3. Stages of the following form completely disappear
Screenshot 2025-02-18 at 14-44-54 minghay_stages
4. It's really difficult to tell what the collapse button indicates --- it looks a bit like a dot.  The black background also doesn't help so the whole styling around it should be revisited.
5. After collapsing, the stage options is completely broken and throws an error.

sagar1312:
> 1. Doesn't work in V2 UI

Fixed

> 2. Stages collapse to different heights

It is as expected to acomodate the stage description.

> 3. Stages of the following form completely disappear

Fixed. Detached stages no longer disappears when collapsed.

> 4. It's really difficult to tell what the collapse button indicates --- it looks a bit like a dot.  The black background also doesn't help so the whole styling around it should be revisited.

Fixed the styling to make `-` more wider 

> 5. After collapsing, the stage options is completely broken and throws an error.

Fixed

@minghay Please take a look

minghay:
1. When a stage is collapsed, an error is still being thrown when selecting the pipeline options within the V2 UI

2. There's also another edge rendering issue with collapsed stages.  In the original uncollapsed state, you can see the edges are rendered without intersecting the job nodes (in particular the `pull` and `publish` jobs):
Screenshot 2025-02-21 at 11-58-28 Events

When the stage is collapsed, the edges cut directly through the job node and actually merge into the previous edge somewhere after the arrowhead
Screenshot 2025-02-21 at 11-59-00 Events

minghay (changes requested):
Still a few outstanding issues:
1. The collapsed state of a stage is not preserved between graph updates in the V1 interface (i.e., build status changes update the graph and then stage that was previously collapsed is re-expanded during the update)
2. In the V1 interface, when a job node updates from running to complete and the following job is part of a collapsed stage, the graph will not update correctly)
3. In the V2 interface, the graph edges into and out of a stage are not rendered correctly
Screenshot 2025-03-18 at 14-24-36 Events
Screenshot 2025-03-18 at 14-40-07 Events
4. The stage border occasionally renders as as green before the stage has fully completed

sagar1312:
> Still a few outstanding issues:
> 
> 1. The collapsed state of a stage is not preserved between graph updates in the V1 interface (i.e., build status changes update the graph and then stage that was previously collapsed is re-expanded during the update)

This is a new ask. The requirement to support this needs to hashed out. We should have a separate issue to track and implement this.
> 2. In the V1 interface, when a job node updates from running to complete and the following job is part of a collapsed stage, the graph will not update correctly)
> 3. In the V2 interface, the graph edges into and out of a stage are not rendered correctly
> 4. The stage border occasionally renders as as green before the stage has fully completed
Please share a sample pipeline to reproduce these

minghay:
> > 2. In the V1 interface, when a job node updates from running to complete and the following job is part of a collapsed stage, the graph will not update correctly)
> 
> Please share a sample pipeline to reproduce
> 
> > 4. The stage border occasionally renders as as green before the stage has fully completed
> 
> Please share a sample pipeline to reproduce

Seems like these are resolved now.

minghay (dismissed):
```
version: 4
shared:
  steps:
            - build: echo DONE
annotations:
    screwdriver.cd/chainPR: true

jobs:
    main:
        requires: [~commit]
    s:
        requires: []

    s2:
        requires: []
        
    d:
        requires: []
        
    d2:
        requires: []
        
    end:
        requires: [~stage@single, ~stage@with-setup-and-teardown, ~stage@detached, ~stage@detached-with-setup-and-teardown]

stages:
    single:
        jobs: [s]
        requires: [main]

    with-setup-and-teardown:
        jobs: [s2]
        requires: [main]
        setup:
            steps: 
                - build: echo 'More done'
        teardown:
            steps:
                - build: echo 'More done'
            
    detached:
        jobs: [d]
        
    detached-with-setup-and-teardown:
        jobs: [d2]
        setup:
            steps: 
                - build: echo 'More done'
        teardown:
            steps:
                - build: echo 'More done'
```

Collapsed stages may swap positions on the user.

In V2 UI:
1. If you collapse the `with-setup-and-teardown` stage, the `single` stage is unable to be collapsed.
2. If you collapse the `single` stage, then collapse the `with-setup-and-teardown` stage, the `with-setup-and-teardown` is unable to be re-expanded.

VonnyJap (approved):

PR #1350, opened 2025-02-26
feat: Modal for deleting pipeline

minghay:
## Context
Introduce a new modal for deleting a pipeline.

## Objective
Creates a new modal that allows for deleting a pipeline.  When a pipeline is deleted, the callback handler is executed and any additional logic for the UI is left to those respective components.

Screenshot 2025-02-25 at 16-28-56 ChildPipelines

## References

## License


VonnyJap:
So this new implementation looks different from how the current UI looks like. 
1. Can we have a warning to tell users that this action cannot be undone?
2. Can we have Yes/No as selections for users?
3. The current modal takes some input for a user to make sure that they are typing the pipeline to be deleted, any reason not to have it in the new implementation?

minghay:
So this implementation was based off of the delete modal that current is used in the child pipeline pages:
Screenshot 2025-02-26 at 08-52-28 minghay_parent Child Pipelines

The plan was to update the modal some more when the pipeline options page was being updated so that both modals would be brought in line, but you're right, the modal shouldn't be quite so simple right now either so I'll make those changes.

As for your second point with the `No` as a button selection, the UI modals in the new interface don't have that as an explicit option as it's the point of the modals is to confirm the action to be taken.  For actions that want to be canceled, the modal just needs to be closed.

minghay:
a223098 updates the modal to the following:
Screenshot 2025-02-26 at 10-58-44 ChildPipelines

app/components/pipeline/modal/delete-pipeline/template.hbs
```diff
@@ -0,0 +1,59 @@
...
+        @icon="triangle-exclamation"
+        @dismissible={{false}}
+      />
+    {{/if}}
+
+    <div id="delete-repository-message">
+      <div class="delete-repository-message-warning">
+        <FaIcon @icon="triangle-exclamation"/>
+        <b>Deleting this pipeline is permanent!</b>
+        <FaIcon @icon="triangle-exclamation"/>
```
VonnyJap, line 23: one icon is enough?

minghay: I don't have a particular preference between 1 or 2 icons.  Which one would you prefer?

VonnyJap: one is good, I feel two becomes cluttered, sorry nitpick.

minghay: One it is!

VonnyJap (approved):

PR #1357, opened 2025-03-04
feat(3064): Adjust the border of stage in the workflow graph based on the status

sagar1312:
## Context

Stage border is always rendered in grey color.

**Before:**
image

image

## Objective

Stage border color should reflect the overall execution status of the stage which is derived from the statuses of individual job builds with in that stage.

Stage border should be:
- green when all jobs are successfully completed
- red when one or more jobs failed

**After**
image

image

## References

https://github.com/screwdriver-cd/screwdriver/issues/3064

## License

### DIFF
--- a/app/components/pipeline/workflow/component.js
+++ b/app/components/pipeline/workflow/component.js
@@ -185,7 +185,9 @@ export default class PipelineWorkflowComponent extends Component {
       this.event.id
     );
 
-    if (!this.isEventComplete(builds)) {
+    const hasEventCompleted = this.isEventComplete(builds);
+
+    if (!hasEventCompleted) {
       this.workflowDataReload.registerBuildsCallback(
         BUILD_QUEUE_NAME,
         event.id,
@@ -198,14 +200,15 @@ export default class PipelineWorkflowComponent extends Component {
         event.id
       );
 
-      if (!stageBuilds) {
+      this.stageBuilds = stageBuilds;
+
+      if (!stageBuilds || !hasEventCompleted) {
         this.workflowDataReload.registerStageBuildsCallback(
           STAGE_BUILD_QUEUE_NAME,
           event.id,
           this.stageBuildsCallback
         );
       }
-      this.stageBuilds = stageBuilds;
     }
 
     this.event = event;


minghay (changes requested):
1.  You've got failing tests that you still need to fix
2.  Specifically for the V2 UI, only pipelines with stages should be making the API call to `/events/${eventId}/stageBuilds`.  Right now you have it so that all pipelines are making those API calls

app/workflow-data-reload/stageBuildsDataReloader.js
minghay, line 1: Add in the respective unit tests for this data reloader class.  The most critical one is the test for `fetchDataForId` since that needs to return data.  The other one is expected to always use the `responseCache`

app/components/pipeline/workflow/component.js
```diff
@@ -143,6 +147,18 @@ export default class PipelineWorkflowComponent extends Component {
         this.event.id,
         this.buildsCallback
       );
+
+      const { workflowGraph } = this.event;
+      const eventStages = extractEventStages(workflowGraph);
+
+      if (eventStages && eventStages.length) {
```
minghay, line 154:
```suggestion
      if (this.stages.length > 0) {
```

app/components/pipeline/workflow/component.js
```diff
@@ -4,10 +4,12 @@ import { action } from '@ember/object';
 import { service } from '@ember/service';
 import { isComplete, isSkipped } from 'screwdriver-ui/utils/pipeline/event';
 import { getDisplayJobNameLength, getWorkflowGraph } from './util';
+import { extractEventStages } from '../../../utils/graph-tools';
```
minghay, line 7:
```suggestion
```

minghay (changes requested):

app/components/pipeline/workflow/component.js
sagar1312: Event stages is a snapshot of pipeline stages when the event was created. We don't need to load/reload stage builds if an event was created when the pipeline did not have stages configured.

So we cannot use pipeline stages here.

minghay: That's fine.  There's other issues with the event snapshot to be addressed at a later time.

`extractEventStages` returns an array so just evaluate the length.
```suggestion
      const { workflowGraph } = this.event;
      const eventStages = extractEventStages(workflowGraph);

      if (eventStages.length) {
```

minghay (changes requested):

minghay:
I missed catching this before, but stage coloring doesn't update in the v2 workflow graph when a different event is selected from the event cards.

app/components/pipeline/workflow/component.js
```diff
@@ -156,22 +172,36 @@ export default class PipelineWorkflowComponent extends Component {
...
     if (!this.isEventComplete(builds)) {
       this.workflowDataReload.registerBuildsCallback(
         BUILD_QUEUE_NAME,
         event.id,
         this.buildsCallback
       );
+      this.workflowDataReload.registerStageBuildsCallback(
+        STAGE_BUILD_QUEUE_NAME,
+        event.id,
+        this.buildsCallback
```
minghay, line 198:
```suggestion
        this.stageBuildsCallback
```

minghay (changes requested):

app/components/pipeline/workflow/component.js
```diff
@@ -170,6 +193,21 @@ export default class PipelineWorkflowComponent extends Component {
...
+      );
+
+      if (!stageBuilds) {
+        this.workflowDataReload.registerStageBuildsCallback(
+          STAGE_BUILD_QUEUE_NAME,
+          event.id,
+          this.stageBuildsCallback
+        );
+      }
+      this.stageBuilds = stageBuilds;
```
minghay, line 208:
```suggestion
      } else {
        this.stageBuilds = stageBuilds;
      }
```
Wrap this in an `else` to prevent any possible race condition where the `undefined` value of `stageBuilds` overwrites the actual fetched data that was set in the callback.

minghay (changes requested):

app/components/pipeline/workflow/component.js
sagar1312: Moved it before to the `if` block. When the `event` is changed, we should reset the `stageBuilds` before the callback is satisfied.

minghay (approved):

PR #1359, opened 2025-03-05
fix: Handle 201 for start all children

minghay:
## Context
The shuttle service configures the ajax client to expect JSON response bodies.  The API may return a 201 without a body causing the ajax client to attempt to parse invalid JSON.  This is what is currently happening on the `v4/postV4PipelinesIdStartall` API call.

## Objective
Updates the start all children pipelines API call to request the whole error object to be returned so that further introspection can be done on the actual API call and the returned status code.

## References
https://github.com/screwdriver-cd/ui/pull/1354

## License

### DIFF
--- a/app/components/pipeline/modal/start-children/component.js
+++ b/app/components/pipeline/modal/start-children/component.js
@@ -25,15 +25,22 @@ export default class PipelineModalStartChildrenComponent extends Component {
     await this.shuttle
       .fetchFromApi(
         'post',
-        `/pipelines/${this.pipelinePageState.getPipelineId()}/startall`
+        `/pipelines/${this.pipelinePageState.getPipelineId()}/startall`,
+        null,
+        true
       )
       .then(() => {
         this.args.closeModal(true);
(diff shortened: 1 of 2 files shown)


app/components/pipeline/modal/start-children/component.js
```diff
@@ -25,15 +25,22 @@ export default class PipelineModalStartChildrenComponent extends Component {
...
+        true
       )
       .then(() => {
         this.args.closeModal(true);
         this.wasActionSuccessful = true;
       })
       .catch(err => {
-        this.wasActionSuccessful = false;
-        this.errorMessage = err.message;
+        if (err.jqXHR.status === 201) {
```
VonnyJap, line 37: what do you think if this is made to be more of a generic check as such?
```
if (err.jqXHR.status >= 200 && err.jqXHR.status < 300) {
```

minghay: We can do that

VonnyJap (approved):

PR #1360, opened 2025-03-06
feat: Glimmer component for child pipelines

minghay:
## Context
Creating a glimmer component to house the child pipelines table and the button to start all child pipelines.

## Objective
1. Creates a glimmer component that can be used directly within the route template to display the whole child pipeline view
Screenshot 2025-03-05 at 16-08-56 Child pipelines
Screenshot 2025-03-05 at 16-09-17 Child pipelines

2. Updates the child pipeline table style to handle various viewport heights with the proper scrolling and sticky table header

## References

## License

### DIFF
--- a/app/components/pipeline/children/template.hbs
+++ b/app/components/pipeline/children/template.hbs
@@ -1,0 +1,22 @@
+<div id="child-pipelines-container">
+  <div id="start-all-button-container">
+    <BsButton
+      id="start-all-button"
+      @type="primary"
+      @outline={{true}}
+      @disabled={{this.childrenPipelineStarted}}
+      @onClick={{this.showStartAllChildrenModal}}
+    >
+      Start all child pipelines
+      {{#if this.isStartAllChildrenModalOpen}}
+        <Pipeline::Modal::StartChildren
+          @closeModal={{this.closeStartAllChildrenModal}}
+        />
+      {{/if}}
+    </BsButton>
+  </div>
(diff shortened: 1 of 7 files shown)


app/components/pipeline/children/template.hbs
```diff
@@ -0,0 +1,22 @@
...
+      {{#if this.isStartAllChildrenModalOpen}}
+        <Pipeline::Modal::StartChildren
+          @closeModal={{this.closeStartAllChildrenModal}}
+        />
+      {{/if}}
+    </BsButton>
+  </div>
+
+  <div id="child-pipeline-table-container">
+    <Pipeline::Children::Table />
```
VonnyJap, line 20: can you check if the delete column under this table is working as expected? I was not able to delete successfully and it looked like the API call did not pass the appropriate pipeline ID in the request URL, see request below. 
Screenshot 2025-03-06 at 2 09 14 PM
```
Request URL:
http://sd.screwdriver.cd:9001/v4/pipelines/undefined
Request Method:
DELETE
Status Code:
400 Bad Request
Remote Address:
127.0.0.1:9001
Referrer Policy:
strict-origin-when-cross-origin
```

minghay: That's an issue in the modal that pops up.  I'll include the fix for that

VonnyJap (approved):

PR #1374, opened 2025-03-20
feat: Banners component

minghay:
## Context
Expanding on #1373, creates the actual Glimmer component for the banners that will be displayed below the page header.

Banners will now take up a little bit more vertical height, but will have a banner count at the bottom right corner.  Note that banners are no longer dismissible.  Dismissibility of the banners can be revisited in the future to properly design out the actual experience.
Screenshot 2025-03-19 at 17-27-11 Dashboard

When there are multiple banners, the bottom right will surface left and right arrows to allow users to switch through the available banner messages.
Screenshot 2025-03-19 at 15-33-03

For banner messages that are large, vertical scrolling is configured.  This is the other reason why the banners now take up additional vertical space.
Screenshot 2025-03-19 at 15-33-19

## Objective
1. Creates a shared stylesheet that is super barebones for shared styles to theme bootstrap alert components with the right color palette.
2. Creates the Glimmer component for the new banners experience.

## License

### DIFF
--- a/app/components/banners/styles.scss
+++ b/app/components/banners/styles.scss
@@ -1,0 +1,61 @@
+@use 'screwdriver-colors' as colors;
+@use 'variables';
+@use 'bootstrap-alert' as alert;
+@use 'screwdriver-button' as sdButton;
+
+@mixin styles {
+  #banners {
+    @include alert.styles;
+
+    .alert {
+      height: variables.$banners-height;
+      display: flex;
+      flex-direction: column;
+      margin: 0;
+      border: none;
+      border-radius: 0;
+
+      #banner {
+        display: flex;
+
+        svg {
+          margin-top: 0.175rem;
+          margin-right: 0.25rem;
+        }
+
+        .message {
+          height: 2.25rem;
+          overflow: auto;
+          padding-right: 5rem;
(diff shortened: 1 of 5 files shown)


app/components/banners/styles.scss
```diff
@@ -0,0 +1,61 @@
...
+@use 'variables';
+@use 'bootstrap-alert' as alert;
+@use 'screwdriver-button' as sdButton;
+
+@mixin styles {
+  #banners {
+    @include alert.styles;
+
+    .alert {
+      height: variables.$banners-height;
```
minghay, line 11: Note that the `$banners-height` variable does not exist yet in the shared variables sheet.  A future PR will refactor the existing `$nav-banners-height` variable to `$banners-height`

jithine: if it doesn't exist, what value does height gets computed to ?

jithine (approved):

app/components/banners/styles.scss
minghay: It's basically undefined since the variable doesn't exist.  After the follow up PR, the height will shift from the current `3rem` to `5rem`

PR #1386, opened 2025-04-04
fix: Fix pipeline header updating

minghay:
## Context
When switching to a new pipeline via the v2 pipeline workflow graph, the route model change in the pipeline is not updating the pipeline header component.

## Objective
Adds in the necessary tracking to the pipeline header component to update properly when a pipeline changes.

## References

## License

### DIFF
--- a/app/components/pipeline/header/component.js
+++ b/app/components/pipeline/header/component.js
@@ -8,22 +8,21 @@ export default class PipelineHeaderComponent extends Component {
 
   @service shuttle;
 
-  @service pipelinePageState;
-
   @tracked addToCollectionModalOpen = false;
 
   @tracked errorMessage;
 
   @tracked collections = null;
 
-  pipeline;
+  @tracked pipeline;
 
-  sameRepoPipeline = [];
+  @tracked sameRepoPipeline = [];
 
   constructor() {
     super(...arguments);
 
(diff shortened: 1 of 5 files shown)


app/components/pipeline/header/component.js
```diff
@@ -8,22 +8,21 @@ export default class PipelineHeaderComponent extends Component {
...
+  @tracked pipeline;
 
-  sameRepoPipeline = [];
+  @tracked sameRepoPipeline = [];
 
   constructor() {
     super(...arguments);
 
-    this.pipeline = this.pipelinePageState.getPipeline();
+    this.pipeline = this.args.pipeline;
```
VonnyJap, line 24: why is this now changed to use `this.args.pipeline;` instead of `this.pipelinePageState.getPipeline();`?

app/components/pipeline/header/component.js
```diff
@@ -8,22 +8,21 @@ export default class PipelineHeaderComponent extends Component {
...
+    this.getPipelinesWithSameRepo().then(() => {});
```
VonnyJap, line 25: nitpick: can we use `async` and `await` for consistency?

minghay: No, the constructor can not be asynchronous.

VonnyJap (approved):

PR #1390, opened 2025-04-14
feat: Adds status sorting as default for event jobs table

minghay:
## Context
The events job table can more useful if jobs are sorted by the build status.

## Objective
Adds status sorting by default to the event jobs table.  The sort order is based on the build *color* (i.e., red, yellow, green, blue).  Secondary sort order is alphabetical by stage name and then job name.

## References

## License


app/components/pipeline/jobs/table/util.js
```diff
@@ -14,3 +14,43 @@ export default function getDisplayName(job, prNum) {
...
+
+    if (aStageName === bStageName) {
+      return a.job.name.localeCompare(b.job.name);
+    }
+
+    if (aStageName && bStageName) {
+      return aStageName.localeCompare(bStageName);
+    }
+
+    return aStageName ? -1 : 1;
```
VonnyJap, line 37: what happen to line 37 when `bStageName` is either `null` or `undefined`?
```
    if (aStageName === undefined || aStageName === null) return 1;
    if (bStageName === undefined || bStageName === null) return -1;
    return aStageName.localeCompare(bStageName);
```
I think the above implementation makes it more explicit.

app/components/pipeline/jobs/table/util.js
```diff
@@ -14,3 +14,43 @@ export default function getDisplayName(job, prNum) {
...
+  statusOrder.set('WARNING', 2);
+  statusOrder.set('UNSTABLE', 2);
+  statusOrder.set('RUNNING', 3);
+  statusOrder.set('CREATED', 3);
+  statusOrder.set('QUEUED', 3);
+  statusOrder.set('BLOCKED', 3);
+  statusOrder.set('SUCCESS', 4);
+  statusOrder.set('COLLAPSED', 5);
+  statusOrder.set('FROZEN', 5);
+  statusOrder.set(undefined, 6);
```
VonnyJap, line 53: add a note here that this is a custom logic and why it is sorted this way.

app/components/pipeline/jobs/table/util.js
minghay: At line 37, only `aStageName` or `bStageName` has a string value set.  It does not matter which is undefined/null, hence line 3 in your suggestion will never execute.  The only comparison that needs to be made is whether `aStageName` is the one that has a string value assigned and should be sorted accordingly.

app/components/pipeline/jobs/table/util.js
minghay: Added additional information on the status ordering on line 20

VonnyJap (approved):

PR #1400, opened 2025-04-25
Disable AWS cluster logic

minghay:
## Context
The AWS cluster determination logic is really simple.  Additional work should be done with the backend to properly determine if an event is built within AWS.

## Objective
Disables the simple AWS logic for the time being.

## License

### DIFF
--- a/app/components/pipeline/workflow/component.js
+++ b/app/components/pipeline/workflow/component.js
@@ -360,10 +360,11 @@ export default class PipelineWorkflowComponent extends Component {
   isEventInAws(builds) {
     this.isBuildClusterAws = false;
 
-    builds.forEach(build => {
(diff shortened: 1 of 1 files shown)


app/components/pipeline/workflow/component.js
```diff
@@ -360,10 +360,11 @@ export default class PipelineWorkflowComponent extends Component {
   isEventInAws(builds) {
     this.isBuildClusterAws = false;
 
-    builds.forEach(build => {
-      if (build.buildClusterName.startsWith('aws')) {
-        this.isBuildClusterAws = true;
-      }
+    builds.forEach(() => {
```
pritamstyz4ever, line 363: can remove the forEach

pritamstyz4ever (approved):

PR #1593, opened 2026-08-31
feat(3527): fix error messages when the pipeline has been deleted

y-oksaku:
## Context

It is not easy to understand that the repository was deleted from the error messages.

## Objective

This PR fix the error messages when the action is rejected by pipeline not found.

## References

Issue - https://github.com/screwdriver-cd/screwdriver/issues/3527

## License

### DIFF
--- a/app/utils/pipeline.js
+++ b/app/utils/pipeline.js
@@ -109,7 +109,7 @@ const getPipelineErrorMessage = err => {
     return 'The repository was not found. It might have been deleted.';
   }
 
-  return err?.message ?? err?.payload?.message ?? 'Unknown error';
+  return err?.payload?.message ?? err?.message ?? 'Unknown error';
 };
 
 export {
--- a/app/utils/pipeline.js
+++ b/app/utils/pipeline.js
@@ -102,6 +102,16 @@ const hasSonarBadge = pipeline => {
   return false;
 };
 
+const getPipelineErrorMessage = err => {
+  const statusCode = err?.payload?.statusCode || err?.status;
+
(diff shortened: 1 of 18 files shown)


app/utils/pipeline.js
```diff
@@ -102,6 +102,16 @@ const hasSonarBadge = pipeline => {
...
 };
 
+const getPipelineErrorMessage = err => {
+  const statusCode = err?.payload?.statusCode || err?.status;
+
+  if (statusCode === 404) {
+    return 'The repository was not found. It might have been deleted.';
+  }
+
+  return err?.message ?? err?.payload?.message ?? 'Unknown error';
```
kumada626: `confirm-action` previously preferred `err.payload.message` over `err.message`. 
https://github.com/screwdriver-cd/ui/pull/1593/changes#diff-23e4a9946e87e12558fc5f1f0108f995ab73cb6e554c86c1d0c5494ee6494985L228
With this change, if both properties are present, the generic top-level err.message will be displayed instead of the more specific API error message.
Is there a reason you changed the order?

kumada626:
It would be better to add test for `getPipelineErrorMessage` on this test.
https://github.com/screwdriver-cd/ui/blob/master/tests/unit/utils/pipeline-test.js

app/utils/pipeline.js
y-oksaku: It was a mistake. I fixed it.

kumada626 (approved):

minghay (approved):

PR #1596, opened 2026-09-11
fix: reload pipeline state when switch the pipeline page

y-oksaku:
## Context

The Start button keeps disabled when user transitions from disabled pipeline page to another branch pipeline page.

https://github.com/user-attachments/assets/70329ffb-471d-4687-9e55-d9c3b5353292

## Objective

Reload the pipeline state when switch the pipeline page. 

## License


app/pipeline-page-state/service.js
```diff
@@ -4,6 +4,10 @@ import { tracked } from '@glimmer/tracking';
 export default class PipelinePageStateService extends Service {
   pipeline;
```
y-oksaku, line 5: We can add `@tracked` to `pipeline`, but it can hook reset to `null` on the `clear` method.
Then some reference as `pipeline.foobar` throws error.

kumada626 (approved):

yakanechi (approved):

minghay (approved):

PR #1597, opened 2026-09-11
feat: improve validator UX

y-oksaku:
## Context

Validator has the following issues:

- It cannot show yaml format errors of the template validation
- It freeze when request was rejected
- It cannot parse a command yaml

before

## Objective

This PR resolve those issues.

template

command

## License


app/components/build-log/component.js
```diff
@@ -542,7 +542,11 @@ export default Component.extend({
     },
 
     logScroll() {
-      const container = this.element.querySelectorAll('.wrap')[0];
+      const container = this.element?.querySelector('.wrap');
+
+      if (!container) {
+        return;
+      }
```
y-oksaku, line 549: This PR does not modify the BuildLog component itself. 

The new tests may have changed the execution timing and exposed this existing lifecycle race.
https://cd.screwdriver.cd/pipelines/7/builds/999549

Adding a guard for the destroyed component state fixed the test.

kumada626 (approved):

minghay (approved):

